PR #10, opened 2017-05-10
Fix the buildifier test.

dlorenc:
buildifier outputs diffs, but returns an exit code of 0.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,7 +24,7 @@ script:
   - bazel clean && bazel build //...
 # We don't have any tests yet so this fails. Uncomment when we do.
 #  - bazel clean && bazel test --test_output=errors //...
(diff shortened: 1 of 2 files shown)


dlorenc:
cc @mattmoor you might want something like this in rules_docker.

r2d4:
Should we make it a generic script `test.sh` and run all the tests there?

I was also wondering if this should be done through `bazel test`, although I couldn't find any rules that did that already.

dlorenc:
> Should we make it a generic script test.sh and run all the tests there?
> 
Most tests will be caught in the bazel test ./... we'll add back in once we get tests.

> I was also wondering if this should be done through bazel test, although I couldn't find any rules that did that already.
I thought about that, but part of the point of this is to catch invalid BUILD files as well as mis-formatted onces, so it's a bit meta to use bazel test to execute it.

PR #11, opened 2017-05-10
Add ca_certificates to base image

### LINKED ISSUE #9: Incorporate cacerts into base.
I think it makes sense for our base to contain cacerts by default.

r2d4:
It was added in the workspace, but not in the base image

fixes #9

### DIFF
--- a/base/BUILD
+++ b/base/BUILD
@@ -16,5 +16,6 @@ docker_build(
         "@glibc//file",
         "@libssl//file",
         "@openssl//file",
+        "@ca_certificates//file",
     ],
 )


mattmoor:
How confident are you that this works?  :)

IIRC cacerts may utilize functionality that the deb support doesn't handle well.

This is what leads me to suspect this: https://github.com/bazelbuild/bazel/issues/2812

r2d4:
Yeah, looks like `/etc/ssl/certs/ca-certificates.crt` is not being written.  I'll hold off on this until we can run the post installation hooks with the deb bazel rules.

mattmoor:
You might talk to @damienmg about whether he's even looking at that.  Not sure what it would take since he wrote that stuff.

damienmg:
I think it was pre-existent to my contribution :) but anyway the problem is that we cannot run action on the bazel docker_build rules. What I was doing at some point was running the update-java-certificates task in the entrypoint of the docker image as a hacky workaround.

mattmoor:
@damienmg Didn't you write support into `pkr_tar` to avoid the `dpkg` dependency?  If so, can we at least detect and warn when the `deb` has these in it, so that it doesn't silently fail?

damienmg:
There was a support for deb before pkg_tar, pkg_tar was to avoid the dep on dpkg but we were just unpacking with dpkg. We could add a warning but that would be spammy in most of the case IMO.

mattmoor:
I understand that we were just unpacking with `dpkg`, my point is that now that we're not maybe we can detect this and error when users try to use this functionality on a `deb` that utilizes functionality outside the scope of what we support.

I have no sense for how noisy this breakage would be, but frankly the best test would be to change this in the mono-repo (where we use this feature extensively) and see what breaks.

I'll TAL at the rules and see if I can make enough sense of them to explore this.

mattmoor:
For glibc (after expanding with `ar -x`):
```
$ tar tzvf control.tar.gz 
drwxr-xr-x root/root         0 2017-04-28 21:26 ./
-rw-r--r-- root/root       617 2017-04-28 21:26 ./control
```

For ca-certs:
```
$ tar tzvf control.tar.gz 
drwxr-xr-x root/root         0 2017-04-28 16:22 ./
-rw-r--r-- root/root     19488 2017-04-28 16:22 ./md5sums
-rw-r--r-- root/root       919 2017-04-28 16:22 ./control
-rw-r--r-- root/root        70 2017-04-28 16:19 ./triggers
-rwxr-xr-x root/root      1629 2017-04-28 16:22 ./postrm
-rwxr-xr-x root/root      5406 2017-04-28 16:22 ./postinst
-rw-r--r-- root/root     23949 2017-04-28 16:22 ./templates
-rwxr-xr-x root/root     11522 2017-04-28 16:22 ./config
```

I think what we would want to implement such a check would be to add code [here](https://github.com/bazelbuild/bazel/blob/b8449f95847c3e667d81352bf53f8274aa1caaa3/tools/build_defs/pkg/build_tar.py#L162) that also looks for `control.tar.gz`, and peaks inside of it to see whether it has any hooks.

mattmoor:
To unblock the addition of CA certs (for which I do not see a way we can use `pkg_tar`), could we instead construct the filesystem layer ourselves?  It [wouldn't be the first time](https://github.com/httplib2/httplib2/blob/master/python2/httplib2/cacerts.txt) someone bundled certs:.

dlorenc:
Bunding ca-certs that way sgtm.

Would it make sense to restore some of the dpkg functionality, maybe as a different rule or a flag?

mattmoor:
Not clear this was ever supported by the `dpkg` approach.

damienmg:
It was never supported by the dpkg approach.

Running an action without mounting the container is tricky because the host environment won't be identical to the container environment...

dlorenc:
So looking through the postinst scripts, what's happening is basically this:

One script cats all the .crt file names into a /etc/ca-certificates.conf file. This file has one entry per line, where the line corresponds to a .crt file under /usr/share/ca-certificates. Certificates can be "deselected" by prefixing the line with a !.

Another script goes through the conf file, and appends the contents of each "selected" certificate into the final /etc/ssl/ca-certificates.crt file.

How crazy would it be to reimplement this as a bazel rule? We could allow users to specify an overridable conf file to add new certs or deselect default ones.

mattmoor:
@dlorenc Seems like an interesting idea. 

What if we exposed the necessary files via `WORKSPACE`, e.g. `@cacerts//path/to/cert`, and the user simply expressed their desire for their inclusion in your rule via a simple Bazel `glob`?  e.g.
```
  glob(["@cacerts//**/*.crt", exclude=["@cacerts//**/*China*"])
```

We could default to `glob(["@cacerts//**/*.crt"])`.

The idea being that we expose file-level dependencies in `WORKSPACE` that the user can then consume at `BUILD`.  WDYT?

PR #15, opened 2017-05-10
Use udeb for base image instead of deb

r2d4:
udeb's are smaller, stripped down debs that are used as part of the debian installation process.  I'm not sure exactly what the downsides to using them over the regular debs, but it makes the image considerably smaller.  

https://wiki.debian.org/udeb

```
bazel/base:new               1.041 MB
bazel/base:base              14.21 MB
```

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -13,8 +13,8 @@ go_repositories()
 # For the glibc base image.
 http_file(
    name = "glibc",
-   url = "http://deb.debian.org/debian/pool/main/g/glibc/libc6_2.19-18+deb8u9_amd64.deb",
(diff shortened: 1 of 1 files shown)


mattmoor:
Cool :)

r2d4:
Note - they don't exist for all packages, only the ones used in the debian installation process, and I had to resolve them manually (I don't think you can resolve their urls with `--print-uris` or `apt-get`)

dlorenc:
From the guide "it removes the naughty bits (the documentation) to save space".

Given that we have no real way for users to see the documentation inside an image, it's probably safe to use when we can.

PR #22, opened 2017-05-16
Add a busybox-based debug container.

dlorenc:
This was really useful in tracking down an issue with the jetty container I'm working on.

Usage:
```shell
docker run -it bazel/base:debug ash
```

cc @mattmoor @r2d4


mattmoor:
@damienmg Is it possible to predicate the definition of a rule on "opt" vs. "dbg"?

damienmg:
I don't think so :(

damienmg:
Or maybe with a config settings

PR #23, opened 2017-05-16
Add a jetty image.

dlorenc:
I had to switch back to the .deb from the udeb to get this to work.


WORKSPACE
```diff
@@ -35,6 +35,13 @@ http_file(
     sha256 = "d3819ceeb2becc96aea4b0fceda48f6e68f84193526bc67f6895c00796ee99d6",
 )
 
+# For the debug base image.
+http_file(
+    name = "busybox",
+    executable = True,
+    url = "https://busybox.net/downloads/binaries/1.21.1/busybox-x86_64"
```
mattmoor: sha256 :)

WORKSPACE
```diff
@@ -60,6 +67,12 @@ http_file(
    sha256 = "a1402290165e8d91b396a33d79580a4501041e92bdb62ef23929a0c207cd9af9",
 )
 
+# For Jetty
+http_file(
+   name = "jetty",
+   url = "http://central.maven.org/maven2/org/eclipse/jetty/jetty-distribution/9.4.4.v20170414/jetty-distribution-9.4.4.v20170414.tar.gz",
+)
```
mattmoor: sha256 :)

java/BUILD
```diff
@@ -16,3 +16,21 @@ docker_build(
...
     ],
 )
+
+# The default jetty tarball has a top-level versioned directory. We need to strip this directory
+# so that we can unpack directly into /jetty in our final image.
+genrule(
+    name = "unpacked_jetty",
+    srcs = ["@jetty//file"],
+    outs = ["_unpacked_jetty.tar.gz"],
+    cmd = "mkdir -p $(@D)/jetty && tar --strip-components=1 -xzf $(<) -C $(@D)/jetty && tar -czf $(@) -C $(@D) jetty  && rm -rf $(@D)/jetty",
```
mattmoor: Not that we've extensively tested windows, but there's no way this will work.  IIRC what I was doing in my prototype *might* have worked...

java/BUILD
```diff
@@ -16,3 +16,21 @@ docker_build(
...
+    visibility = ["//visibility:private"],
+)
+
+docker_build(
+    name = "jetty",
+    base = ":java8",
+    cmd = ["/jetty/start.jar"],
+    tars = [":_unpacked_jetty.tar.gz"],
+    workdir = "/jetty/demo-base/",
```
mattmoor: I think workdir should be `/jetty`, so that users don't have to override it.  Can you add a blurb similar to what we have for Java about how we expect folks to consume this?  e.g. Add your WAR under `/jetty/webapps`

java/examples/BUILD
```diff
@@ -0,0 +1,23 @@
+package(default_visibility = ["//visibility:public"])
+
+load("@io_bazel_rules_docker//docker:docker.bzl", "docker_build")
+
+docker_build(
+    name = "image",
+    base = "//java:jetty",
+    directory = "/jetty/webapps",
+    files = [":helloworld.war"],
+    workdir = "/jetty",
```
mattmoor: I'd move this into the base.  We could add the jetty-demo as an example here with:
```
docker_build(
   name = "jetty-demo",
   base = "//java:jetty",
   workdir = "/jetty/jetty-demo",
)
```

java/examples/BUILD
```diff
@@ -0,0 +1,23 @@
+package(default_visibility = ["//visibility:public"])
+
+load("@io_bazel_rules_docker//docker:docker.bzl", "docker_build")
+
+docker_build(
+    name = "image",
+    base = "//java:jetty",
+    directory = "/jetty/webapps",
+    files = [":helloworld.war"],
```
mattmoor: I don't think this is defined, is it?

java/examples/HelloJava.java
```diff
@@ -0,0 +1,7 @@
+package examples;
+
+public class HelloJava {
+    public static void main(String[] args) {
+        System.out.println("Hello world");
+    }
+}
```
mattmoor: newline at EOF

dlorenc: Done.

java/BUILD
dlorenc: Switched it to something closer to yours. I moved the jetty BUILD file out to it's own file so we can buildifier it.

mattmoor: I don't see this change?

dlorenc: Looks like github is having some issues? I refreshed and it's back but I got a few 500s in the middle.

WORKSPACE
```diff
@@ -68,9 +61,12 @@ http_file(
 )
 
 # For Jetty
-http_file(
+new_http_archive(
    name = "jetty",
+   build_file = "BUILD.jetty",
    url = "http://central.maven.org/maven2/org/eclipse/jetty/jetty-distribution/9.4.4.v20170414/jetty-distribution-9.4.4.v20170414.tar.gz",
+   strip_prefix = "jetty-distribution-9.4.4.v20170414/",
```
mattmoor: Consider hoisting a single version constant?

I did this myself here: https://github.com/bazelbuild/rules_docker/blob/master/docker/docker.bzl#L28

mattmoor: Also, did you try using `maven_jar` for this?

buildifier.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-files=$(buildifier -mode=check $(find . -name BUILD -o -name '*.bzl' -type f))
+files=$(buildifier -mode=check $(find . -name 'BUILD*'' -o -name '*.bzl' -type f))
```
mattmoor: extra quote

WORKSPACE
dlorenc: I didn't try, but it would probably be pretty similar. maven_jar just exposes a filegroup that we would need to tar:
https://github.com/bazelbuild/bazel/blob/516f7e9912ad3130ce18322d43be3e2f57c97ca1/tools/build_defs/repo/maven_rules.bzl#L167

Issue #7, opened 2017-05-10
Create a first-class deb WORKSPACE rule


dlorenc:
This rule should make it easier for users to depend on debs from package repositories without needing to use `http_file`.

For UX, we should consider something like:

```BUILD
deb_package(
  name = 'openssl',
  version = '1.0.1f-1ubuntu2.22'
)
```

How should the apt sources fit in here though? Maybe we could add a list of those to the deb_package rule, or create another type of rule to reference apt sources, and depend on those in dep_package?

dlorenc:
cc @mattmoor @r2d4 wdyt?

mattmoor:
I think that we'll want two rules:
```
   deb_mirror(
       name = "blah",
       url = "https://...",
      # Forget what these are called off the top of my head :)
       foo = "main",
       bar = "jessie",
   )

   deb_pkg(
      name = "openssl",
      mirror = ":blah",
      version = "...",
   )
```

Alternately we could specify the channel in the rules, but I think this provides a nice way of sharing the deb source.  WDYT?

r2d4:
Just looking that travis rules we recently added, maybe we want to expose something similar?

https://docs.travis-ci.com/user/installing-dependencies/#Installing-Packages-on-Container-Based-Infrastructure

Maybe a top level

```
deb(
    sources = ["trusty"],
    packages = [
        "gcc-4.8",
        "...",
    ],
)
```

mattmoor:
@r2d4 Fixed?

dlorenc:
I think there's a bit more work around SHA/checksum validation, but that could be a separate issue.

r2d4:
Closing, there's still a bit more work to be done around UX and validation, but I think this issue can be closed because the rules are in a usable state.

PR #26, opened 2017-05-17
dpkg and dpkg_src rules for deb packages

r2d4:
These are bazel rules to help resolve remote locations of the .deb
dependencies in our images.

ref #7 

This is a first pass at adding rules that make the .deb resolution more user friendly and less fragile.  No need to resolve the URLs manually and hardcode into an http_rule.  I'm not sure how to really incorporate them in the WORKSPACE, but I imagined something like this in BUILD files.

```
load("//packages:deb_pkg.bzl", "deb_pkg")
load("//packages:deb_src.bzl", "deb_src")

deb_pkg(
    name = "openssl",
    pkg_name = "openssl",
    source = ":debian_jessie",
)

deb_src(
    name = "debian_jessie",
    arch = "amd64",
    distro = "jessie",
    mirror_url = "http://httpredir.debian.org",
)
```

then you can reference the debs in your docker_rule by 

```
docker_build(
    name = "base",
    base = ":with_tmp",
    debs = [
        ":openssl",
        ...
    ],
)
```

This could also the be the entrypoint for future deb rules like dependency resolution.  I probably need to add version checking and checksum verification to these rules also.

These are my first bazel rules so let me know if I'm on the right track :)

### DIFF
--- a/package_manager/cloudbuild.yaml
+++ b/package_manager/cloudbuild.yaml
@@ -1,0 +1,29 @@
+# A cloud build config to release a PAR binary of
+# the dpkg_parser.par used for the package manager bazel rules
+
+# A cloudbuild is triggered by every commit and $COMMIT_SHA is a built-in 
+# substitution
+
+steps:
+# Build the dpkg_parser PAR file
+# this binary is used by the package manager rules
+# to download debian package lists and debian packages
+- name: gcr.io/cloud-builders/bazel
+  args: [
+    '--output_base', '/workspace',
+    'build', '//package_manager:dpkg_parser.par',
+    # TODO(r2d4): Remove once PAR compilation runs properly inside
+    # the Bazel sandbox on cloudbuild.
+    '--strategy', 'PythonCompile=standalone'
+  ]
+
+# Upload the dpkg_parser PAR file to a GCS bucket
+- name: gcr.io/cloud-builders/gsutil
+  args: [
+    'cp',
+    'bazel-bin/package_manager/dpkg_parser.par',
+    'gs://distroless/package_manager_tools/$COMMIT_SHA/dpkg_parser.par'
+  ]
+
+# We produce no Docker images.
+images: []
--- a/package_manager/dpkg_parser.py
+++ b/package_manager/dpkg_parser.py
@@ -1,0 +1,112 @@
+# Copyright 2017 Google Inc. All rights reserved.
+
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+
+#     http://www.apache.org/licenses/LICENSE-2.0
+
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import argparse
+import gzip
+import io
+import urllib2
+import json
+
+
+from package_manager.parse_metadata import parse_package_metadata
+
+PACKAGES_FILE_NAME = "file/Packages.json"
+DEB_FILE_NAME = "file/pkg.deb"
+FILENAME_KEY = "Filename"
+
+parser = argparse.ArgumentParser(
+    description="Downloads a deb package from a package source file"
+)
+
+parser.add_argument("--packages-file", action='store',
+                    help='The file path of the Packages.gz file')
+parser.add_argument("--package-name", action='store',
+                    help='The name of the package to search for and download')
+
+parser.add_argument("--download-and-extract-only", action='store',
+                    help='If True, download Packages.gz and make urls absolute from mirror url')
+parser.add_argument("--mirror-url", action='store',
+                    help='The base url for the package list mirror')
+parser.add_argument("--arch", action='store',
+                    help='The target architecture for the package list')
+parser.add_argument("--distro", action='store',
+                    help='The target distribution for the package list')
+
+def main():
+    """ A tool for downloading debian packages and package metadata """
+    args = parser.parse_args()
+    if args.download_and_extract_only:
+        download_package_list(args.mirror_url, args.distro, args.arch)
+    else:
+        download_dpkg(args.packages_file, args.package_name)
+
+
+def download_dpkg(packages_file, package_name):
+    """ Using an unzipped, json package file with full urls,
+     downloads a .deb package
+
+    Uses the 'Filename' key to download the .deb package
+    """
+    with open(packages_file, 'rb') as f:
+        metadata = json.load(f)
+    pkg = metadata[package_name]
+    buf = urllib2.urlopen(pkg[FILENAME_KEY])
+    with open(DEB_FILE_NAME, 'w') as f:
+        f.write(buf.read())
+
+
+def download_package_list(mirror_url, distro, arch):
+    """Downloads a debian package list, expands the relative urls,
+    and saves the metadata as a json file
+
+    A debian package list is a gzipped, newline delimited, colon separated
+    file with metadata about all the packages available in that repository.
+    Multiline keys are indented with spaces.
+
+    An example package looks like:
+
+Package: newmail
+Version: 0.5-2
+Installed-Size: 76
+Maintainer: Martin Schulze <joey@debian.org>
+Architecture: amd64
+Depends: libc6 (>= 2.7-1)
+Description: Notificator for incoming mail
+Homepage: http://www.infodrom.org/projects/newmail/
+Description-md5: 49b0168ce625e668ce3031036ad2f541
+Tag: interface::commandline, mail::notification, role::program,
+ scope::utility, works-with::mail
+Section: mail
+Priority: optional
+Filename: pool/main/n/newmail/newmail_0.5-2_amd64.deb
+Size: 14154
+MD5sum: 5cd31aab55877339145517fb6d5646cb
+SHA1: 869934a25a8bb3def0f17fef9221bed2d3a460f9
+SHA256: 52ec3ac93cf8ba038fbcefe1e78f26ca1d59356cdc95e60f987c3f52b3f5e7ef
+
+    """
+    url = "%s/debian/dists/%s/main/binary-%s/Packages.gz" % (
+        mirror_url,
+        distro,
+        arch
+    )
+    buf = urllib2.urlopen(url)
+    f = gzip.GzipFile(fileobj=io.BytesIO(buf.read()))
+    data = f.read()
+    metadata = parse_package_metadata(data, mirror_url)
+    with open(PACKAGES_FILE_NAME, 'w') as f:
+        json.dump(metadata, f)
+
+if __name__ == "__main__":
+    main()
--- a/package_manager/parse_metadata_test.py
+++ b/package_manager/parse_metadata_test.py
@@ -1,0 +1,42 @@
+import unittest
+import os
+
+from package_manager.parse_metadata import parse_package_metadata
+
+class TestParseMetadata(unittest.TestCase):
+
+
+    def setUp(self):
+        current_dir = os.path.dirname(__file__)
+        filename = os.path.join(current_dir, 'testdata', 'Packages.txt')
+        with open(filename) as f:
+            data = f.read()
+        self.data = data
+        self.mirror_url = "http://debian.org"
+        self.metadata = parse_package_metadata(self.data, self.mirror_url)
+
+    def test_url_rewrite(self):
+        """ Relative url should have gotten rewritten with absolute url """
+        self.assertEqual(
+            self.metadata["libnewlib-dev"]["Filename"],
+            self.mirror_url + "/debian/" + "pool/main/n/newlib/libnewlib-dev_2.1.0+git20140818.1a8323b-2_all.deb")
+
+    def test_get_all_packages(self):
+        """ Parser should identify all packages """
+        expected_packages = ["libnewlib-dev",
+                             "libnewlib-doc",
+                             "newlib-source",
+                             "newmail",
+                             "zzuf",]
+        for pkg in expected_packages:
+            self.assertEqual(expected_packages.sort(), self.metadata.keys().sort())
+
+    def test_multiline_key(self):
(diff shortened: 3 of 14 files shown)


packages/deb_pkg.bzl
```diff
@@ -0,0 +1,31 @@
...
+  ]
+
+  ctx.action(
+    executable = ctx.executable.package_getter,
+    inputs = [ctx.file.source],
+    arguments = args,
+    outputs = [ctx.outputs.deb],
+  )
+
+deb_pkg = rule(
```
mattmoor: We want this to be a `repository_rule`, so that it can run in `WORKSPACE`, which brings with it new challenges (e.g. you can't depend on building the tool you plan to run).

Let's talk about this more once you are back, but what I do for `docker_pull` is publish a PAR file.
 - cloudbuild to produce it on release tags: https://github.com/google/containerregistry/blob/master/puller.yaml
 - where the binary is pulled down: https://github.com/bazelbuild/rules_docker/blob/master/docker/docker.bzl#L32
 - where the binary is used: https://github.com/bazelbuild/rules_docker/blob/master/docker/pull.bzl#L67

Once this is working, we should consider where the best place for this rule would be.  @damienmg FYI

mmikulicic:
cool; I was about doing something similar then I found this

shouldn't we force a version and/or sha of the `deb_pkg` being referenced (in order to have correct incremental and reproducible builds)?

mattmoor:
@mmikulicic Yes, for reproducible builds, you'd want a checksum.  I generally call this "pseudo-hermetic" because while it's not hermetic you always get the same thing (or fail due to availability).  Sort of like `-Werror` for reproducibility I guess.

mmikulicic:
@mattmoor I don't know the exact definition of hermeticity that everybody can agree on, but I thought the interesting property is that tooling can reason about producing outputs as a pure function of the inputs by having a complete visibility of all the components that are touched while doing a build (including the tooling itself).

If something is a local file or a remote file surely has practical implications on how likely is to fail, but from a theoretical point of view a local file is no different: the disk I/O can file as well :-)

Or did I misunderstand your comment?

mattmoor:
@mmikulicic So hermetic is an effectively closed environment, e.g. restricted to your mono-repo.  This would be Bazel without a `WORKSPACE` at all, as we use it internally.

When references reach outside of your not-so-mono-repo, they can be either mutable (in the sense of different versions over time) or immutable (maybe it can go away, but not change).

So when I say "pseudo-hermetic", what I mean is that everything you need is contained within your repo EXCEPT stuff (transitively) referenced in an immutable way.  

So it's sort of hermetic in the sense that your repo *defines* the world exactly (reproducibly), but doesn't fully *contain* it.

mmikulicic:
I always wondered why bazel's `local_repository` requires an absolute path; with a relative path it would be straightforward to vendor other repositories thus making it easier to achieve this kind of fully local hermeticity.

mattmoor:
@mmikulicic I think the docs say absolute, but in at least one failed (for other reasons) experiment, I was able to use relative paths w/ `local_repository`.

Consider opening an issue on `bazelbuild/bazel` to discuss?  Feel free to CC me.

r2d4:
Sorry for the delay - I've updated this PR with a few things:

* Split into two commits for now - one introducing the rules and one using the rules in the distroless images.  This is because these rules might not live in this repo.
* The dpkg parser is now written in python rather than go.  Since we are building and uploading binaries, using go would have required uploading 3 binaries and doing platform detection which turned out to be a little tricky.

Future TODOs:

* Support specifying a version in the dpkg rule
* Support expanding dependencies for packages (This may be trickier than it seems)
* Checksum validation for dpkg_src rule.  We can validate checksums from the Packages.gz file in the dpkg rule.
* Not reusing the rule name for the package name, since not all package names are valid rule names (e.g. libstdc++)

cc @dlorenc @mattmoor

package_manager/cloudbuild.yaml
```diff
@@ -0,0 +1,18 @@
+steps:
+- name: gcr.io/cloud-builders/bazel
```
dlorenc: Could you add a few file-level comments explaining what this is for?

package_manager/dpkg_parser.py
```diff
@@ -0,0 +1,99 @@
+import argparse
```
dlorenc: I think we need a copyright string up here, also a file-level comment explaining this.

package_manager/dpkg_parser.py
```diff
@@ -0,0 +1,99 @@
...
+    else:
+        download_dpkg(args.packages_file, args.package_name)
+
+
+def download_dpkg(packages_file, package_name):
+    with open(packages_file, 'rb') as f:
+        metadata = json.load(f)
+    for pkg in metadata:
+        if pkg["Package"] == package_name:
+            print(pkg)
```
dlorenc: Probably drop the print?

package_manager/dpkg_parser.py
```diff
@@ -0,0 +1,99 @@
...
+    if args.download_and_extract_only:
+        download_package_list(args.mirror_url, args.distro, args.arch)
+    else:
+        download_dpkg(args.packages_file, args.package_name)
+
+
+def download_dpkg(packages_file, package_name):
+    with open(packages_file, 'rb') as f:
+        metadata = json.load(f)
+    for pkg in metadata:
```
dlorenc: Maybe add an else: to the for loop to fail if we never find the package?

package_manager/dpkg_parser.py
```diff
@@ -0,0 +1,99 @@
...
+                f.write(buf.read())
+            break
+
+def download_package_list(mirror_url, distro, arch):
+    url = "%s/debian/dists/%s/main/binary-%s/Packages.gz" % (
+        mirror_url,
+        distro,
+        arch
+    )
+    print(url)
```
dlorenc: Drop the print?

package_manager/dpkg_parser.py
```diff
@@ -0,0 +1,99 @@
...
+    for pkg in metadata:
+        if pkg["Package"] == package_name:
+            print(pkg)
+            buf = urllib2.urlopen(pkg["Filename"])
+            with open(DEB_FILE_NAME, 'w') as f:
+                f.write(buf.read())
+            break
+
+def download_package_list(mirror_url, distro, arch):
+    url = "%s/debian/dists/%s/main/binary-%s/Packages.gz" % (
```
dlorenc: path.join is probably a bit safer.

package_manager/dpkg_parser.py
```diff
@@ -0,0 +1,99 @@
...
+    buf = urllib2.urlopen(url)
+    f = gzip.GzipFile(fileobj=io.BytesIO(buf.read()))
+    data = f.read()
+    metadata = parse_package_metadata(data, ":")
+    for pkg in metadata:
+        pkg["Filename"] = mirror_url + "/debian/" + pkg["Filename"]
+    with open(PACKAGES_FILE_NAME, 'w') as f:
+        json.dump(metadata, f)
+ 
+def parse_package_metadata(data, separator):
```
dlorenc: is the separator ever not a colon?

package_manager/dpkg_parser.py
```diff
@@ -0,0 +1,99 @@
...
+        distro,
+        arch
+    )
+    print(url)
+    buf = urllib2.urlopen(url)
+    f = gzip.GzipFile(fileobj=io.BytesIO(buf.read()))
+    data = f.read()
+    metadata = parse_package_metadata(data, ":")
+    for pkg in metadata:
+        pkg["Filename"] = mirror_url + "/debian/" + pkg["Filename"]
```
dlorenc: path.join

package_manager/dpkg_parser.py
```diff
@@ -0,0 +1,99 @@
...
+    with open(PACKAGES_FILE_NAME, 'w') as f:
+        json.dump(metadata, f)
+ 
+def parse_package_metadata(data, separator):
+    metadata = []
+    found = False
```
dlorenc: found looks unused.

package_manager/dpkg_parser.py
r2d4: fixed

package_manager/dpkg_parser.py
r2d4: will path.join work on windows?  Not too familiar with the python impl but can it also be used for URLs?

package_manager/cloudbuild.yaml
r2d4: done

package_manager/dpkg_parser.py
r2d4: fixed

package_manager/dpkg_parser.py
r2d4: fixed

package_manager/dpkg_parser.py
r2d4: Some other package managers other than debian/apt-get use a different separator. I added it as a constant instead.

package_manager/dpkg_parser.py
r2d4: fixed

package_manager/dpkg_parser.py
```diff
@@ -0,0 +1,113 @@
...
+    buf = urllib2.urlopen(url)
+    f = gzip.GzipFile(fileobj=io.BytesIO(buf.read()))
+    data = f.read()
+    metadata = parse_package_metadata(data)
+    for pkg in metadata:
+        pkg["Filename"] = mirror_url + "/debian/" + pkg["Filename"]
+    with open(PACKAGES_FILE_NAME, 'w') as f:
+        json.dump(metadata, f)
+ 
+def parse_package_metadata(data):
```
mattmoor: We probably want this in a py_library we can write unit tests for.

r2d4:
I split `parse_metadata` into its own py_lib and added a few unit tests.  Pending more review feedback, the only thing left to do is it figure out where we can upload `dpkg_parser.par`, maybe into a distroless bucket for `distroless` @mattmoor ?

package_manager/dpkg_parser.py
```diff
@@ -0,0 +1,75 @@
...
+    url = "%s/debian/dists/%s/main/binary-%s/Packages.gz" % (
+        mirror_url,
+        distro,
+        arch
+    )
+    buf = urllib2.urlopen(url)
+    f = gzip.GzipFile(fileobj=io.BytesIO(buf.read()))
+    data = f.read()
+    metadata = parse_package_metadata(data, mirror_url)
+    with open(PACKAGES_FILE_NAME, 'w') as f:
```
dlorenc: Some comments explaining the schema of this JSON file and explaining why we need it at all would be useful.

package_manager/parse_metadata_test.py
```diff
@@ -0,0 +1,33 @@
...
+            self.mirror_url + "/debian/" + "pool/main/n/newlib/libnewlib-dev_2.1.0+git20140818.1a8323b-2_all.deb")
+
+    def test_get_all_packages(self):
+        expected_packages = ["libnewlib-dev",
+                             "libnewlib-doc",
+                             "newlib-source",
+                             "newmail",
+                             "zzuf",]
+        for pkg in expected_packages:
+            ok = self.metadata[pkg]
```
dlorenc: What is this supposed to do, just check that the package is in metadata? You can use something like self.assertContains instead.

package_manager/dpkg_parser.py
r2d4: I added a few more comments throughout, and docstrings to each function

package_manager/parse_metadata_test.py
r2d4: Yep, switched it over to self.assertContains

package_manager/cloudbuild.yaml
```diff
@@ -0,0 +1,30 @@
+# A cloud build config to release a PAR binary of
+# the dpkg_parser.par used for the package manager bazel rules
+
+# Takes a substitution of _VERSION
```
dlorenc: Does this substitution work via triggers? We can set this up to run automatically everytime a PR is merged.

package_manager/dpkg_parser.py
```diff
@@ -0,0 +1,109 @@
...
+def main():
+    args = parser.parse_args()
+    if args.download_and_extract_only:
+        download_package_list(args.mirror_url, args.distro, args.arch)
+    else:
+        download_dpkg(args.packages_file, args.package_name)
+
+
+def download_dpkg(packages_file, package_name):
+    """ Using an unzipped, json package file with full urls, downloads a .deb package
```
dlorenc: Nit: no spaces before or after """. We might want to enable pylint via travis.

package_manager/dpkg_parser.py
```diff
@@ -0,0 +1,109 @@
...
+    else:
+        download_dpkg(args.packages_file, args.package_name)
+
+
+def download_dpkg(packages_file, package_name):
+    """ Using an unzipped, json package file with full urls, downloads a .deb package 
+    
+    Uses the 'Filename' key to download the .deb package
+    """
+    found = False
```
dlorenc: found looks unused.

package_manager/parse_metadata_test.py
```diff
@@ -0,0 +1,33 @@
+import unittest
+import os
+
+from parse_metadata import parse_package_metadata
+
+class TestParseMetadata(unittest.TestCase):
+
+    def setUp(self):
+        dir = os.path.dirname(__file__)
```
dlorenc: nit: dir is technically a builtin you're shadowing here. A different name would be better.

package_manager/dpkg_parser.py
r2d4: Fixed

package_manager/dpkg_parser.py
r2d4: Fixed

package_manager/parse_metadata_test.py
r2d4: Fixed

package_manager/cloudbuild.yaml
r2d4: I added a trigger on the project for the master branch and versioned it with $COMMIT_SHA.  Not sure if this is the best solution, since we really don't care if its built on _every_ commit, but tags are probably not appropriate since this is just a tool that lives in the distroless repo.

PR #31, opened 2017-05-18
Reorganize examples to make them language-specific directories.

dlorenc:
This will make it easier to link to a language-specific README inside a language-specific directory :)


examples/go/BUILD
```diff
@@ -0,0 +1,19 @@
...
+    name = "hello_go",
+    srcs = ["main.go"],
+)
+
+docker_build(
+    name = "go_example",
+    base = "//base:base",
+    cmd = ["./hello_go"],
+    files = [":hello_go"],
+)
```
mattmoor: newline @ EOF

dlorenc:
> Can you confirm the Java move works? I remember battling with java because I originally had it under examples/, but Bazel freaked out about stuff.

Ah good call, it builds but doesn't run. I'll dig in a bit more.

dlorenc:
OK, this should be good to go.

PR #25, opened 2017-05-17
Add structure tests for our images.

dlorenc:
This is a companion PR to https://github.com/GoogleCloudPlatform/runtimes-common/pull/180

This won't work until that is merged and I include the workspace changes, but I've tested this locally using local_repository.

PR #35, opened 2017-05-18
Switch back to JRE.

mattmoor:


dlorenc:
Nice, sorry about this. Once I get #25 in this should be easier to catch.

PR #37, opened 2017-05-19
Add an image that uses a basic libc call from go.

dlorenc:
This test fails without /etc/passwd or glibc.

### DIFF
--- a/base/BUILD
+++ b/base/BUILD
@@ -2,6 +2,7 @@ package(default_visibility = ["//visibility:public"])
 
 load("@io_bazel_rules_docker//docker:docker.bzl", "docker_build")
 load("@io_bazel_rules_docker//docker/contrib:passwd.bzl", "passwd_file")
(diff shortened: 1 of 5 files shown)


base/BUILD
```diff
@@ -41,10 +43,30 @@ docker_build(
...
+
+# Test that the /etc/passwd entry is set and that libc works (for Go)
+
+go_binary(
+    name = "user",
+    srcs = ["testdata/user.go"],
+    visibility = ["//visibility:private"],
+)
+
+docker_build(
```
mattmoor: Maybe spice it up with a custom user and move to examples?

dlorenc: I started with it in examples, but I changed my mind. The idea is that this is supposed to be a test of the base image's glibc installation, so I thought it would make more sense here. But it is also sort of an example...

WDYT makes more sense?

mattmoor: I think if we switch it away from `root` it would make a good example :)

dlorenc: Done.

PR #38, opened 2017-05-20
Add ca-certificates to our base image.

dlorenc:
Ref #11

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -88,9 +88,3 @@ git_repository(
     remote = "https://github.com/GoogleCloudPlatform/runtimes-common.git",
     commit = "4e9b3b57efb237ba09b4a319bea42bcdd4eb91e3",
 )
-
-git_repository(
-    name ="bazel",
-    remote = "https://github.com/bazelbuild/bazel.git",
-    commit = "2633a599532e73f8f66d346b40b6c06896bab0dd",
-)
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -88,3 +88,9 @@ git_repository(
     remote = "https://github.com/GoogleCloudPlatform/runtimes-common.git",
     commit = "4e9b3b57efb237ba09b4a319bea42bcdd4eb91e3",
 )
+
+git_repository(
+    name ="bazel",
+    remote = "https://github.com/bazelbuild/bazel.git",
+    commit = "2633a599532e73f8f66d346b40b6c06896bab0dd",
+)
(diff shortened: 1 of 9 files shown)


dlorenc:
We should add a test to make sure these are actually usable by an app. I started to do this in Go, but I'm on my mac laptop and bazel can't cross compile :(

dlorenc:
> We should add a test to make sure these are actually usable by an app. I started to do this in Go, but I'm on my mac laptop and bazel can't cross compile :(

Got the test in. I kept this one out of examples and base/testdata

WORKSPACE
```diff
@@ -88,3 +88,9 @@ git_repository(
     remote = "https://github.com/GoogleCloudPlatform/runtimes-common.git",
     commit = "4e9b3b57efb237ba09b4a319bea42bcdd4eb91e3",
 )
+
+git_repository(
```
mattmoor: I don't understand the need for this?

cacerts/BUILD
```diff
@@ -0,0 +1,6 @@
+package(default_visibility = ["//visibility:public"])
```
mattmoor: Let's make this visible to just our repo for now (let's us iterate on the signature before folks depend on it).

I think this would be `//:__subpackages__`

base/testdata/check_certs.go
```diff
@@ -0,0 +1,20 @@
+package main
```
mattmoor: I think we need to be more vigilant about including license headers.

WORKSPACE
dlorenc: Ah, I was originally trying to use their archive.py library to crack open the .deb, but then ran into a wall when I found Python can't open .xz files easily.

base/testdata/check_certs.go
dlorenc: Yeah I have a script for this over in minikube. I'll see if it's easily portable over to here.

dlorenc:
@mattmoor PTAL

WORKSPACE
mattmoor: Yeah, the library shells out to `xzcat`.

cacerts/extract.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+
+# This script extracts the CA certs from the typical debian ca-certificates debian package.
+# It would be nicer to do this in Python or Go, but neither of these languages have packages
+# that can extract .xz files in their stdlibs.
+
+DEB=$1
```
mattmoor: IIRC this has a `copyright`, we should include that as well.

dlorenc: Added our headers.

Issue #68, opened 2017-06-29
dpkg rule does not create output reproducibly


MarkusTeufelberger:
If I run the exact same rule today and in three weeks, I might get a completely different version of the specified package.

Please add a hash to the rule, so it is possible to pin down exact versions.

Also it would be nice to be able to specify several package sources, since not all mirrors might keep older versions of packages around.

MarkusTeufelberger:
#71 does fix the dpkg_src rule, not the dpkg rule itself.

I want control over which package gets installed, not only which global state of all possible package versions.
If the dpkg_src defines libfoo 0.1.2, but I want libfoo 0.1.3, I have to change to a more recent snapshot in dpkg_src and I risk that all my other dependencies suddenly change as well.

r2d4:
I'm going to close this - the new `dpkg_src` rule takes a sha256 argument.  The source contains checksums for each package, which are checked upon downloading in `dpkg_list`.

As for package versioning - 
A `dpkg_list` rule can take multiple sources and will check each one in order.  If you need more granularity, you can use multiple `dpkg_src` and `dpkg_list` rules.

In the future, I think it would probably be good to include a "versioned" syntax to the `dpkg_list` rule, so that you could easily specify versions with something like `openssl=1.0.0`

PR #71, opened 2017-07-05
Refactor dpkg rules

r2d4:
* Allows multiple sources on the dpkg() rule
Sources will be tried in the order they appear.  For different priority, split up your package bundles into different dpkg rules.

* Automatically renames packages that are not valid bazel rule names.
Prints out a warning with instructions on how to reference the resulting deb.  The best solution I can think of here is to eventually output a tarball of deb packages.  Then, the output wouldn't have to be the individual deb files (which for some packages won't be able to be referenced by bazel labels).

* Allows multiple packages to be part of the same dpkg() rule

* Bumps version of dpkg_parser.par to v0.2

Fixes https://github.com/GoogleCloudPlatform/distroless/issues/70
Partially addresses  #68

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -14,7 +14,7 @@ load(
     "//package_manager:package_manager.bzl",
     "package_manager_repositories",
     "dpkg_src",
-    "dpkg",
+    "dpkg_list",
 )
 
 package_manager_repositories()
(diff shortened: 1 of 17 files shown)


base/BUILD
```diff
@@ -32,9 +32,9 @@ docker_build(
     name = "base",
     base = ":with_tmp",
     debs = [
-        "@libc6//file:pkg.deb",
-        "@libssl1.0.0//file:pkg.deb",
-        "@openssl//file:pkg.deb",
+        "@package_bundle//file:libc6.deb",
+        "@package_bundle//file:libssl1.0.0.deb",
+        "@package_bundle//file:openssl.deb",
```
mattmoor: This is cool.  I really like this model, it is basically what I'd like to see us do for `package.json` or `requirements.txt`, e.g. have a single `repository` and export multiple targets from within.

cc/BUILD
```diff
@@ -7,7 +7,7 @@ docker_build(
     name = "cc",
     base = "//base:base",
     debs = [
-        "@libgcc1//file:pkg.deb",
-        "@libstdcpp6//file",
+        "@package_bundle//file:libgcc1.deb",
+        "@package_bundle//file:libstdc__6.deb",
```
mattmoor: I thought of another trick for how to handle invalid characters over the weekend, it is relevant here.

within `@package_bundle` generate a `.bzl` file, then you:
```python
load("@package_bundle//file:the.bzl", "packages")
```

Where `packages` is a dictionary of unescaped package names to `BUILD` target, then you replace this line with:
```python
   packages["libstdc++6"],
```

... and voila :)

This enables us to employ `base64` encoding or `sha256` to better avoid collisions.

@alexeagle has been battling something similar, but I don't believe this completely solves his problem.

package_manager/dpkg.bzl
```diff
@@ -1,13 +1,24 @@
 def _dpkg_impl(repository_ctx):
   repository_ctx.file("file/BUILD", """
 package(default_visibility = ["//visibility:public"])
-exports_files(["pkg.deb"])
+deb_files = glob(["*.deb"])
+exports_files(deb_files)
 """)
 
+  for pkg in repository_ctx.attr.packages:
+    if pkg.find("+") > 0:
```
mattmoor: @damienmg is there a better way to check if a string is a valid label?

cc/BUILD
r2d4: Nice idea!  I implemented it

package_manager/dpkg.bzl
r2d4: This check is unnecessary now that the user doesn't have to do any conversion on their end.

r2d4:
In the latest commit I addressed some feedback as well as adding checksum validation for the Packages.gz file and individual .debs.  The builds should be reproducible and secure now.

package_manager/util_test.py
```diff
@@ -0,0 +1,13 @@
+import unittest
+import os
+
+from package_manager import util
+
+CHECKSUM_TXT = "1915adb697103d42655711e7b00a7dbe398a33d7719d6370c01001273010d069"
+
+class TestUtil(unittest.TestCase):
+
+
```
mattmoor: nit: one blank line, we should add a pylint check to Travis... :-/

r2d4:
Fixes #68
Fixes #72

package_manager/util_test.py
r2d4: Opened #73, will rebase and fix all the warnings on top of that once it gets in

MarkusTeufelberger:
Please document how exactly one can generate the sha256 hash for the dpkg_src rule. It is a bit hidden in the code...

WORKSPACE
```diff
@@ -23,73 +23,43 @@ dpkg_src(
...
     arch = "amd64",
     distro = "jessie-backports",
-    url = "http://deb.debian.org",
+    sha256 = "2a493443581bdb4be071359f7fb62122741f233d3596545d88239a4e4ec445e8",
+    snapshot = "20170701T034145Z",
+    url = "http://snapshot.debian.org/archive",
 )
 
-# For the glibc base image.
 dpkg(
```
dlorenc: Maybe rename this rule to dpkg_list?

package_manager/dpkg_parser.py
```diff
@@ -17,22 +17,34 @@
 import io
```
dlorenc: It looks like we're not running pylint on this file, could you do that manually and open a bug to get it in travis?

package_manager/dpkg_parser.py
```diff
@@ -17,22 +17,34 @@
...
 import urllib2
 import json
+import hashlib
+
+import re
+
+from os import path
```
dlorenc: Just import os.

package_manager/dpkg_parser.py
```diff
@@ -42,31 +54,51 @@
...
-    pkg = metadata[package_name]
-    buf = urllib2.urlopen(pkg[FILENAME_KEY])
-    with open(DEB_FILE_NAME, 'w') as f:
-        f.write(buf.read())
-
-
-def download_package_list(mirror_url, distro, arch):
+    package_to_rule_map = {}
+    for pkg_name in packages.split(","):
+        found = False
```
dlorenc: You can do this with a for: else: loop. I find those clearer.

r2d4: done

package_manager/dpkg_parser.py
r2d4: done

package_manager/dpkg_parser.py
r2d4: done - pylint was enabled, but after this PR, so i rebased

WORKSPACE
r2d4: renamed

r2d4:
Done.  Added a README to the package manager folder and updated our main readme to point to that for extra info.

PR #77, opened 2017-07-20
Add a version-agnostic symlink

mattmoor:


mattmoor:
I needed this here:
  https://github.com/mattmoor/distroless_images/blob/second-attempt/python2.7/image.bzl#L82

PR #79, opened 2017-07-24
Add the netbase package for /etc/services.

dlorenc:
This is needed for grpc.


base/testdata/base.yaml
```diff
@@ -18,3 +18,6 @@ fileExistenceTests:
 - name: certs_copyright
   path: '/usr/share/doc/ca-certificates/copyright'
   shouldExist: true
+- name: services
+  path: '/etc/services'
+  shouldExist: true
```
r2d4: nit: newline

PR #81, opened 2017-08-04
Add the standard /usr/bin/java symlink.

mattmoor:
This removes some of the guesswork about where the Java binary is in derivative layers.


mattmoor:
@dlorenc When playing with a `java_image`, I want to change the entrypoint to be `java -jar <jar>`, and hard-coding the java path we have seems wrong.  This seems better.

PR #93, opened 2017-08-28
Don't hardcode /debian in apt URLs

benley:
This is required in order to support repositories like `debian-security` and `ubuntu`.  The base `/debian` path isn't invariant, it's meant to be a part of the repository URL.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -25,7 +25,7 @@ dpkg_src(
     distro = "jessie",
     sha256 = "142cceae78a1343e66a0d27f1b142c406243d7940f626972c2c39ef71499ce61",
     snapshot = "20170821T035341Z",
-    url = "http://snapshot.debian.org/archive",
+    url = "http://snapshot.debian.org/archive/debian",
 )
 
 dpkg_src(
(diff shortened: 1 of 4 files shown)


benley:
The travis build is failing because the tests depend on a pre-built version of dpkg_parser.par that's downloaded from cloud storage in https://github.com/GoogleCloudPlatform/distroless/blob/master/package_manager/package_manager.bzl ... I'm not sure how to go about making this change while keeping all the tests passing without uploading a new dpkg_parser.par and changing that URL to point at it.  Maybe it has to be done in two PRs?

r2d4:
I've been thinking about that @benley.  Testing locally is also difficult.  @mattmoor solved this for the docker_rules + containerregistry helper since the containerregistry helper is in its own repository.  (See https://github.com/GoogleCloudPlatform/distroless/pull/26#discussion_r117253592).  This tool is still part of the distroless repo, so it might be tricky to build-on-commit, then use that tool without explicitly depending on it through bazel.

benley:
Hmm .... I think it could work if you used `bazel build --override_repository` to point at a locally-built copy of the tool in test.sh, if it's okay to make it a two-step build process.

dlorenc:
@r2d4 what do you think the best way of merging this would be?

r2d4:
This should be fixed by #121 and #122, since you can now pass in entirely custom urls

PR #96, opened 2017-09-05
Add deb_packages rule

MarkusTeufelberger:
Split from #94, this PR only contains the (slightly updated) deb_loader ruleset.

This will not work out of the box as-is, due to the fact that the binary for downloading files has to be served via HTTP (see `deb_loader/deb_loader.bzl`). It'll currently look on localhost instead of a publicly available location.

It is very likely that 2 additional parameters (ignored by the deb_loader tool but necessary to auto-update `deb_packages` rules in a safe manner) will be added later: `pgp_keyid` and `pgp_keyservers` (or maybe `pgp_keyurls`?) - these will contain the keyid and a list of keyservers that might host the defined key which signed the Releases file on the mirror(s). These can't really be hardcoded to support internal mirrors and people that might not want to publish their internal signing keys.

I hope this makes it easier to reason about the changes, the `update_workspace` helper tool will be the next thing to review/add in case you are willing to move forward with this. I can also add a container that will use the rules outlined here and/or convert currently existing rules to this new style in case you want to see how it looks like, though that can be deferred for later.

Changes between this and #94:

* Renamed `architecture` to `arch` to be more in line with existing ruleset
* smaller scope of PR (only rule, no other changes)
* documented adding packages manually and with the `update_workspace` tool
* Mirror URLs now have to point to the root folder of the repository, `/debian` or `/ubuntu` will not be automatically assumed or added (this solves the same issue as #93 and makes it easier to use the snapshots mirror in `deb_packages` rules for example)

I didn't split the attribute (`packages` and `packages_sha256`) since the only mention of Fileset I could find is on the bottom of https://www.bazel.build/support.html with the comment `we do not intend to support them in Bazel, ever`

I also didn't build upon `package_manager` since the whole code for deb_loader is just a few dozen lines and I don't need most of the functionality in `package_manager`. This way it is also possible to have both rule sets side by side, in case you want to make both available.


dlorenc:
Thanks a lot for sending this PR! I'll take a close look soon.

I'm a big fan of the flow you proposed for automatic updating of deb versions via a WORKSPACE updater tool.

One thing that would help is if you could outline the behavioral differences between this and the existing package_manager solution. It looks mostly the same now, with a few minor tweaks. Would it be possible to iterate on the existing ones instead of starting a new set of rules?

MarkusTeufelberger:
The major differences are:

* Package hashes and paths are declared in the WORKSPACE file, not fetched and parsed from a remote server at build(!) time
* Package sources (mirror servers) and package names are combined in the same, single rule (the package_manager solution splits them, because it first needs to fetch the Packages file, parse it to the Packages.json file and only then it knows where to get the actual deb files from)
* The key in the packages/packages_sha256 dict doesn't actually _have_ to be a valid package name (unlike the package_manager ruleset), but it makes it much easier to automatically update them using the `update_workspace` tool if they contain the package name (and optionally the version).
* Any Debian style mirror is supported, not only the snapshots.debian.org one. This includes internal/local mirrors as well as other distributions like Ubuntu.
* If you don't use the updater tool, you can even just use this as an abstraction for http_file() rules (though a rather poor one - I'd much prefer fetching files using Bazel than `urllib.urlretrieve()`).

Some minor differences:

* downloaded deb files are renamed internally after their sha256 hash, not their base64 encoded package name
* You need to enter the full URL to the root of the Debian style mirror, similar to a `sources.list` file - not http://deb.debian.org but http://deb.debian.org/debian/ (the former used to be possible before the current rule set was limited to the snapshots mirror)
* It can properly handle mirrors that serve multiple versions of the same package with the same name (only the tuple `(package_name, version, architecture)` is unique on a mirror)

Similarities:

* This ruleset is intentionally similar to the existing one in how rules look like and are used in BUILD files to make it easy for anyone to switch to these rules. Also I really liked the design of having a dictionary of deb files with their exact package names (and maybe even versions).

Initially I would just have written the updater tool to update http_file() rules, but these would have a HUGE amount of duplicated lines over time compared to this approach and they would have been much harder to reason about.

I consider the "download and parse a Packages.gz file at build time" a design flaw of the existing ruleset, so iterating on it would just lead to the removal of the dpkg_source() rules and adding mirrors, file paths and hashes to dpkg_list() instead. At this point you arrive at the implementation in this PR. The additional  information (arch, distro, distro_type) is not used by the package downloader, but is useful metadata that helps the updater tool figure out which exact package is referenced and (in the future) how to verify the root signature of the whole hosted package repository.

dlorenc:
Thanks for clarifying! Some comments inline:

> The major differences are:
> 
> Package hashes and paths are declared in the WORKSPACE file, not fetched and parsed from a remote server at build(!) time
By paths, are you referring to URLs? Wouldn't using something like http_archive also achieve that goal?

> Package sources (mirror servers) and package names are combined in the same, single rule (the package_manager solution splits them, because it first needs to fetch the Packages file, parse it to the Packages.json file and only then it knows where to get the actual deb files from)

I *think* we could combine these into a single rule with the existing design if we wanted, we just did it this way to be more similar to how debian packages are typically fetched (admins add a set of package lists, apt-get fetches the package from one of the installed lists).

I do like the idea of having fewer rules though :)

> The key in the packages/packages_sha256 dict doesn't actually have to be a valid package name (unlike the package_manager ruleset), but it makes it much easier to automatically update them using the update_workspace tool if they contain the package name (and optionally the version).

This definitely seems like a win. Bazel's naming rules don't match up well with debian packages.

> Any Debian style mirror is supported, not only the snapshots.debian.org one. This includes internal/local mirrors as well as other distributions like Ubuntu.

I'm having a hard time seeing how this will work in practice in combination with bazel/WORKSPACE's assumption about things being mostly hermetic. Making it easy to use any debian-style mirror makes it likely that packages and their SHAs will change out from under you, which makes bazel unhappy.

How do you see this working in practice?

Standard debian mirrors change all the time, and packages appear/disappear. Internally at Google we solve these problems by checking everything we depend on into a single source control system. If you don't want to rely on the snapshot-style mirrors, I think the only other way to satisfy Bazel/WORKSPACE's constraints around reproducibility would be to commit the .debs you need (or at least package list files that reference debs in locations that won't change).

> If you don't use the updater tool, you can even just use this as an abstraction for http_file() rules (though a rather poor one - I'd much prefer fetching files using Bazel than urllib.urlretrieve()).

MarkusTeufelberger:
> By paths, are you referring to URLs? Wouldn't using something like http_archive also achieve that goal?

By paths I refer to the part of the URL after the mirror portion (<mirror URL> + <path> = <deb_file URL>). http_archive is about downloading something with a BUILD file inside, I think you mean http_file. 

Yes, you could use these rules too, but then you'd have a block of mirrors below every single deb file with only the file location changing. There's also the issue of the rule name, which can't map 1:1 with the package name due to name restrictions in Bazel, so it is not really possible to automate this. I guess because of this, the package_manager ruleset was implemented in the first place.

> Also, could you elaborate on the issues with fetching and parsing the packages list file at build time?

Aside from using up ressources on public infrastructure, it:

* externalizes tracking desired state, hashes and versions to a third party (with no SLA by the way)
* does the same work at every build (which - if pinning the hash of the Packages.gz file - has the same result every time)
* makes it very hard to reason about which actual package version is used inside your containers (imagine openssl x.y.z being vulnerable - do you actually use it in one of your containers? If snapshot is down, you'll have to dig around in your Bazel cache and hope the deb file is still there)
* would make the situation worse if you implement signature checking instead of pinning the Packages.gz hash (since then it could change the referred deb file without you noticing with a valid signature)
* is only useful to use Packages.gz pinning on snapshot.debian.org, since on normal mirrors, this file changes even if all deb files you're interested in are still at the same location with the same hash

> I think we could combine these into a single rule with the existing design if we wanted

I wonder how, since there are 3 actions (ensuring the downloader exists, fetch and parse the Packages.gz file into Packages.json, fetch packages using information in the Packages.json file) in the existing design. In my proposed design, the fetching and parsing of Packages.gz file(s) is done out of the loop (similar to the design in Debian, where `apt-get update` and `apt-get upgrade` are distinct actions and there is no actual single command for `update && upgrade`

> This definitely seems like a win. Bazel's naming rules don't match up well with debian packages.

To be fair, the current ruleset also supports naming the referred deb files exactly like the package name, they just have to enforce it (since the name is the key in the Packages.json file).

> How do you see this working in practice?

As I wrote in #94, there are three approaches you can take to make sure that builds of historic commits work as long as possible:

1. (the only supported approach in package_manager): Use snapshot.debian.org at a certain time stamp as one of your mirrors. As long as it is available, your deb file actually is/was public and the mirror works as advertised, this works.
2. Run a local/internal mirror that stores multiple versions per package. Update packages from a public source (e.g. deb.debian.org), drop every package you end up using also on your own mirror and refer to it first, before any other mirror. This is also more friendly towards public infrastructure, since you are caching your files locally. It also is probably faster.
3. Combine both approaches: first enter your local mirror, then (if it fails) the public, up-to-date one (deb.debian.org is load balanced and refers to a MUCH larger infrastructure than snapshot) and if that one also already moved forward, refer to snapshot as the public archive. Once this fails, you still know exactly which file you'd need (full package name and version as well as a sha256 hash) and can start combing backups and caches.

You might also not care that much about being able to build a container from 2 years ago, if you can exactly know what packages were in there to begin with, which currently is a bit hard to do (Which exact openssl version is included in your base container? How long does it take you to find out? More difficult: Without an internet connection and just this repository here?). Currently you just know that some package named e.g. "openjdk-8-jre-headless" was in your container at one point in time. If it turns out to have been malicious because someone at Debian got compromised, it might even get pulled from snapshot.debian.org too.

I don't really understand the comment about hermeticity, as far as I see it, there are no side effects other than a file potentially not being available at multiple locations. Filesystems can fail, servers/mirrors can go down (or no longer serve the same content, which is the same thing in practice) and I'm not sure that Bazel considers this a failure of hermeticity, more of availablity. There are no side effects to downloading a deb file from a mirror. If I had to adapt the existing ruleset to support any mirror (again), this would lead exactly to problems like you described, this is why this PR exists.

This makes it also easier to reason about changes like https://github.com/GoogleCloudPlatform/debian-docker/pull/85 - why is it possible to change the referenced date by nearly three weeks, yet keep the sha256 hash of the Packages file the same? If the file is still the same, the update is not an update at all, if not, there could be an error somewhere and you're depending on unverified deb packages downloaded via http(!). With deb_loader, you'd see exactly if/which packages changed.

dlorenc:
Thanks again for taking the time to walk me through this. I think I'm getting close to understanding the fundamental differences :)

> Aside from using up ressources on public infrastructure, it:

Everything should be cached by bazel for most builds, right? Unless you do a clean --expunge or change the mirror url, we won't be downloading anything afaik.

> externalizes tracking desired state, hashes and versions to a third party (with no SLA by the way)

It's true that the source of truth lives in the third party with this design, but I think that's true for any rules that use a mirror you don't control, right? There's nothing stopping someone that needs a stricter SLA from running their own SNAPSHOT-style mirror, or vendoring the Packages.gz/.debs directly into their source control, though.

> does the same work at every build (which - if pinning the hash of the Packages.gz file - has the same result every time)

Maybe I'm missing something - this is by design. Bazel can very effectively cache build results if nothing changes. If we're repeating work every build, we should fix that :)

> makes it very hard to reason about which actual package version is used inside your containers (imagine openssl x.y.z being vulnerable - do you actually use it in one of your containers? If snapshot is down, you'll have to dig around in your Bazel cache and hope the deb file is still there)

This is definitely an unfortunate problem with package_manager now, I think we're planning to tackle it by including extra metadata inside the image itself though, similar to /var/lib/dpkg/status. @r2d4 has some ideas here. Bazel has some limitations that prevent us from populating a single file like that, so the metadata will likely need to be sharded across several files - one per layer or one per package.

> would make the situation worse if you implement signature checking instead of pinning the Packages.gz hash (since then it could change the referred deb file without you noticing with a valid signature)

Is there any reason we couldn't do both signature and hash pinning?

> is only useful to use Packages.gz pinning on snapshot.debian.org, since on normal mirrors, this file changes even if all deb files you're interested in are still at the same location with the same hash

I think I agree here - these rules are currently only useful with mirrors that preserve packages.gz files and .debs.

> Run a local/internal mirror that stores multiple versions per package. Update packages from a public source (e.g. deb.debian.org), drop every package you end up using also on your own mirror and refer to it first, before any other mirror. This is also more friendly towards public infrastructure, since you are caching your files locally. It also is probably faster.

I think this is the main sticking point I have with my understanding of your design - I don't see this as an actual solution that fits into bazel's constraints around reproducibility. I think this solution works by outsourcing reproducibility to your local mirror.

I think there's a fourth option from your list that also satisfies all of the contraints without depending on snapshot-style mirrors. You would fetch the packages.gz file, vendor it into your source repository and reference it as an ordinary file. When you want to update it, fetch it again and update the file in source control. You would still be depending on the .debs being hosted by another system, but the SHAs and locations are vendored. This is conceptually the same as checking in a packages.json or maven configuration. If you want to take it further, you can vendor all the .debs as well.

We could definitely add some tooling to make this update process easier, similar to the gazelle and update_workspace ideas you proposed earlier.

> I don't really understand the comment about hermeticity, as far as I see it, there are no side effects other than a file potentially not being available at multiple locations. Filesystems can fail, servers/mirrors can go down (or no longer serve the same content, which is the same thing in practice) and I'm not sure that Bazel considers this a failure of hermeticity, more of availablity. There are no side effects to downloading a deb file from a mirror. If I had to adapt the existing ruleset to support any mirror (again), this would lead exactly to problems like you described, this is why this PR exists.

Yeah, hermeticity was the wrong word here. The main point is that these rules need to either result in the same exact output every time they are run or fail.

I think we have some more options on making the existing ruleset a bit more user friendly without scrapping them and rewriting. I'm in favor of making vendoring Packages.gz files easier, and thinking of ways to relax the constraint we have now on pinning the entire packages.gz file.

dlorenc:
> This makes it also easier to reason about changes like GoogleCloudPlatform/debian-docker#85 - why is it possible to change the referenced date by nearly three weeks, yet keep the sha256 hash of the Packages file the same? If the file is still the same, the update is not an update at all, if not, there could be an error somewhere and you're depending on unverified deb packages downloaded via http(!). With deb_loader, you'd see exactly if/which packages changed.

One other question - I'm not sure what you mean here. That repo doesn't actually use package_manager, it uses normal apt-get today. What Packages file are you referring to as not having changed? The sha of the resulting image, and the versions of the debian packages inside changed after updating the SNAPSHOT variable.

MarkusTeufelberger:
> There's nothing stopping someone that needs a stricter SLA from running their own SNAPSHOT-style mirror, or vendoring the Packages.gz/.debs directly into their source control, though.

The "snapshot style mirror" is not a generally supported use case and I am not so sure if there is software out there that supports a layout like this. Vendoring binaries is frowned upon, but would be a generally valid solution for this problem, yes.

> Is there any reason we couldn't do both signature and hash pinning?

The signature pinning would be useless in that case (as you already know which file you will receive). Of course you _can_ check them, but in general you should be checking signatures if you get files you haven't encountered yet and hashes of files you have already verified to be ok. Hashing a file is a stronger guarantee that it will be safe (as you have already had it 100% locally before hashing it), verifying a signature of a file that contains hashes of other files is only as good as the trust you put into the person that signed this.

> I think this solution works by outsourcing reproducibility to your local mirror.

You mean availability, right? Since files are hashed, the resulting container will always be reproducibly built - as long as the input files are available.

> I think there's a fourth option from your list that also satisfies all of the contraints without depending on snapshot-style mirrors. [...]

Yes this would be a valid option with the `package_manager` ruleset.

BUT: The only information you actually need these files for, is to obtain a trusted record of the sha256 hash and the path of a `(package, version)` tuple. Why drag these files along (including several megabytes of data that you first parse and then will just discard) and version them, just to get these 2 values per deb file? You can just enter them in the WORKSPACE file and save an extra build step, several MB of data per update in your repository and get more flexibility too (the proposed ruleset can already handle versions, the existing one can't and needs quite some work to support this). It also makes the data sent to evaluate the `dpkg_list` rule (or its equivalent, `deb_packages`) shrink from a few MB of JSON to a few lines containing only the information that is actually needed to fetch the files.

> One other question - I'm not sure what you mean here. That repo doesn't actually use package_manager, it uses normal apt-get today. What Packages file are you referring to as not having changed? The sha of the resulting image, and the versions of the debian packages inside changed after updating the SNAPSHOT variable.

https://github.com/GoogleCloudPlatform/debian-docker/pull/85/commits/570df208b79d11e674ecc10bccff3fe17ba49813 changed the date of the snapshot mirror. This means you are no longer referencing http://snapshot.debian.org/archive/debian/20170816T214423Z/dists/jessie/main/binary-amd64/Packages.gz (SHA256: 142cceae78a1343e66a0d27f1b142c406243d7940f626972c2c39ef71499ce61) but http://snapshot.debian.org/archive/debian/20170906T160441Z/dists/jessie/main/binary-amd64/Packages.gz (same SHA256 hash). This file hasn't actually changed since July 22nd 2017. This also means not a single package in Debian Jessie has changed over the summer (so much for fast changing files on official mirros... :D).

Since the hashes are the same, the whole PR was good for nothing (other than removing a line in the README in an unrelated change). It did not "switch to the newest snapshot", it did just change the URL where your deb files are fetched from and I'm not so sure if Bazel actually could/did properly detect that it already had these files in its cache - after all the inputs to the downloader tool changed and it has no idea what's going on in that downloader tool.

I really hope that your "builder" containers did NOT change a single bit when you changed that variable. Also at least your Jessie containers also should be bit-identical, otherwise you have at least a reproducibility problem.

dlorenc:
> Since the hashes are the same, the whole PR was good for nothing (other than removing a line in the README in an unrelated change). It did not "switch to the newest snapshot", it did just change the URL where your deb files are fetched from and I'm not so sure if Bazel actually could/did properly detect that it already had these files in its cache - after all the inputs to the downloader tool changed and it has no idea what's going on in that downloader tool.

Ah, I see what you mean now. Yeah, had anything changed the build would have failed unless we updated the SHA as well. This code isn't actually live yet though, we're still switching over to it from the old, non-reproducible deboostrap system. That PR was a test of some build hooks we're working on internally :)

No images have actually been built or released using those debian rules yet. Hopefully soon though.

MarkusTeufelberger:
Alright, it's been about 3 weeks for this PR now, over 2 weeks without additional comments or questions and I hope you had enough time to discuss this internally. I would really like to move forward either way with this.

@dlorenc, @mattmoor, @r2d4:

1. Do you want to remove the "create a Package.json file from the Packages.gz" stage from each build or not? If yes, I guess you'll reach a state close to this PR sooner or later and I'd love to bring it up to your standards, if not, I'd like to know why and which advantages you see with that approach that wouldn't be possible with my proposal.

2. Would you be OK with adding this rule set as an alternative to the existing one into this repository, in case you want to take the "wait and see"/"let the user decide" approach or do you only want to support a single mechanism for defining and retrieving deb files in `distroless`? I can think of advantages and disadvantages for that approach of course, so I'd love to know your opinion.

3. Thanks a lot for the great work already, especially on rules_docker and the existing helper tool for distroless. :-)

MarkusTeufelberger:
This ruleset is a superset of the one you are currently using (it can also handle your current use case with the snapshots mirror). It would be easy to test this in the future with the same caveats/guarantees as the current ruleset (if snapshots is down, you have to provide alternative deb sources with my rules and you have to provide alternative Packages.gz and deb sources with the current rules).

If you are interested, I can convert the existing containers and rules to the ones in here, so you can see what I mean (in the end the mirror URL would be "http://snapshot.debian.org/archive/debian/20170821T035341Z/" instead of a dynamic one like "http://deb.debian.org/debian/").

The reason that I want to merge instead of fork is mainly that I would strongly prefer not to maintain a full fork of this including CI, hosting binaries of the helper tool, converting fixes you are adding (e.g the "/etc/os-release" file), new example base containers etc. just because the method of defining and fetching the exact same deb files is improved.

There will be a small additional commit in here (`pgp_keyid` and `pgp_keyurls` fields which already work in the rewritten version of the update helper and probably porting the `/etc/os-release` code), but all in all the rule set in here is mostly complete. Of course I'll go ahead with it anyways if you really don't like this ruleset in this PR, but I'd really like to share the work and do my best to convince you that this approach has its merits while removing quite some risks and downsides yours currently has.

MarkusTeufelberger:
Alright, since there doesn't seem to be much interest in merging this upstream, I'll close this PR at the end of next week.

I prepared a branch at https://github.com/mgit-at/distroless/tree/deb_packages_and_updater that I'll try to regularly either rebase onto master or at least make sure it can be merged cleanly onto master.

It contains the rewritten updater tool, this ruleset from the PR here and as examples _all_ containers from this repository also as a build and test target using this ruleset. I didn't replace the existing rules/BUILD files on purpose, as I don't really want to manually deal with merge conflicts all the time.
I also uploaded the deb_file_loader tool to a public location, so travis should feel more comfortable with testing this.

If there is interest in still merging this here, I'm all up for it, the only changes necessary would be to replace "deb.debian.org" as a mirror with the snapshot one (since building historic versions from public sources seems to be a stronger focus for you) and deciding which ruleset that you want to use (I'm not sure if it makes sense to build everything with both rulesets).

I chose to replace `pgp_keyid` and `pgp_keyurls` with a single `pgp_key` field as well as a mandatory `http_file` rule that points to the actual key. The reason is that golang (the language of the helper tool) only support 64bit pgp keyids, which is a bit short, while it is possible to define a full sha256 hash of the key that way.

MarkusTeufelberger:
After thinking about it a bit more, I rewrote the functionality of the Python tool in here in skylark to use Bazel's https://docs.bazel.build/versions/master/skylark/lib/repository_ctx.html#download function. This also removes the `deb_repositories` rule.

After all, all it does is download a few files with known paths from one or more mirrors, check hashes and create a few files with contents that are also known when looking at the inputs to this rule. All in all it is mostly a nicer abstraction for `http_file` anyways, since there is no need to deal with `Packages.gz` files (handling this is externalized to the updater tool).

MarkusTeufelberger:
I can also offer to include this in https://github.com/bazelbuild/rules_docker, if you think it might fit better there. After all it is closely related to the docker_build/container_image rules. @mattmoor

mattmoor:
@MarkusTeufelberger @dlorenc @r2d4 Ultimately there may be a better home for these than this repo.

I had a good conversation with @damienmg @dslomov @davidstanke on Friday, which may be relevant to the core of the discussion here.  Essentially what it boils down to is `WORKSPACE` vs. `WORKSPACE.lock`.

I think that we want the capacity for users to express dependencies in similar terms to what they use with `apt-get install` where they simply have a set of mirrors and a package name (maybe expressing a particular version).  However, this is inadequate for reproducible builds because it drifts.  Pinning to a particular package resolution is adequate for reproducible builds, but will often break because packages disappear as the resolution starts pointing to something new.

Here's a strawman:
1. Users express their desire at a very high-level in `WORKSPACE`
1. Bazel can resolve `WORKSPACE` to `WORKSPACE.lock`, which fetches exact deps with checksums.
1. Bazel reads `WORKSPACE.lock` if present, otherwise it resolves it from `WORKSPACE`

There are two workflow variations on top of this that I would recommend:
1. Only `WORKSPACE` is committed to `master`, but `WORKSPACE.lock` is checked into release branches.  This keeps `master` from breaking often, but gives deterministic release builds.
1. Both `WORKSPACE` and `WORKSPACE.lock` are committed to `master`.  This keeps all builds deterministic, but makes HEAD more susceptible to breaks.

In both cases, producing `WORKSPACE.lock` should involve a single Bazel command.


I think it's relevant here because IIUC we've adopted snapshot mirrors to get something closer to the first workflow (stable HEAD) without the repo churn.  However, I think @MarkusTeufelberger wants something closer to the second workflow.


Am I understanding the nuances of the conversation correctly, or am I horribly mischaracterizing something?  Do you think that given this kind of facility in Bazel that we could potentially agree on a common definition and just have varied workflows for how we achieve our respective workflows?

MarkusTeufelberger:
I personally don't see the need for a `WORKSPACE.lock` file, as I already wrote elsewhere:

If you want to lock down exact versions and only state general locations in `WORKSPACE` (e.g. `WORKSPACE` would contain only the git repository location, `WORKSPACE.lock` would contain commit IDs or tags), in the end the only thing that you could accomplish is to strip out everything that would usually change when updating a `WORKSPACE` file into a `WORKSPACE.lock` file. To understand what is going on, one still would need to look at both files.

A way to implement your design and experiment with its merits and drawbacks right now would be to write a rules file for `buildozer` that reads some keys (package version, git commits...) from a `lock` file and the basic structure from a `WORKSPACE` template and generates a `WORKSPACE` file from that on the fly which is not checked into version control. This rules file also could be generated via a script to get updated.

In your second scenario "HEAD" only breaks if links become invalid. There are ways to make sure this doesn't happen - using snapshot mirrors, internal artifact storages/mirrors... This is an orthogonal issue, not related to this rule set or approach. As I already wrote, if you care about having build inputs available, it is easily possible to just specify the snapshot mirror at a certain time as a mirror in the rules.

My argument is that currently `WORKSPACE` files typically contain all information that is necessary to assess which external software in which version etc. is being used in a build. Extending Bazel to optionally allow "latest" versions (e.g. all git rules fetch from HEAD, specialized rules like the deb packages rules in distroless fetch from an up-to-date mirror...) for some or all rules might sound interesting for some use cases, but is not useful enough to be implemented as a main pattern.

Here is my desired use case for Bazel in general:

1. `WORKSPACE` files contain the most exact possible description of every input to a build process for the software described in the repository.
2. Every build of a freshly checked out repository must create the exact same outputs every time, as long as all inputs are available. If they are not, the build must fail, not try to do some magic trying to find a newer/compatible version or whatever else.
3. If one (optionally) wants to build with different inputs, the `WORKSPACE` file must be changed before executing `bazel build ...` - in the case that more recent versions or other, more dynamic things are required, this is a trivial task (since the more exact, locked down version is already in the `WORKSPACE` file, so one just needs to transform that one to a different one)

The current approach in this repository violates rule 1 a bit (it is not clear which packages come from backports for example).
Rule 2 currently is fulfilled because you trust the snapshot mirror, not because your rules are written in a way that supports this use case.
Rule 3 is kinda fulfilled, it is only possible to change everything at once by selecting a different snapshot, not to change different packages. This is because of your implicit dependency on the `Packages.gz` file (which then guides every further decision down the line)

Instead of splitting `WORKSPACE`, it is or could be possible to tell bazel that some or all rules in `WORKSPACE` also support an "unlocked" mode, which would be the opposite of your proposal (locked down by default, unlocked on demand, for canary builds for example). I don't see the value in having default unlocked builds, as you already stated that `master` would break with the unlocked approach from time to time and would require locking for releases at least.

I value bazel for its ability to exactly specify build inputs as exactly as possible and while Google might not have this problem with having to lock down external dependencies (since you don't have any in a monorepo), this is a big problem in the general IT landscape.

Currently it is extremely hard to know which actual deb files are put into which distroless container. A `WORKSPACE.lock` file won't help with your current approach (incorrectly parsing Debian archive files, generating a json file and reading everything from there while depending on an external service to provide the input file which you mostly throw away anyways - and no automated way to update to a trusted current mirror state) and imho breaks some things and assumptions that make bazel great. It would be nice to know what @damienmg @dslomov and @davidstanke had to say on this matter or if you have a more concrete design proposal.

As an aside: `apt-get install ...` only will resolve the local state and not fetch any updates from a `Package` file on a mirror - that's the job of `apt-get update`. This makes `apt-get install ...` deterministic and is the reason why it was implemented that way in Debian (the `WORKSPACE.lock` file equivalent there is the apt cache). `apt-get install ...` NEVER drifts unless users willingly update the apt cache with a different command. My design reflects this by splitting the process (`update_workspace` tool = `apt-get update && apt-get dist-upgrade`, `bazel build` = `apt-get install`), the current one is a mixture of fetching and parsing `Package` files (`apt-get update`) and fetching the actual deb file (`apt-get install`), all in one single step, currently _only_ secured by the person that checked in the SHA256 hash of the `Packages` file, added after complaints about the procedure in https://github.com/GoogleCloudPlatform/distroless/issues/72.

This is an issue for me and one reason I still continue to push for changes in this process is that images produced using this process are used upstream in the https://github.com/bazelbuild/rules_docker repository as default base images. all while not even clearly documenting which of your packages are from which repository (`jessie` vs. `jessie-backports`, no versions documented even in comments...).

Lastly I'd still like to know answers to the questions I asked above: Do you want to continue with the `Package.json` approach and why? Would you be OK with merging this one, would you prefer having this one replace yours or would you prefer a fork? Your answer at least sounds like the latter for the second question. You are also often committing in https://github.com/bazelbuild/rules_docker though, which would be the other logical place for this rule upstream but I'm not so sure if I should submit it there.

Tl;dr: It is trivial to change a `git_repository` rule in a  `WORKSPACE` file to use the current HEAD before building, it does not help anyone to only record a repository location in a `WORKSPACE` file if the commit to be used is not specified in a `WORKSPACE.lock` file, as this will break builds sooner or later and not be reproducible (similar for deb package fetching rules). This means that the `lock` files are required anyways and could be merged into the `WORKSPACE` file, creating exactly the situation we already have.
I'm all for (optional) builds with "unlocked" `WORKSPACE` rules, I am strongly against having these as default.

damienmg:
I have a design that summarize our discussion, I'll share it wisely
tomorrow when everybody involved had time to take a quick look.

MarkusTeufelberger:
Thanks! :+1:

dlorenc:
We've talked a bit about putting these rules (or different versions of them) into a repo in the bazel org. Maybe rules_debian or rules_dpkg? Or maybe include rules for other package managers there?

Like I said earlier, I'd prefer to keep the design we have for now in distroless. We're not really planning to use these rules too widely for now, and what we have meets our use-case. The bazel org is the right place for rules that will be publicly shared, we're mainly using this repo to host the images and samples.

I'm definitely interested in seeing the new design, @damienmg!

MarkusTeufelberger:
Great thanks, trying to set up a repository like that would have been my next way of action anyways. :-)

Maybe it would even be useful to add the one(s) from here to something like `rules_pkg_files` as well and only maintain BUILD files in `distroless` so you really just host images and samples? Anyways, thanks for responding and working on solutions.

mattmoor:
@MarkusTeufelberger Are you by chance coming to the Bazel conference?  If so, that might be a good place to discuss some of this face-to-face?

MarkusTeufelberger:
I'd love too, really, but it is a bit far (and expensive) from here unfortunately. I'll try to come to the European version though.

mattmoor:
@damienmg Can you create an issue to track this on the main repo, so that we can capture the discussion in a more appropriate forum?  This has become a huge tangent here, for which I apologize.


@MarkusTeufelberger all of your responses further convince me we're after the same thing, so clearly I'm not articulating my goal well. :)

> it does not help anyone to only record a repository location in a WORKSPACE file if the commit to be used is not specified in a WORKSPACE.lock file

e.g. here I would fully expect `WORKSPACE` => `WORKSPACE.lock` to add a `commit` attribute.  @damienmg's example has an example of how we might achieve this IIRC.

Let's create the umbrella issue and continue the discussion there.  It may make sense to set up a video call to discuss this further.

MarkusTeufelberger:
Agreed, I think we are after very similar things and only disagreeing about some semantics (e.g. which file should be the one where some specific info like a file hash or commit ID should be stored).

I'm on the same time zone as the Munich office with regards to the video call.

Back on topic, I'll set up a repository (and come up with a good name for it) asap to contain at least the rule and tool from here, maybe (if you want to) also the `dpkg_*` ones from distroless itself.

dlorenc:
To keep the tangent going, I've spent some time looking into the Ubuntu mirrors, and am a bit worried that neither of these designs will work well there.

I'll open an issue with some thoughts on how we might support Ubuntu.

mattmoor:
@dlorenc I don't think that's the tangent :)

@MarkusTeufelberger @damienmg We might also consider splitting `pkg_tar` off of Bazel Core while we're at it (this contains the logic to expand `deb`s into a `tar`.

MarkusTeufelberger:
@dlorenc, Ubuntu mirrors work out of the box with my rules, though there is no "snapshot" mirror I am aware of. What excatly didn't work, design wise? `deb_packages` rules are more or less aliased `http_file` rules, so I don't see an immediate issue.

dlorenc:
Yeah, they have no snapshot and aggressively delete old versions of packages. Any WORKSPACE that pins to full shas of Ubuntu packages will be broken very frequently.

This might just be "working as intended", though...

MarkusTeufelberger:
I would say that's by design, the snapshot mirror is the exceptional case...

dlorenc:
Yeah, the downside is that you won't be able to reproduce any old builds without running your own mirror.

MarkusTeufelberger:
...or hosting the deb files elsewhere, yes. At least you'd know exactly which deb file with which name and which SHA256 hash you'd need, with the `dpkg_*` rules you'd need to archive the `Packages` file too.

damienmg:
So I propose we move the discussion to the design doc: https://docs.google.com/document/d/1HfRGRW4MwnVUG24rw3HJIBkdEYfXlESnlKOboO97A3A/

For moving the rules I am up for that, want to create rules_pkg to contains RPM, DEB and so on for packaging rules.

MarkusTeufelberger:
I created https://github.com/mgit-at/rules_pkg (with the boilerplate files like LICENSE, README.md etc. missing) and added a smallish CI script for Travis-ci. Is this kinda what you have in mind, @damienmg ? If yes, as soon as `bazelbuild/rules_pkg` is available, I would open a pull request.

damienmg:
Sounds good, I'll try to bootstrap that soonish, I'll move docker / pkg
rules outside of bazel_tools definitely with that but I cannot guarantee
any ETA.

MarkusTeufelberger:
Closing in favor of rules_pkg. Would be great to have the repo in the bazelbuild org soonish, it is not that urgent to immediately consolidate all rules in there...

MarkusTeufelberger:
ping @damienmg  - it would be nice to get started on bazelbuild/rules_pkg. :-)

damienmg:
Sorry I have a hundred things on my plate, I cannot guarantee any ETA for
this.

MarkusTeufelberger:
Maybe there's someone else that you could delegate the repo creation to? It is not extremely urgent of course, I just didn't want to let it slip completely out of sight. I'll ping again in a while, if that's ok.

MarkusTeufelberger:
Another sneaky ping @damienmg - I hope the conference was great! :-)

damienmg:
The conference was great :) Unfortunately I have still no visibility on
doing that.

MarkusTeufelberger:
And according to the mailing list you're gone from bazel! :-(

Hope your new job is great, whom should I contact about getting this repo created?

davidstanke:
Hey hi! Apologies that this got lost in the shuffle. I'll pick this up now. My only nit is this: AFAICT, these rules are specific to debian packages, yes? Given that, I believe the name `rules_pkg` is too generic.  Can we do `rules_dpkg`? 

Or something else that specifically references Debian, to preclude any possible confusion with other packaging mechanisms? Thanks!

MarkusTeufelberger:
The idea was to move several packaging rules (including dpkg, deb, tar... maybe yum or rpm in the future?) into a common repo as far as I remember.

davidstanke:
Cool, in that case, I'm good with it. Will proceed with the repo... when
you push to it, can you include a README that speaks to those future plans?

MarkusTeufelberger:
Of course. I would also love to see the current rules + helper tool of distroless go in there, since right now this repository is a mix of production BUILD files, bzl rules and some more general examples.

davidstanke:
Update: I've kicked off the process to create the repo (github.com/bazelbuild/rules_pkg). It takes a few days for Google to review the request, so bear with us. :)

davidstanke:
Okay, the repo is live: https://github.com/bazelbuild/rules_pkg

Can you make a PR there with the initial content?

MarkusTeufelberger:
Awesome, I'm currently working on it and hope to be done tomorrow. :+1:

PR #102, opened 2017-09-15
Add /etc/os-release file generation package rules

r2d4:
Adds another output file to the dpkg_src rule with os-release-like
metadata about the distribution

Add a structure test validation for /etc/os-release in the base image

### DIFF
--- a/package_manager/util.py
+++ b/package_manager/util.py
@@ -1,5 +1,8 @@
 import hashlib
 import base64
+import collections
+import tarfile
+import os
 
 def sha256_checksum(filename, block_size=65536):
     sha256 = hashlib.sha256()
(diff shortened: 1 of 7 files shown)


package_manager/util.py
```diff
@@ -13,3 +16,45 @@ def package_to_rule(workspace_name, s):
...
+
+def generate_os_release(distro, os_release_file):
+    """ Generates an /etc/os-release like file with information about the
+    package distribution.  VERSION and VERSION_ID are left unset if the package
+    source is from an unknown debian release.
+    """
+
+    os_release = collections.OrderedDict([
+        ("PRETTY_NAME", "Distroless"),
+        ("NAME", "Debian GNU/Linux"),
```
mattmoor: Have you put any thought into how we'd support an Ubuntu feed?

r2d4: I think we could (and probably should) definitely loosen the parameterization and the URL parsing of the sources list and accept arbitrary sources, although in that case we'd have to take the user at their word on what distribution and version the source list could correspond to (which is probably fine).

Then we'd still have to have some hardcoded metadata about how the ubuntu os-release looks but it could follow the same pattern.

dlorenc: Yeah, this basically sounds right to me. We hardcode debian now later on which would need to chane first.

PR #103, opened 2017-09-21
Add packae tzdata to distroless/base.

### LINKED ISSUE #95: request to have debian package tzdata in distroless/base
Not sure if other language runtimes use this, but the Go time package does.  It would be nice if at least the directory /usr/share/zoneinfo and files under it are packaged into distroless/base.

dlorenc:
Fixes #95

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -45,6 +45,7 @@ dpkg_list(
         "openssl",
         "libssl1.0.0",
         "netbase",
+        "tzdata",
 
         #java
         "zlib1g",
--- a/base/BUILD
+++ b/base/BUILD
@@ -34,6 +34,7 @@ docker_build(
         packages["libssl1.0.0"],
         packages["openssl"],
         packages["netbase"],
+        packages["tzdata"],
     ],
     tars = [
         ":base_passwd.passwd.tar",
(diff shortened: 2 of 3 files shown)


cybrcodr:
Thanks much!  Do you know when GCR will have an image with this change?

Also, seems like the GCR image locations for this image does not have a timestamps on them --

```
$ gcloud beta container images list-tags gcr.io/distroless/base --format=json
[
  {
    "digest": "sha256:616ea72e69c65a89f8e0622010578654bec060e1f7bbc27ffd239555ca781180",
    "tags": [],
    "timestamp": null
  },

<snip>
  {
    "digest": "sha256:782777a1e7465f7b6c3065fc69e829db5f731c1c7ba7ece7739140015ca918bd",
    "tags": [
      "latest"
    ],
    "timestamp": null
  }
]
```

dlorenc:
The images should be up to date on GCR.

> Also, seems like the GCR image locations for this image does not have a timestamps on them --

The no timestamps is somewhat intentional - even though it causes issues for gcloud. The timestamps are unfortunately part of the metadata that gets hashed into the digest, so the same contents with different timestamps will look like different images :(

We'll need to set something up that publishes datetime based tags in addition to the digests to help users figure out which image is latest without using the latest tag.

mattmoor:
@cybrcodr container-diff won't work with these images yet.

cybrcodr:
Thanks for the info.  We'll go ahead and try it out then.

PR #106, opened 2017-09-27
Add a rule to generate java cacerts files using a debian docker image.

dlorenc:
This isn't ideal, but works.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -126,6 +126,7 @@ load(
 # Used to generate java ca certs.
 docker_pull(
     name = "debian8",
+    # From tag: 2017-09-11-115552
     digest = "sha256:6d381d0bf292e31291136cff03b3209eb40ef6342fb790483fa1b9d3af84ae46",
     registry = "gcr.io",
     repository = "google-appengine/debian8",
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -120,6 +120,15 @@ git_repository(
 load(
     "@io_bazel_rules_docker//docker:docker.bzl",
     "docker_repositories",
+    "docker_pull",
+)
+
+# Used to generate java ca certs.
+docker_pull(
+    name = "debian8",
+    digest = "sha256:6d381d0bf292e31291136cff03b3209eb40ef6342fb790483fa1b9d3af84ae46",
+    registry = "gcr.io",
+    repository = "google-appengine/debian8",
 )
 
 docker_repositories()
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -113,7 +113,7 @@ http_file(
 # Docker rules.
 git_repository(
     name = "io_bazel_rules_docker",
-    commit = "839a297d4e874216b4fd93f09dd35be5592dc10e",
+    commit = "cdd259b3ba67fd4ef814c88070a2ebc7bec28dc5",
     remote = "https://github.com/bazelbuild/rules_docker.git",
 )
 
--- a/cacerts/java.bzl
+++ b/cacerts/java.bzl
@@ -1,0 +1,71 @@
+# Copyright 2017 Google Inc. All rights reserved.
+
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+
+#     http://www.apache.org/licenses/LICENSE-2.0
+
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Rule for building and extracting java ca-certificates inside of a docker image."""
+
+def _impl(ctx):
+    # Strip off the '.tar'
+    image_name = ctx.attr._builder_image.label.name.split('.', 1)[0]
+    # docker_build rules always generate an image named 'bazel/$package:$name'.
+    builder_image_name = "bazel/%s:%s" % (ctx.attr._builder_image.label.package,
+                                          image_name)
+
+    # Generate a shell script to run the build.
+    build_contents = """\
+#!/bin/bash
+set -ex
+docker load -i {0}
+# Install the certs in the builder image.
+cid=$(docker run -d {1} sh -c "apt-get update && apt-get install -y -q ca-certificates-java")
+docker attach $cid
+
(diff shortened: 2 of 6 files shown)


cacerts/java.bzl
```diff
@@ -0,0 +1,71 @@
...
+
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Rule for building and extracting java ca-certificates inside of a docker image."""
+
+def _impl(ctx):
```
dlorenc: I'd like to come back to my old idea of a "docker_run" rule that supports this general pattern of "run a certain command in a docker image, and extract some paths as outputs. We could use it here and in a few other cases.

mattmoor: Holy package list, Batman....
```
  ca-certificates dbus fontconfig-config fonts-dejavu-core java-common
  krb5-locales libalgorithm-c3-perl libarchive-extract-perl libasyncns0
  libavahi-client3 libavahi-common-data libavahi-common3 libcap-ng0
  libcgi-fast-perl libcgi-pm-perl libclass-c3-perl libclass-c3-xs-perl
  libcpan-meta-perl libcups2 libdata-optlist-perl libdata-section-perl
  libdbus-1-3 libexpat1 libfcgi-perl libffi6 libflac8 libfontconfig1
  libfreetype6 libgdbm3 libglib2.0-0 libglib2.0-data libgmp10
  libgnutls-deb0-28 libgssapi-krb5-2 libhogweed2 libice6 libjpeg62-turbo
  libjson-c2 libk5crypto3 libkeyutils1 libkrb5-3 libkrb5support0 liblcms2-2
  liblog-message-perl liblog-message-simple-perl libmodule-build-perl
  libmodule-pluggable-perl libmodule-signature-perl libmro-compat-perl
  libnettle4 libnspr4 libnss3 libogg0 libp11-kit0 libpackage-constants-perl
  libparams-util-perl libpcsclite1 libpng12-0 libpod-latex-perl
  libpod-readme-perl libpulse0 libregexp-common-perl libsctp1 libsm6
  libsndfile1 libsoftware-license-perl libsqlite3-0 libssl1.0.0
  libsub-exporter-perl libsub-install-perl libtasn1-6 libterm-ui-perl
  libtext-soundex-perl libtext-template-perl libvorbis0a libvorbisenc2
  libwrap0 libx11-6 libx11-data libx11-xcb1 libxau6 libxcb1 libxdmcp6 libxext6
  libxi6 libxml2 libxtst6 lksctp-tools openjdk-7-jre-headless openssl perl
  perl-base perl-modules rename sgml-base shared-mime-info tcpd tzdata-java
  ucf x11-common xdg-user-dirs xml-core
```

OMG.

cacerts/java.bzl
```diff
@@ -0,0 +1,71 @@
...
+        ctx.attr._builder_image.data_runfiles.files.to_list() + ctx.attr._builder_image.default_runfiles.files.to_list(),
+        executable=script,
+    )
+
+    return struct()
+
+cacerts_java = rule(
+    attrs = {
+        "_builder_image": attr.label(
+            default = Label("@debian8//image:image.tar"),
```
mattmoor: Technically `docker_pull` should support incremental loading too, so you might get better performance by dropping the tar here and invoking the executable target above.  It's overkill for this, but is probably how we should handle the image input in a more generalized `docker_run`.

cacerts/java.bzl
mattmoor: I think what you are describing is effectively `genrule` where the command is invoked within a specified container image; I think we should strive to keep the signature as close to `genrule` as possible if we head this route (I wonder if we'd want dicts for the inputs / outputs?).  It would certainly make a useful building block for this.

WORKSPACE
```diff
@@ -120,6 +120,15 @@ git_repository(
 load(
     "@io_bazel_rules_docker//docker:docker.bzl",
     "docker_repositories",
+    "docker_pull",
+)
+
+# Used to generate java ca certs.
+docker_pull(
+    name = "debian8",
+    digest = "sha256:6d381d0bf292e31291136cff03b3209eb40ef6342fb790483fa1b9d3af84ae46",
```
mattmoor: I've started tagging these with comments indicating what tag I got them from and when.

java/BUILD
```diff
@@ -2,6 +2,11 @@ package(default_visibility = ["//visibility:public"])
 
 load("@io_bazel_rules_docker//docker:docker.bzl", "docker_build")
 load("@package_bundle//file:packages.bzl", "packages")
+load("//cacerts:java.bzl", "cacerts_java")
+
+cacerts_java(
+    name = "cacerts_java.tar",
```
mattmoor: Given that `out` is `%{name}.tar` I think the `tar` here is redundant because the actual `tar` is going to get named `cacerts_java.tar.tar` :)

java/testdata/java.yaml
```diff
@@ -3,3 +3,7 @@ commandTests:
   - name: java
     command: ["/usr/lib/jvm/java-8-openjdk-amd64/jre/bin/java", "-version"]
     expectedError: ['openjdk version "1\.8.*"']
+fileExistenceTests:
+  - name: certs
+    path: "/etc/ssl/certs/java/cacerts"
+    shouldExist: true
```
mattmoor: Is it possible to have an app test this?  I'm not 100% clear on what this is needed for, but if it's just outbound SSL maybe hitting something that never goes down like `https://www.google.com`? :)

cacerts/java.bzl
dlorenc: Yeah, that's basically what I was thinking. I have a few other places I think we'll need to implement a rule like this, and a docker_genrule would definitely help.

cacerts/java.bzl
dlorenc: I've tried this a few times but I've never actually gotten all the runfiles to work. Maybe your refactors have fixed things.

I'll send another PR soon that tries to do that and see if you can help me debug. I think the issue is that the loader script wasn't referencing it's runfiles completely correctly, so when it gets consumed by another rule it explodes.

mattmoor: I've recently gained an understanding of what this takes, so happy to help.

java/testdata/java.yaml
dlorenc: Yep, added one that does exactly that.

PR #110, opened 2017-10-07
Add debug variants of all containers that include busybox.

mattmoor:
Switch from publishing things through `cloudbuild.yaml` explicitly to creating a bundle and publishing all of the images within.


mattmoor:
As context, I'd like to enable using the debug container for debug builds of `lang_image` rules.

mattmoor:
No clue why travis failed, I'll kick it.

mattmoor:
@dlorenc Do you have any idea why this might be failing?

mattmoor:
[This PR](https://github.com/bazelbuild/rules_docker/pull/184) shows how I plan to make use of these.

r2d4:
Looks like buildifier errors, not sure why the error message didnt show up on travis though.

mattmoor:
Alright, `buildifier` issues fixed.  It is taking the first `exit 1` in `buildifier.sh`, which is why there is no output.

mattmoor:
@r2d4 This should be RFAL if you want to do the honors?

cc/BUILD
```diff
@@ -4,11 +4,14 @@ load("@io_bazel_rules_docker//docker:docker.bzl", "docker_build")
 load("@package_bundle//file:packages.bzl", "packages")
 
 # An intermediate image for Java and other "mostly statically" compiled languages
-docker_build(
-    name = "cc",
-    base = "//base:base",
+[docker_build(
+    name = "cc" if not mode else mode[1:],
```
r2d4: Does it make more sense to iterate the other way?
```
docker_build(
  name = image + ":debug"
  base = "//" + image
  ...
  directory = "/",
  entrypoint = ["/busybox/sh"],
  env = {"PATH": "/busybox"},
  tars = ["//busybox:busybox.tar"],
...
for image in [
   "cc",
   "java",
   "python",
   ...
```

mattmoor: Name can't have a `:` in it.  It also can't have a `/` so things start getting awkward for `java/jetty:debug`.  I also like the symmetry of having the tags match the BUILD labels, but I'm certainly not married to it given another suggestion.

mattmoor: Things are further complicated by the lack of symmetry for targets like `java:java8` and `python2.7:python27` :(

r2d4: That make sense.  Just wondering if there was a way we could do it without writing the code for each image.  LGTM

mattmoor: I was debating writing a little skylark macro for it, but the quirk in my second comment made that hard enough to stop me.  If it were just `//foo` and `//foo:debug` everywhere I probably would have.  I don't recall the history of why we let these diverge...

mattmoor:
Ok, I'm gonna merge to unblock https://github.com/bazelbuild/rules_docker/pull/184

PR #113, opened 2017-10-09
Invoke the busybox we're including in the image instead of the system…

mattmoor:
…-installed copy.


mattmoor:
With this, a cloud build in my project gives:
```shell
$ docker run -ti --rm --entrypoint=sh gcr.io/<my-project>/java:debug -c "echo Hi"
Hi
```

PR #115, opened 2017-10-10
This corrects the PATH in debug images.

mattmoor:
Omitting PATH seems to make Docker default to a typical value, but when we add `busybox` we block this.

This gives `//base` the typical value, and makes `//base:debug` extend it to include `/busybox`.


mattmoor:
This is causing the `py_image` testing to fail when I attempt to use it.

mattmoor:
This also seems to have caught a single bad typo in my PR! :)

PR #119, opened 2017-10-21
Explicitly list out the commands for busybox.

### LINKED ISSUE #117: Genrule for generating busybox.tar is broken on non Linux
I'm trying to build a docker image on my mac laptop and push it to a locally running linux VM, where I will run it. When trying to build //base:debug and run it I found that it fails to run because /busybox/sh doesn't exist.

This is because there are multiple issues which are broken about //busybox:busybox.tar

* It doesn't check errors from commands that it runs and so if they don't run it silently passes.
* It attempts to run the busybox on the host machine even though busybox may not be built for the host configuration.
* It uses a value from srcs as a tool, which is not supposed to be done. https://docs.bazel.build/versions/master/be/general.html#genrule

This means that when building from a mac it 'successfully' builds but is lacking all the right symlinks.

endobson:
This allows building the debug container without running busybox, which
is needed to build from mac OS or Windows.

Closes #117


dlorenc:
Thanks!

PR #130, opened 2017-11-14
Add a python3 distroless image 

hwright:
This should help with GoogleCloudPlatform/distroless#111 and
bazelbuild/rules_docker#229.

A few of caveats:
 * This still needs to be uploaded to gcr.io, presumably by some automated
   process that I'm not privy to.
 * This only tests importing a single module from the standard library
   (as with the 2.7 image).  I suspect there may be other dependencies
   which are required for other modules in the standard library; it
   would be nice to have tests for importing other modules which require
   additional dependencies.
 * Since the distroless base image is currently based on Debian Jessie, this
   uses version 3.4.2.  Ideally, we'd update distroless to be based on
   Stretch, and the newer versions of the dependent packages (across the
   board).  I suspect this would require some additional qualification
   on Google's end.

### DIFF
--- a/examples/python3/BUILD
+++ b/examples/python3/BUILD
@@ -1,0 +1,16 @@
+load("@io_bazel_rules_docker//python:image.bzl", "py_image")
+load("@io_bazel_rules_docker//docker:docker.bzl", "docker_build")
+
+py_image(
+    name = "hello_py",
+    srcs = ["hello.py"],
+    base = "//python3:python3",
+    main = "hello.py",
+)
+
+# This example runs a python program that walks the filesystem under "/etc" and prints every filename.
+docker_build(
+    name = "hello",
+    base = ":hello_py",
+    cmd = ["/etc"],
+)
--- a/BUILD
+++ b/BUILD
@@ -18,6 +18,8 @@ docker_bundle(
         "gcr.io/{PROJECT_ID}/java:debug": "//java:debug",
         "gcr.io/{PROJECT_ID}/java/jetty:latest": "//java/jetty",
         "gcr.io/{PROJECT_ID}/java/jetty:debug": "//java/jetty:debug",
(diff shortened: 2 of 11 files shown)


r2d4:
re: the first point, you should be able to add it to the list here and it will be built and uploaded along with the other images

https://github.com/GoogleCloudPlatform/distroless/blob/master/BUILD#L10-L36

examples/python3/BUILD
```diff
@@ -0,0 +1,16 @@
+load("@io_bazel_rules_docker//python:image.bzl", "py_image")
+load("@io_bazel_rules_docker//docker:docker.bzl", "docker_build")
+
+py_image(
```
mattmoor: Is it your expectation that we should change this to `py3_image` in the limit?

mattmoor:
@hwright re: upload process, this is done by a cloud build specified in the `yaml` file you updated once the commit lands in `master`.

examples/python3/BUILD
hwright: In the absence of something more clever, yes.

(It's unfortunate that there is a circular dependency between these repositories.)

hwright:
SGTM.

Also, sorry about the goofy history after rebasing.  I could clean up the commits here on another branch, but I don't think it matters that much (github will squash them for merge, anyway).

hwright:
Any next steps here?

mattmoor:
Ping @dlorenc @r2d4 

@duggelz FYI

duggelz:
To answer one of your questions, I've been sporadically working on creating Debian 8 packages for the Python 3.5 and 3.6 interpreters we use in App Engine Flex.  Note that updating to Debian 9 still only gets us Python 3.5, not Python 3.6.

hwright:
@duggelz Correct, though #135 is at least an attempt to get us that far.

Issue #72, opened 2017-07-06
dpkg/dpkg_src re-design / update and security


gebi:
Hi,

after a short look into dpkg/dpkg_src i'm still not sure how they are thought to be used (securely).

They do not validate signatures, they do not check the hash of packages and download from http, which all in all excludes them IMHO from use on production infrastructure.

Additionally i find it strange, that a core part of distroless now creates non-repeatable builds as neither a fixed version number is used nor a hash checked.

Thus i propose to split out what dpkg/dpkg_src currently does in skylark rules into an external script which just modifies/updates the WORKSPACE file to update packages and uses normal http_file declarations.
This would fix all the problems introduced with the current dpkg/dpkg_src rules and further clearly show what changed with debian dependencies in the repository diff.

This external tool then could also properly validate the signatures and provide secure download of debian packages with a sha256 sum it should automatically include (i can provide help with this if you want)

An additional nice feature would be to add an additional link to the same debian package in the http_file rule to point to http://snapshot.debian.org/, because mirrors typically don't hold on old packages for long. 

btw... if it's really the case that currently no security check/validation is done on the included debian packages what so ever, then the last versions of distroless images currently on gcr.io should be considered compromised and be removed ASAP.

r2d4:
I think this could be simply fixed by just changing the source to point to a snapshot and checking the hash.  Additionally, each .deb can be checked against the hash included in the Packages.gz file.

mattmoor:
@r2d4 we absolutely need to check the package against what's in the Packages.gz.

r2d4:
@gebi I added checksum validation to the `dpkg_src` rule as well as checksum validation while downloading the .debs from the Packages.gz provided checksum.  I also switched it over to use the snapshot archives.  I think that should resolve this issue?

https://github.com/GoogleCloudPlatform/distroless/pull/71

gebi:
It fixes the special case for debian because there is a snapshot.debian.org, all other package sources still have the problem.

It also removes the possibility to fix problems in debian as far as i can see because it just uses packages from the snapshot on snapshot.debian.org.

All in all it seems we will keep our fork for the forseeable future and explore the possibility to upgrade the WORKSPACE file with helper tools instead of using the new the dpkg/dpkg_src rules.

mattmoor:
[...]

PR #137, opened 2017-11-21
Use the latest Python 3.6 in the python3 image

hwright:
Debian Stretch (which is the current stable and used as the base image
for `distroless`) doesn't have Python 3.6, nor will it.  Instead, we
grab the python3.6 packages from Debian Buster, which is the current
testing branch.

This falls heavily into the Works For Me category.  Since these images
are minimalist, using a package from `testing` shouldn't conflict with
other bits in the image, but my `dpkg`-fu is sufficiently weak that I
can't guarantee that.  It'd be nice to have somebody take a deeper look.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -39,6 +39,15 @@ dpkg_src(
     url = "http://snapshot.debian.org/archive",
 )
 
+dpkg_src(
+    name = "debian_buster",
+    arch = "amd64",
+    distro = "buster",
+    sha256 = "5113371655f80a25a67982360f1b5e33383281bf884aa12ab556b7a3613f031b",
+    snapshot = "20171101T160520Z",
+    url = "http://snapshot.debian.org/archive",
+)
+
 dpkg_list(
     name = "package_bundle",
     packages = [
@@ -65,9 +74,9 @@ dpkg_list(
         "libpython2.7-stdlib",
 
         #python3
-        "libpython3.5-minimal",
-        "python3.5-minimal",
-        "libpython3.5-stdlib",
+        "libpython3.6-minimal",
+        "python3.6-minimal",
+        "libpython3.6-stdlib",
 
         #dotnet
         "libcurl3",
@@ -81,6 +90,7 @@ dpkg_list(
     sources = [
         "@debian_stretch//file:Packages.json",
         "@debian_stretch_backports//file:Packages.json",
+        "@debian_buster//file:Packages.json",
     ],
 )
 
--- a/python3/BUILD
+++ b/python3/BUILD
@@ -11,16 +11,16 @@ load("@package_bundle//file:packages.bzl", "packages")
     debs = [
         packages["zlib1g"],
         packages["libexpat1"],
-        packages["python3.5-minimal"],
-        packages["libpython3.5-minimal"],
-        packages["libpython3.5-stdlib"],
+        packages["python3.6-minimal"],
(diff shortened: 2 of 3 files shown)


MarkusTeufelberger:
This sounds like some extreme risk of having buster packages silently overwriting jessie ones (already the jessie-backports stuff is shady).

hwright:
@MarkusTeufelberger That very well may be true.

Alternatives?

MarkusTeufelberger:
My rejected alternative set of build rules, or a separate dpkg_list() rule that only contains buster packages (as well as a separate dpkg_list() rule for jessie-backports). Still won't work since the "packages" name is hardcoded, but it would be easier to fix that one.

Also, how exactly did you verify that the hash `5113371655f80a25a67982360f1b5e33383281bf884aa12ab556b7a3613f031b` is signed by some Debian authority?

hwright:
@MarkusTeufelberger I'm not quite following your suggested alternative.  Do you have a PR or doc to reference?

As for validating the hash, that can be done here: http://snapshot.debian.org/archive/debian/20171101T160520Z/dists/buster/main/binary-amd64/

MarkusTeufelberger:
In that directory you only have the hash of the Packages.gz file, it could have been put there by anyone - this is even just a HTTP only connection, so you don't even have TLS to help you. You would need to download the Release file for Debian Buster and verify the signature of that file with the Debian FTP masters' key.

Yes, PRs #94 and #96 as well as issue #72

hwright:
@MarkusTeufelberger Ah, yes.  Thanks for the clarification.  Presumably this is something that you do on a regular basis, so I would expect there to be tooling for verifying the signature?  Could you point me at the relevant tools?

MarkusTeufelberger:
Here is the documentation on how to do it manually (using `gpg2`, `wget` and a web browser): https://github.com/mgit-at/rules_pkg/tree/master/deb_packages

In general this is commonly done by `apt`/`apt-get` and various other Debian tools, but this project here does not use any of them. Returning a SHA256 hash of a Packages.gz file is not a very standard operation, so it is a bit hard to give a single command.

If you check in hashes to generate base images from, I would expect you to have followed the Debian guidelines for https://wiki.debian.org/SecureApt.

Anyways, the main issue still remains that this PR very likely is either not deterministic or just overwrites all contained packages with Buster ones. Since this project does more magic than maybe necessary, it is hard to detect this though, hence my warning instead of letting this silently seep into all base images for the official `rules_docker` rule set... ;-)

hwright:
@MarkusTeufelberger Yep.  I appreciate the "deeper look" that I had hoped for. :) I agree that the approach taken here of just pulling in arbitrary `buster` packages over a `stretch` installation probably isn't a good one.  I can close the PR as a result.

A better answer would probably be to build or reference updated packages for `stretch`, but that would likely involve hosting our own packages, or trusting somebody else to do that same, neither of which seems very palatable.

A more radical alternative would be to use ubuntu releases and packages instead of debian ones, since they are updated much more frequently, but in doing so we'd loose determinism, since Ubuntu doesn't maintain the same sort of timestamped FTP mirror that debian does.

It's probably still important that we have the ability to provide more granular package control in these images at some point.  Users aren't going to want to use a long out-of-date Python (or Java or Go), and in the case of underlying bugs it may be important to address them sooner than debian stable.  I don't have a good path forward here, but hope it's something the core team is thinking about.

Issue #120, opened 2017-10-25
Missing dependencies in python2.7 image


drigz:
The python2.7 image installs the following packages:

```
        packages["zlib1g"],
        packages["python2.7-minimal"],
        packages["libpython2.7-minimal"],
        packages["libpython2.7-stdlib"],
```

However, [libpython2.7-stdlib depends on more](https://packages.debian.org/jessie/libpython2.7-stdlib): (note - this is from the debian:jessie-backports image, which already has some packages that distroless doesn't - eg libreadline6).

```
The following extra packages will be installed:
  libexpat1 libffi6 libpython2.7-minimal libsqlite3-0 libssl1.0.0 mime-support
Recommended packages:
  file
```

Are these dependencies intentionally excluded? Should/could the deb extractor check dependencies or is this out-of-scope?

I ran into this while trying to use defusedxml inside a container: tiran/defusedxml#10

drigz:
FYI, the results with `--no-install-recommends`:

```
Recommended packages:
  bzip2 file xz-utils
The following NEW packages will be installed:
  libexpat1 libffi6 libpython2.7-minimal libpython2.7-stdlib libreadline7 libsqlite3-0 libssl1.1
  mime-support readline-common
```

PR #138, opened 2017-11-21
Add import tests for the py3 standard library

hwright:
This also adds additional shared libraries to the python 3 image which
are required by some of the standard library.

The one standard library bit we don't test is the `tkinter` integration,
which brings in a bunch of dependencies, but doesn't seem very
relevant for deploying server-side containers.


hwright:
Ping.

hwright:
This also addresses #120 for the python 3 case.

python3/testdata/python3.yaml
```diff
@@ -12,3 +12,583 @@ commandTests:
   - name: symlink 
     command: ["/usr/bin/python", "--version"]
     expectedOutput: ["Python 3.5.3"]
+
+  - name: import_abc
```
r2d4: @nkubala is there a simpler way to do this?

hwright:
What are the next steps here?

r2d4:
Sorry, LGTM.  My only question was whether we could take this and have a macro directly in the structure tests since it seems like a pattern that would be useful for other projects.

hwright:
@r2d4 Yeah, the pattern is probably abstractable at some level, I just don't want that to be part of this yak shave.

Thanks for the review.

PR #165, opened 2018-02-28
Update rules_docker to 0.4.0

kornholi:
also added a group file with the root group, some packages expect it to exist


r2d4:
Thanks!

PR #177, opened 2018-04-05
WORKSPACE: debian_stretch_security: reference stretch not jessie

evanj:
This URL should have been updated as part of commit 3f24bf1216.


evanj:
Closing because #178 includes a fix that prevents these security updates from being used at all!

Issue #174, opened 2018-03-21
[python3]  /etc/hosts not used in DNS resolution ?


jiel:
Hello,

I use a python3 distroless image on k8s with an HostAliases config that create specific entries in the /etc/hosts file of the container. It appear that the DNS resolution in the container doesn't use the /etc/hosts. Is it a bug or a feature ? Am I missing something ?

POC: 
```python
import socket

print("/etc/hosts content:\n")
with open('/etc/hosts', 'r') as fin:
    print(fin.read())
print("\nDNS Resolution: www.google.com={}".format(socket.gethostbyname('www.google.com')))

```

output:
```
/etc/hosts content:

# Kubernetes-managed hosts file.
127.0.0.1	localhost
::1	localhost ip6-localhost ip6-loopback
fe00::0	ip6-localnet
fe00::0	ip6-mcastprefix
fe00::1	ip6-allnodes
fe00::2	ip6-allrouters
10.244.0.162	feedcool-1521637560-t58q8

# Entries added by HostAliases.
127.0.0.1	www.google.com


DNS Resolution: www.google.com=216.58.212.196
```

salrashid123:
i think ```/etc/nsswitch.conf``` is missing (which i think comes from ```build-files``` package)

try this with these three files in the same folder

1)  main.py

```python
import socket

print("/etc/hosts content:\n")
with open('/etc/hosts', 'r') as fin:
    print(fin.read())
print("\nDNS Resolution: www.google.com={}".format(socket.gethostbyname('www.google.com')))

print("----------------------------------")

with open("/etc/hosts", "a") as myfile:
    myfile.write("8.8.8.8 www.google.com")

print("/etc/hosts content:\n")
with open('/etc/hosts', 'r') as fin:
    print(fin.read())
print("\nDNS Resolution: www.google.com={}".format(socket.gethostbyname('www.google.com')))
```

2) Dockerfile

```dockerfile
FROM python:3-slim AS build-env
ADD . /app
WORKDIR /app

FROM gcr.io/distroless/python3
COPY --from=build-env /app /app
COPY --from=build-env /app/nsswitch.conf /etc/nsswitch.conf
WORKDIR /app

CMD ["main.py"]
```

3) nsswitch.conf

```
passwd:         files cache
group:          files cache
shadow:         files cache

hosts:          files dns
networks:       files

protocols:      db files
services:       db files
ethers:         db files
rpc:            db files

netgroup:      files
```

-- build and rn

[...]

PR #180, opened 2018-04-09
Add nsswitch.conf

salrashid123:
Fixex [#174](https://github.com/GoogleCloudPlatform/distroless/issues/174) 
by adding the ```/etc/nsswitch.conf``` file.

---

Verification steps for this change:

1. create Dockerfile
```dockerfile
FROM python:3-slim AS build-env
ADD . /app
WORKDIR /app

#FROM gcr.io/distroless/python3
FROM bazel/python3:python3
COPY --from=build-env /app /app
WORKDIR /app

CMD ["main.py"]
```

2. create main.p

```python
import socket

print("/etc/hosts content:\n")
with open('/etc/hosts', 'r') as fin:
    print(fin.read())
print("\nDNS Resolution: www.google.com={}".format(socket.gethostbyname('www.google.com')))

print("----------------------------------")

with open("/etc/hosts", "a") as myfile:
    myfile.write("8.8.8.8 www.google.com")

print("/etc/hosts content:\n")
with open('/etc/hosts', 'r') as fin:
    print(fin.read())
print("\nDNS Resolution: www.google.com={}".format(socket.gethostbyname('www.google.com')))
```

then sync w/ the changes in this PR

```
bazel build all
bazel build python3:python3
bazel run python3:python3
```

then build and run the sample app using the gcr.io verison and then again with the local image


kevingessner:
@dlorenc It looks like the build is broken since this is merged, so there are no published images with this change in GCR.  When will fixed images be available?

PR #184, opened 2018-04-17
Fix the workspace status issue by moving the file.

dlorenc:
Workaround for https://github.com/bazelbuild/bazel/issues/5002


kevingessner:
Thanks @dlorenc !

PR #204, opened 2018-06-13
package_manager: two optimizations in `dpkg_list`

kevingessner:
In clean builds of our project (three package sources, ~100 imported dpkgs), `dpkg_list` is by far the slowest part at 4-5 minutes.  I found these two changes vastly speed up the dpkg parser; the time is now dominated by downloading the packages, instead of being CPU-bound in python.  The `dpkg_list` for our build now takes 0:45-1:30, a substantial improvement.

Thanks for taking a look.

### DIFF
--- a/package_manager/dpkg_parser.py
+++ b/package_manager/dpkg_parser.py
@@ -82,7 +82,7 @@ def download_dpkg(package_files, packages, workspace_name):
     pkg_vals_to_package_file_and_sha256 = {}
     package_to_rule_map = {}
     package_file_to_metadata = {}
-    for pkg_vals in packages.split(","):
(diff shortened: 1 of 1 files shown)


kevingessner:
@dlorenc It looks the published PAR for dplg_parser needs to be updated for this change to actually matter. How does that work? https://github.com/GoogleContainerTools/distroless/blob/4a20649aac319b2ee2f9c058c6a100784904ce7d/package_manager/package_manager.bzl#L3

kevingessner:
@dlorenc I just filed a new PR to update the PAR source:https://github.com/GoogleContainerTools/distroless/pull/206

PR #227, opened 2018-08-03
Fix python 3 compatibility

### LINKED ISSUE #199: package_manager doesn't support python3
It depends on urllib2 which is not available in python3.

JoostvDoorn:
This should fix python 3 compatibility of the package manager.

Tested with:
```
bazel test //package_manager/... --test_output=errors
```

Fixes #199


dlorenc:
Thanks!

JoostvDoorn:
@dlorenc could you update the par file?

Issue #150, opened 2017-12-09
python2.7: ctypes.utils.find_library fails due to missing dependencies


evanj:
The Python Standard Library's `ctypes.utils.find_library("c")` fails because it attempts to execute two programs. This causes gunicorn to not work (see the stack trace below). The "real" fix is to install `ldconfig` and generate the `/etc/ld.so.cache` file somehow. However, it also "works" for this specific case if you add `dash` to the container image. This is ugly, but gunicorn is pretty widely used, so it would be nice if it worked.

For details, a reproduction, and some examples of how I fixed this, see https://github.com/evanj/pydistroless_ctypes

The trace back I got when running gunicorn in a distroless python2.7 image is:

```
Traceback (most recent call last):
  File "/tmp/install/gunicorn-19.7.1-py2.py3-none-any.whl.9d2a2d914e9838232744d125e8d8a41509c11c9d/gunicorn-19.7.1-py2.py3-none-any.whl/gunicorn/util.py", line 134, in load_class
    mod = import_module('.'.join(components))
  File "/usr/lib/python2.7/importlib/__init__.py", line 37, in import_module
    __import__(name)
  File "/tmp/install/gunicorn-19.7.1-py2.py3-none-any.whl.9d2a2d914e9838232744d125e8d8a41509c11c9d/gunicorn-19.7.1-py2.py3-none-any.whl/gunicorn/workers/sync.py", line 16, in <module>
    import gunicorn.http.wsgi as wsgi
  File "/tmp/install/gunicorn-19.7.1-py2.py3-none-any.whl.9d2a2d914e9838232744d125e8d8a41509c11c9d/gunicorn-19.7.1-py2.py3-none-any.whl/gunicorn/http/wsgi.py", line 24, in <module>
    from ._sendfile import sendfile
  File "/tmp/install/gunicorn-19.7.1-py2.py3-none-any.whl.9d2a2d914e9838232744d125e8d8a41509c11c9d/gunicorn-19.7.1-py2.py3-none-any.whl/gunicorn/http/_sendfile.py", line 27, in <module>
    _libc = ctypes.CDLL(ctypes.util.find_library("c"), use_errno=True)
  File "/usr/lib/python2.7/ctypes/util.py", line 285, in find_library
    return _findSoname_ldconfig(name) or _get_soname(_findLib_gcc(name))
  File "/usr/lib/python2.7/ctypes/util.py", line 103, in _findLib_gcc
    stdout=subprocess.PIPE)
[...]

PR #228, opened 2018-08-04
python: add libc-bin and ld.so.cache so ctypes works

evanj:
Third-party packages like gunicorn and monotonic assume that the standard
library function ctypes.util.find_library can find 'c' and 'rt', which are
parts of glibc. To do this, they execute '/sbin/ldconfig -p'. To make this
work, install the binary in the libc-bin deb and generate /etc/ld.so.cache.

Ideally, we would generate /etc/ld.so.cache from the container at build
time. However, that would require users to have Docker. Instead, check in a
pre-generated binary, and add a test that checks that it equals the
generated version. This allows end users to build python images without
Docker, but developers working on Distroless will have a check that it is
up to date.

Fixes https://github.com/GoogleContainerTools/distroless/issues/150


dlorenc:
I think you have to rebase this from your last commit. Otherwise LGTM.

evanj:
Awesome! I just rebased it; let's hope that Travis still passes :)

PR #239, opened 2018-09-19
WORKSPACE: Update Debian snapshot for security fixes.

evanj:
Includes a fix to openjdk, curl and maybe others I didn't notice. This
might also pick up newer backports.
Use https to access snapshot.debian.org.

DSA-4286 curl: https://www.debian.org/security/2018/dsa-4286
DSA-4268 openjdk-8: https://www.debian.org/security/2018/dsa-4268


lathama:
You missed the second s/http/https/

PR #244, opened 2018-10-04
Add Docker limits and urandom

SvenBayer:
Should the Docker image provide automatic container limits awareness to the Java app?
I think it does not make sense to leave out the JVM options for memory limits since it is an unexpected behaviour.

MaxRAMFraction=1 sets the maximum Heap to around 95% of the Docker container memory. Otherwise, only around half or less of the Docker container memory is set as maximum Heap. Is this safe since there is not much anything else running in a Distroless Docker container?

Is URandom really safe? https://www.2uo.de/myths-about-urandom/
This might be arguable?

### DIFF
--- a/java/BUILD
+++ b/java/BUILD
@@ -18,6 +18,10 @@ cacerts_java(
     ],
     entrypoint = [
         "/usr/bin/java",
+        "-XX:+UnlockExperimentalVMOptions",
+        "-XX:+UseCGroupMemoryLimitForHeap",
+        "-XX:MaxRAMFraction=1",
+        "-Djava.security.egd=file:/dev/./urandom",
         "-jar",
         # We expect users to use:
         # cmd = ["/path/to/deploy.jar", "--option1", ...]


Fleshgrinder:
The latest Java 8 release has additional options that allow you to tweak the JVM in a simpler fashion. Hence, using the proposed values as forced defaults seems like a bad idea to me.

References:

- https://stackoverflow.com/questions/49854237
- https://medium.com/adorsys/jvm-memory-settings-in-a-container-environment-64b0840e1d9e

briandealwis:
Sorry @SvenBayer I didn't see this prior to #295.

loosebazooka:
@briandealwis can we close this now?

briandealwis:
Using `/dev/urandom` involves trade-offs that I don't think a base image like distroless/java is in a position to decide.  Pivotal have a [nice blog post summarizing the issue](https://content.pivotal.io/blog/challenges-with-randomness-in-multi-tenant-linux-container-platforms).

Issue #241, opened 2018-09-30
dotnet core 2.1 support


liguobao:
dotnet core 2.1 had released on 2018-05-30,we maybe can support it to help our work.

[Announcing .NET Core 2.1](https://blogs.msdn.microsoft.com/dotnet/2018/05/30/announcing-net-core-2-1/)

Thanks .

Issue #214, opened 2018-07-10
Version/Tag distroless images


saturnism:
For Java image, it's `gcr.io/distroless/java`, but doesn't have any reference to which Java version. This is important to be tagged.

lathama:
@saturnism can you share how you would suggest the tagging be done?

saturnism:
similar to https://hub.docker.com/_/openjdk/

we should at least call out 8u???, and always update 8 to point to the latest.

saturnism:
quick ping.

saturnism:
Hmm.. I'm not sure I'm following how this relates to tagging the container images.

The issue is that untagged images aren't stable, and relaying on `latest` tag is not too great.

dlorenc:
Can you explain what you mean by stable? All images are tagged with the git commit they were built at, and are also available by their sha256 digest.

asaikali:
As a user I would like to know exactly what version of java my code is running with. The latest tag does not tell me if it's JDK 8 update 171 or JDK 8 update 131 or some other value. It makes my build unstable since every time I run it I might end up with a different version of the java distroless image. Once you add support for Java 11 I assume latest will point at java 11 not 8 so my builds will break because I am using the wrong version of java with my app.

saturnism:
Where is the list of git commit? and is there a documentation on which git commit represents which JDK build?

a tag is helpful to expose this in a readable and easily understood way.

darkobits:
Adding my support for this.

In addition to providing a quick, easy way to determine which version of X we're getting, tags provide a way to **pin to a specific version**. Otherwise, you're always implicitly pulling the `latest` tag, which is a moving target. Without the ability to pin to a specific tag, consumers have no way of knowing what they're getting.

---

Brief addendum, root cause possibly being my docker-fu is lacking:

Because these images are not hosted on the public registry and because the Docker CLI doesn't seem to implement a command that enumerates the tags for an image, I've had a difficult time trying to determine what tags actually available.


**Specific to the Node image:**

It would be extremely helpful to indicate the Node version used [here](https://github.com/GoogleContainerTools/distroless/tree/master/experimental/nodejs), especially considering that none of the source files under that directory make any reference to the Node version being used. I did finally find it [here](https://github.com/GoogleContainerTools/distroless/blob/6a87d98e2d00b98534cb711382046d2f8164461d/WORKSPACE#L165). 😕

briandealwis:
[...]

PR #245, opened 2018-10-06
upgrade dotnet to 2.1

salrashid123:
Upgrade distroless `dotnet` to 2.1  https://github.com/GoogleContainerTools/distroless/issues/241

pls do not merge here yet; just review would be fine at the moment;

we don't have versions/tags here so the call on upgrading is upto maintainers on this and per https://github.com/GoogleContainerTools/distroless/issues/214

---

Also address incorrect, unnecessary dependencies added in previously:

dotnet 2+ only requires these:
 - [https://docs.microsoft.com/en-us/dotnet/core/linux-prerequisites?tabs=netcore2x#ubuntu](https://docs.microsoft.com/en-us/dotnet/core/linux-prerequisites?tabs=netcore2x#ubuntu)


salrashid123:
added in support for explicit version similar to `

```
gcr.io/distroless/python2.7
gcr.io/distroless/python3
```
eg.
```
gcr.io/distroless/dotnet     << latestcoreclr (after this PR, its 2.7)
gcr.io/distroless/dotnet2.1  << named version
```

PR #252, opened 2018-10-25
Include ca-certificates status file for traceability

dprotaso:


dprotaso:
I might rethink the approach to not use `dpkg-deb` executable - which isn't installed by default on Mac OS X

dprotaso:
@dlorenc take a look when you get a chance

PR #264, opened 2018-12-10
WORKSPACE: Update to latest Debian snapshot

evanj:
This includes a security fix to OpenSSL released on 2018-11-30:
https://www.debian.org/security/2018/dsa-4348

This is reported as a vulnerability by the Google Cloud container
security scanner.


evanj:
Closing because I made the commit on a machine without the right git configuration; #265 is the same change.

PR #283, opened 2019-01-23
Pull the newest snapshot

chanseokoh:
`openjdk-11` was back-ported to Debian "stretch" quite recently, and the Java 11 landscape is changing fast as we approach the end of January. Therefore, although the previous stretch-packports snapshot included `openjdk-11`, I think it's good to advance the snapshot.

~~This PR also reads in `openjdk-11-jre-headless` for later use.~~

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -35,8 +35,8 @@ dpkg_src(
     name = "debian_stretch",
     arch = "amd64",
     distro = "stretch",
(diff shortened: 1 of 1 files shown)


chanseokoh:
Realized there's a much more recent snapshot.

@briandealwis I've also reverted reading in the `openjdk-11-jre-headless` package.

dlorenc:
> @dlorenc by any chance, does this repo has an automatic hook that will republish all the images whenever the repo changes the snapshots? I'm just wondering if I have to hold off merging this if that's the case.

It does. Why would you have to hold off merging though?

Issue #236, opened 2018-09-02
Java 11 and future Java version support


anuraaga:
I noticed a closed bug about Java 9 here and one about Java 10 in [jib](https://github.com/GoogleContainerTools/jib/issues/550) but didn't find an open issue here so figured it'd be good to track it.

From the jib issue, it sounds like Java 10 will be skipped in favor of Java 11 LTS. This is good news, but I was wondering if there's also a version support plan going forward? LTS-only, LTS + latest non-LTS, or something similar? If distroless always supports the latest Java, at least for 6 months, it means people can use it while staying at the cutting edge of Java :)

barmic:
What is the plan? Do you think use https://adoptopenjdk.net/ and there API?

jabrena:
One idea to consider: 

`jib` could allow using complete OpenJDK builds or Java Runtime Images to save memory.

Example:
https://github.com/ev3dev-lang-java/openjdk-ev3/blob/master/scripts/zip.sh

Results:

https://ci.adoptopenjdk.net/view/ev3dev/job/openjdk11_build_ev3_linux/

Complete JDK: 265mb
JRI: 22mb 

Juan Antonio

arashbi:
is there any jdk 11 image at the moment?

chanseokoh:
Java 11 images (`gcr.io/distroless/java:11` and `gcr.io/distroless/java:11-debug`) are now available, in parity with the Java 8 images (`gcr.io/distroless/java:8` and `gcr.io/distroless/java:8-debug`). The `latest` and `debug` continue to point to the Java 8 images (i.e., `8` and `8-debug`, respectively).

Like the Java 8 images, the Java 11 images unpack the headless OpenJDK JRE Debian package (`openjdk-11-jre-headless`). However, I noticed the size of the 11 images are a bit larger than the 8 images.

```
gcr.io/distroless/java   8                   9c0bc85b4448        49 years ago        119MB
gcr.io/distroless/java   11                  a4338971ab0a        49 years ago        189MB
```

PR #285, opened 2019-01-24
Add Java 11 images (latest and debug) / add more tags

chanseokoh:
About #236.

~~This is on top of #283, so #283 should be reviewed first.~~

Debian stretch recently back-ported `openjdk-11`. The Java 11 images get the `openjdk-11-jre-headless` (in parity with the Java 8 images) from the `stretch-backports` snapshot archive.

`gcr.io/distroless/java:latest` and `gcr.io/distroless/java:debug` still point to Java 8.

Two new images:
`gcr.io/distroless/java:11`
`gcr.io/distroless/java:11-debug`

Additional tags:
`gcr.io/distroless/java:8`
`gcr.io/distroless/java:8-debug`
`gcr.io/distroless/java:11`
`gcr.io/distroless/java:11-debug`
`gcr.io/distroless/java/jetty:java8`
`gcr.io/distroless/java/jetty:java8-debug`


chanseokoh:
Oops, hit the enter key mistakenly. This has to wait until #283 is merged.

loosebazooka:
So this generally matches the format of other java images: `java:<version>-<more_descriptors>`, but should we also include the detailed version numbers? for example `java:8u181-xxx` or `java:11.0.1-xxx` as a secondary tag? I'm not really sure at that point that the different vendors will even share version numbers :\, so we might even have to include vendor?

chanseokoh:
Yeah, we do plan to add more detailed tags along the way. It's just that I don't do everything in this PR.

I think we can switch vendors for existing tags. Of course, we can add tags with vendors too (later).

briandealwis:
Can we put the version and vendor information as image labels?  Putting this information into the tags will get pretty unwieldy.   And there's no choice being offered to consumers: we're not planning to publish multiple vendors, and we're not maintaining parallel streams (e.g., 11.0.1 with security updates now that 11.0.2 is released).

briandealwis:
@chanseokoh were you going to offer jetty with java11 too?

chanseokoh:
Took out #283 (advancing the snapshot), so this PR is standalone and no longer depend on #283.

cloudbuild.yaml
```diff
@@ -31,19 +31,49 @@ steps:
   args: ['tag', 'bazel/java:java8', 'gcr.io/$PROJECT_ID/java:latest']
 
 - name: gcr.io/cloud-builders/bazel
-  args: ['run', '//java:debug']
+  args: ['run', '//java:java8']
+- name: gcr.io/cloud-builders/docker
```
loosebazooka: I'm not sure how this whole section works, but do we need to rerun the bazel build of the image for each tag of the same build?

looks like we have
```
- name: gcr.io/cloud-buidlers/bazel
  args: ['run', '//java:java8']
```
on line 28 and line 33

chanseokoh: Ah, good catch. I think we can just tag it multiple times. Will updated it.

Note, it's possible that we will remove this `cloudbuild.yaml` at any point (#286).

cloudbuild.yaml
```diff
@@ -29,21 +29,39 @@ steps:
   args: ['run', '//java:java8']
 - name: gcr.io/cloud-builders/docker
   args: ['tag', 'bazel/java:java8', 'gcr.io/$PROJECT_ID/java:latest']
+- name: gcr.io/cloud-builders/docker
```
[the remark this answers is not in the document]
ImJasonH: When you `blaze run` a `docker_build` target, it builds the image and puts it in the local Docker daemon tagged as (I believe) `bazel/[path/to]:target`. This looks right to me, and most importantly if running the build works, you've got it correct. `docker tag` will fail if the `src` tag isn't found locally.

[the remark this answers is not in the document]
loosebazooka: so the docker daemon is shared between containers?

java/BUILD
```diff
@@ -23,12 +23,14 @@ cacerts_java(
...
     ],
     symlinks = {
-        "/usr/bin/java": "/usr/lib/jvm/java-8-openjdk-amd64/jre/bin/java",
+        "/usr/bin/java": java_executable_path,
     },
     tars = [":cacerts_java"],
-) for mode in [
-    "",
-    ":debug",
+) for (rule_name, jre_deb, java_executable_path) in [
```
loosebazooka: Probably just get some bazel person to look at this, is there a way to share the java_executable_path reference or would that make this section look super wonky?
@nlopezgi wdyt?

loosebazooka: I don't feel super strongly about this. I was concerned about each java path could potentially be shared for each java8, java11 option, but it seems like it's not worth the trouble right now

PR #287, opened 2019-01-30
don't explicitly tag base:latest or cc:latest in cloudbuild.yaml

sharifelgamal:


cloudbuild.yaml
```diff
@@ -1,24 +1,20 @@
 timeout: 1800s
-steps:  
+steps:
 - name: gcr.io/cloud-builders/bazel
   args: ['run', '//base:static']
 - name: gcr.io/cloud-builders/docker
   args: ['tag', 'bazel/base:static', 'gcr.io/$PROJECT_ID/static:latest']
 
 - name: gcr.io/cloud-builders/bazel
```
chanseokoh: I wonder we can just remove `run //base:base` too. Unsure. For `run //cc:cc` too.

chanseokoh:
Also, is it worth adding some comments whey they are missing? (If I did not have the background, I would have wondered why.)

cloudbuild.yaml
```diff
@@ -106,9 +102,7 @@ steps:
 
 images:
   - 'gcr.io/$PROJECT_ID/static:latest'
```
chanseokoh: Remove `static:latest` too, right?

sharifelgamal: Yep. Fixed.

BUILD
```diff
@@ -5,10 +5,7 @@ load("@io_bazel_rules_docker//container:container.bzl", "container_bundle")
 container_bundle(
     name = "all",
     images = {
-        "gcr.io/{PROJECT_ID}/go:latest": "//base:static",
```
chanseokoh: Oh, actually this is not `static:latest` but `go:latest`. But probably this has to go anyway?

sharifelgamal: We don't actually maintain anything called go anymore, the name change to static happened awhile ago.

Issue #276, opened 2019-01-16
Add Java version labels to Java distroless images


liyinan926:
The following two labels are expected to be added:

* `jdk.vendor`: the vendor of the JDK.
* `jdk.version`: the version of the JDK, e.g., 1.8.0_131 or 9.0.1 that can be obtained by `java -version`. 

@coollog @loosebazooka

loosebazooka:
@mattmoor @dlorenc is there way to extract "Version" from the pkg metadata and expose it as a label during build?

chanseokoh:
I've briefly thought about this. The current repo setup downloads the entire metadata list of all the packages from a certain Debian snapshot archive (for example, the metadata list file is https://snapshot.debian.org/archive/debian/20190125T150847Z/dists/stretch/main/binary-amd64/Packages.gz). The setup actually downloads and unpacks such a file and parses it. (In the end, it parses it into a JSON.) It is easy to check the contents of the metadata manually:

```
$ wget https://snapshot.debian.org/archive/debian/20190125T150847Z/dists/stretch/main/binary-amd64/Packages.gz
$ gunzip -c Packages.gz | grep "Package: openjdk-8"
Package: openjdk-8-dbg
Package: openjdk-8-demo
Package: openjdk-8-doc
Package: openjdk-8-jdk
Package: openjdk-8-jdk-headless
Package: openjdk-8-jre
Package: openjdk-8-jre-headless
Package: openjdk-8-jre-zero
Package: openjdk-8-source
Package: openjdk-8-jre-dcevm
```

The actual code that does this is `package_manager/dpkg_parser.py` and `parse_metadata.py`. (However, seems like it pulls in a compiled `dpkg_parser.par` binary from a GCS bucket though.) I wonder if we can utilize this existing infrastructure. Might need some work though.

chanseokoh:
BTW, here's basically what's happening in this repo.

```
$ cp package_manager/dpkg_parser.py .
$ mkdir file
$ python2 dpkg_parser.py \
    --download-and-extract-only=true \
    --distro=stretch \
    --arch=amd64 \
    --snapshot=20190125T092538Z \
    --mirror-url=https://snapshot.debian.org/archive \
    --sha256=4b981bd2445d85cf1d93ea9f2d2dc235d20543ecdadd0d8065a10793b94eab9b
```
Then the script above will download/generate the following files:
```
Packages
file/os_release.tar
file/Packages.json
etc/os-release
```

[...]

PR #290, opened 2019-02-06
Maintain package version mapping

briandealwis:
#276 and #214 request branding the distroless/java images with metadata about the underlying language runtime version. Unfortunately the package information is not available when building the image: all that is available is the actual `.deb` file.

https://github.com/GoogleContainerTools/distroless/blob/55b0eb29ae8228e59ec14b02ca6b904c20340f06/java/BUILD#L12-L17

Rather than try to extract the `.deb`'s `control.tar.gz` and its `control` file, this PR changes the `package_manager` to also record a package-to-version mapping (`versions`) that is written out with the `@package_bundle//file:packages.bzl`.

The resulting `packages.bzl` file now looks like:
```
packages = {"libgcc1": "@package_bundle//file:bGliZ2NjMQ==.deb", "libgomp1": "@package_bundle//file:bGliZ29tcDE=.deb", ... }
versions = {"libgcc1": "1:6.3.0-18+deb9u1", "libgomp1": "6.3.0-18+deb9u1", ...}
```

With this change, the `java/BUILD` can now pull in the versions information and use `versions[jre_deb]` to annotate the generated images. See [my proposed branch for an example](https://github.com/GoogleContainerTools/distroless/compare/i276-jre).

**Still outstanding:** I'm not sure how the `package_manager_tools` bundle is updated with `package_manager/package_manager.bzl`.

### DIFF
--- a/package_manager/dpkg_parser.py
+++ b/package_manager/dpkg_parser.py
@@ -82,6 +82,7 @@ def download_dpkg(package_files, packages, workspace_name):
     """
     pkg_vals_to_package_file_and_sha256 = {}
     package_to_rule_map = {}
+    package_to_version_map = {}
     package_file_to_metadata = {}
     for pkg_vals in set(packages.split(",")):
         pkg_split = pkg_vals.split("=")
@@ -102,6 +103,7 @@ def download_dpkg(package_files, packages, workspace_name):
                 pkg = metadata[pkg_name]
                 buf = urllib.request.urlopen(pkg[FILENAME_KEY])
                 package_to_rule_map[pkg_name] = util.package_to_rule(workspace_name, pkg_name)
+                package_to_version_map[pkg_name] = metadata[pkg_name][VERSION_KEY]
                 out_file = os.path.join("file", util.encode_package_name(pkg_name))
                 with io.open(out_file, 'wb') as f:
                     f.write(buf.read())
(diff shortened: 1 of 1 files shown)


chanseokoh:
> **Still outstanding:** I'm not sure how the `package_manager_tools` bundle is updated with `package_manager/package_manager.bzl`.

At least I know `dpkg_parser.par` is auto-uploaded whenever a PR is merged. For example, the latest commit from `git log` is
```
commit 55b0eb29ae8228e59ec14b02ca6b904c20340f06 (HEAD -> master, origin/master, origin/HEAD)
Author: Sharif Elgamal <selgamal@google.com>
Date:   Wed Jan 30 17:32:41 2019 -0800
```
and we already have the binary built from it: https://storage.googleapis.com/distroless/package_manager_tools/55b0eb29ae8228e59ec14b02ca6b904c20340f06/dpkg_parser.par

briandealwis:
@sharifelgamal is it safe to merge this patch to `distroless/package_manager`?  Is there anything else that needs to be done for it to be picked up?

chanseokoh:
I think this is good to go and safe. Even if the new `dpkg_parser.par` breaks things if used, we won't be picking up the new one, because we are fetching a specific `dpkg_parser.par` built in the past.

PR #291, opened 2019-02-06
Record JRE version details in Java image metadata

briandealwis:
#276 requests that the `java` images provide information as to the JRE version. This PR follows on from #290 and adds the following metadata to the `distroless/java` images:
  - it sets a `JAVA_VERSION` environment variable to the JRE version, currently `8u181` for the Java 8 image and `11.0.1` for the Java 11 image
  - ~~it sets three labels:~~
    - ~~`jdk.vendor` is hard-coded to `openjdk` since we explicitly select the `openjdk-*-jre-headless` packages~~
    - ~~`jdk.version` is the same release number reported in `JAVA_VERSION`~~
    - ~~`jdk.source` records soe provenance information listing the Debian package name and version: currently `openjdk-8-jre-headless 8u181-b13-2~deb9u1` and `openjdk-11-jre-headless 11.0.1+13-2~bpo9+1`~~

_Update: I'm not convinced of these label keys, and since environment variables are as easy to access as labels, the `JAVA_VERSION` will do for now._

It adds some metadata tests too ~~though the `jdk.source` tests are commented out as they require container-structure-test v1.7.0 (bazelbuild/rules_docker#685)~~.

~~I'm not particularly enamoured by the label names.~~  My priority is the `JAVA_VERSION` environment variable.

### DIFF
--- a/java/jre_ver.bzl
+++ b/java/jre_ver.bzl
@@ -1,9 +1,9 @@
 def jre_ver(version):
-  """Extract JRE version from an OpenJDK debian package.
-    debian packages have versions in the form:
+  """Extract JRE version from a Debian openjdk package.
+    Debian packages versions are of the form:
        openjdk-8-jre*: 8u171-b11-1~deb9u1
        openjdk-11-jre*: 11.0.1+13-2~bpo9+1
   """
-  if version.startswith("8u"): return version.split("-",2)[0]
-  if version.startswith("11."): return version.split("+",2)[0]
-  return "unknown: " + version
+  if version.startswith("8u"): return version.split("-")[0]
+  if version.startswith("11."): return version.split("+")[0]
+  fail("unrecognized openjdk package version: " + version)
--- a/java/jre_ver.bzl
+++ b/java/jre_ver.bzl
@@ -1,0 +1,9 @@
+def jre_ver(version):
+  """Extract JRE version from an OpenJDK debian package.
+    debian packages have versions in the form:
+       openjdk-8-jre*: 8u171-b11-1~deb9u1
+       openjdk-11-jre*: 11.0.1+13-2~bpo9+1
+  """
+  if version.startswith("8u"): return version.split("-",2)[0]
+  if version.startswith("11."): return version.split("+",2)[0]
+  return "unknown: " + version
--- a/package_manager/dpkg_parser.py
+++ b/package_manager/dpkg_parser.py
@@ -82,6 +82,7 @@ def download_dpkg(package_files, packages, workspace_name):
     """
     pkg_vals_to_package_file_and_sha256 = {}
     package_to_rule_map = {}
+    package_to_version_map = {}
     package_file_to_metadata = {}
     for pkg_vals in set(packages.split(",")):
         pkg_split = pkg_vals.split("=")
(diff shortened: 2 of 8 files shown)


chanseokoh:
There are formatting errors. You need to use the same `buildifier` version being used in Travis: `wget https://github.com/bazelbuild/buildtools/releases/download/0.11.1/buildifier`

Once #290 is merged, I expect the updated `dpkg_parser.par` will become available soon. You can then update the URL in `package_manager.bzl` to pull in the new binary, I think.

briandealwis:
Ok will do — thanks for the pointer.

loosebazooka:
Is the problem with the label keys that we're not choosing standard ones?

java/jre_ver.bzl
```diff
@@ -0,0 +1,9 @@
+def jre_ver(version):
+  """Extract JRE version from an OpenJDK debian package.
+    debian packages have versions in the form:
+       openjdk-8-jre*: 8u171-b11-1~deb9u1
+       openjdk-11-jre*: 11.0.1+13-2~bpo9+1
+  """
+  if version.startswith("8u"): return version.split("-",2)[0]
+  if version.startswith("11."): return version.split("+",2)[0]
```
chanseokoh: nit: I believe `version.split("+")[0]` (no optional second parameter) will work and look simpler. I don't think we need this kind of optimization.

java/jre_ver.bzl
```diff
@@ -0,0 +1,9 @@
...
+  return "unknown: " + version
```
chanseokoh: Hmm... would it be difficult to not set `JAVA_VERSION` if unknown?

java/jre_ver.bzl
loosebazooka: Agree. Perhaps the consumer can handle this on their own.

java/jre_ver.bzl
briandealwis: I'd actually like to trigger a build error.  Thinking to the day that we pull in Java 15 and end up with a broken version string.

briandealwis: Ok, that should use the `fail()` function.

chanseokoh: Nice! I like that idea.

chanseokoh:
I wonder why the status of the "push" Travis isn't updated. The build actually succeeded. I retriggered the build a few hours ago, but it didn't help.

java/jre_ver.bzl
briandealwis: `fail()` does the trick:
```
ERROR: /Users/bsd/Manumitting/Projects/GooglePluginEclipse/repo-distroless/java/BUILD:13:1: Traceback (most recent call last):
	File "/Users/bsd/Manumitting/Projects/GooglePluginEclipse/repo-distroless/java/BUILD", line 13
		container_image(name = rule_name, base = "//cc" if...", <5 more arguments>)
	File "/Users/bsd/Manumitting/Projects/GooglePluginEclipse/repo-distroless/java/BUILD", line 27, in container_image
		jre_ver(versions[jre_deb])
	File "/Users/bsd/Manumitting/Projects/GooglePluginEclipse/repo-distroless/java/jre_ver.bzl", line 10, in jre_ver
		fail(("unrecognized openjdk package v...))
unrecognized openjdk package version: 15.0.2
ERROR: error loading package 'java': Package 'java' contains errors
INFO: Elapsed time: 0.093s
INFO: 0 processes.
FAILED: Build did NOT complete successfully (1 packages loaded)
```

package_manager/dpkg_parser.py
```diff
@@ -82,6 +82,7 @@ def download_dpkg(package_files, packages, workspace_name):
     """
     pkg_vals_to_package_file_and_sha256 = {}
     package_to_rule_map = {}
+    package_to_version_map = {}
```
chanseokoh: I guess these changes will go away when you merge with master.

briandealwis:
> @loosebazooka: Is the problem with the label keys that we're not choosing standard ones?

Yeah. Both [Docker](https://docs.docker.com/config/labels-custom-metadata/#key-format-recommendations) and [OCI](https://github.com/opencontainers/image-spec/blob/master/annotations.md) recommend namespacing labels/annotations using reverse-DNS prefixes.

PR #294, opened 2019-02-11
Move zlib to cc image

loosebazooka:


WORKSPACE
```diff
@@ -78,9 +78,9 @@ dpkg_list(
         "libgcc1",
         "libgomp1",
         "libstdc++6",
+        "zlib1g",
```
chanseokoh: There's another instance of `"zlib1g"` under `#dotnet`. Remove that too.

```
        #dotnet
        "libcurl3",
        "libgssapi-krb5-2",
        "libicu57",
        "liblttng-ust0",
        "libssl1.0.2",
        "libunwind8",
        "libuuid1",
        "zlib1g",
        "curl",
```

loosebazooka: done, but left the one in package_manager/README.md, I'm going to take another look at that file

chanseokoh: Yeah, I think we can leave the one in `package_manager/README.md`.

loosebazooka:
I'm going to close this for now. I'm less enamored with this solution than I would like.

PR #295, opened 2019-02-11
Use container-friendly arguments for Java 8 images

### LINKED ISSUE #289: Make Java images container-friendly by default
set default container env var `JAVA_TOOL_OPTIONS="-XX:+UseContainerSupport"`  similar to https://hub.docker.com/r/adoptopenjdk/openjdk11/

briandealwis:
Update Java 8 images to use `-XX:+UnlockExperimentalVMOptions -XX:+UseCGroupMemoryLimitForHeap`.

Fixes #289

### DIFF
--- a/java/testdata/java8.yaml
+++ b/java/testdata/java8.yaml
@@ -19,7 +19,9 @@ fileExistenceTests:
     path: "/bin/sh"
     shouldExist: false
 metadataTest:
-  entrypoint: ["/usr/bin/java", "-XX:+UnlockExperimentalVMOptions", "-XX:+UseCGroupMemoryLimitForHeap", "-jar"]
+  entrypoint: ["/usr/bin/java", "-jar"]
   env:
     - key: 'JAVA_VERSION'
       value: '8u181'
+    - key: '_JAVA_OPTIONS'
+      value: '-XX:+UnlockExperimentalVMOptions -XX:+UseCGroupMemoryLimitForHeap'
--- a/java/testdata/java8.yaml
+++ b/java/testdata/java8.yaml
@@ -1,10 +1,12 @@
-schemaVersion: "1.0.0"
+schemaVersion: "2.0.0"
 commandTests:
   - name: java
-    command: ["/usr/lib/jvm/java-8-openjdk-amd64/jre/bin/java", "-version"]
+    command: "/usr/lib/jvm/java-8-openjdk-amd64/jre/bin/java"
+    args: ["-version"]
     expectedError: ['openjdk version "1\.8.*"']
   - name: java-symlink
-    command: ["/usr/bin/java", "-version"]
+    command: "/usr/bin/java"
+    args: ["-version"]
     expectedError: ['openjdk version "1\.8.*"']
 fileExistenceTests:
   - name: certs
@@ -16,3 +18,6 @@ fileExistenceTests:
   - name: no-shell
     path: "/bin/sh"
     shouldExist: false
+metadataTest:
+  entrypoint: ["/usr/bin/java", "-XX:+UnlockExperimentalVMOptions", "-XX:+UseCGroupMemoryLimitForHeap", "-jar"]
+
--- a/java/BUILD
+++ b/java/BUILD
@@ -17,17 +17,31 @@ cacerts_java(
         packages[jre_deb],
     ],
     entrypoint = ["/usr/bin/java"] + java_vm_args + ["-jar"],
-        # We expect users to use:
-        # cmd = ["/path/to/deploy.jar", "--option1", ...]
+    # We expect users to use:
+    # cmd = ["/path/to/deploy.jar", "--option1", ...]
     symlinks = {
         "/usr/bin/java": java_executable_path,
     },
     tars = [":cacerts_java"],
 ) for (rule_name, jre_deb, java_executable_path, java_vm_args) in [
-    ("java8", "openjdk-8-jre-headless", "/usr/lib/jvm/java-8-openjdk-amd64/jre/bin/java",
-      ["-XX:+UnlockExperimentalVMOptions", "-XX:+UseCGroupMemoryLimitForHeap"]),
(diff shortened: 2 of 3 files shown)


loosebazooka:
In the case of jib, this wouldn't help us would it?

briandealwis:
Jib can't really know what JDK is in the base image (without running it).  These options are no longer needed when/if distroless/java moves to 8u191 or later.  So I think this actually helps Jib.

loosebazooka:
Ah, but by default don't we redefine the entrypoint? I was wondering if putting in the environment might be more helpful?

java/testdata/java8.yaml
```diff
@@ -16,3 +18,6 @@ fileExistenceTests:
   - name: no-shell
     path: "/bin/sh"
     shouldExist: false
+metadataTest:
+  entrypoint: ["/usr/bin/java", "-XX:+UnlockExperimentalVMOptions", "-XX:+UseCGroupMemoryLimitForHeap", "-jar"]
```
chanseokoh: I like putting these options to the images. The only concern I have is that these options won't work or may fail for version  >= 191 (if I'm not mistaken). If what I am thinking is right, I think we should check the version is 181 in this test and add a comment that these flags should be removed (or a new set of options should be used, I don't know) for >= 191.

chanseokoh:
+1 to put these in an applicable env variable.

I can't remember. These flags are not necessary at all starting from 191? Or does 191 need a different set of flags?

briandealwis:
@loosebazooka good point

@chanseokoh these options [are the default on 191](
https://www.oracle.com/technetwork/java/javase/8u191-relnotes-5032181.html#JDK-8146115), but having them doesn't hurt.  Using `JAVA_TOOL_OPTIONS` (or `_JAVA_OPTIONS`) results in a `Picked up` message to the output:
```
$ JAVA_TOOL_OPTIONS='-XX:+UnlockExperimentalVMOptions' java -version
Picked up JAVA_TOOL_OPTIONS: -XX:+UnlockExperimentalVMOptions
java version "1.8.0_192"
Java(TM) SE Runtime Environment (build 1.8.0_192-b12)
Java HotSpot(TM) 64-Bit Server VM (build 25.192-b12, mixed mode)
$ _JAVA_OPTIONS='-XX:+UnlockExperimentalVMOptions' java -version
Picked up _JAVA_OPTIONS: -XX:+UnlockExperimentalVMOptions
java version "1.8.0_192"
Java(TM) SE Runtime Environment (build 1.8.0_192-b12)
Java HotSpot(TM) 64-Bit Server VM (build 25.192-b12, mixed mode)
```

Perhaps we should instead teach jib not to overwrite the entrypoint if it seems satisfactory.

java/testdata/java8.yaml
chanseokoh: Marking resolved: given that it doesn't hurt to have the flags for >= 191, it's enough to add the comment that the flags are unnecessary for >= 191. No need for a specific 181 version check.

chanseokoh:
> Perhaps we should instead teach jib not to overwrite the entrypoint if it seems satisfactory.

But Jib doesn't do `java -jar some.jar`. By default it has to overwrite to do `java -cp ...`.

Besides that, this PR looks good to me.

briandealwis:
Good point @chanseokoh.  I suppose jib could check if the last component of the base image's entrypoint is `-jar` and adjust accordingly.

chanseokoh:
Do you plan to add the environment variable `JAVA_TOOL_OPTIONS` for the people who override the entrypoint? In this PR or in a follow-up? Other significant JDK images do the same, so I think it makes sense to add it to Distroless.

briandealwis:
Only the `adoptopenjdk` images specify `JAVA_TOOL_OPTIONS`.  But its use leads to _Picked up JAVA_TOOL_OPTIONS_ output messages from the JVM.  This message [cannot be suppressed](https://bugs.openjdk.java.net/browse/JDK-8039152):

```sh
$ docker run -it adoptopenjdk/openjdk11 java 
Picked up JAVA_TOOL_OPTIONS: -XX:+UseContainerSupport
Usage: java [options] <mainclass> [args...]
[...]
```

The other concern I have is that `JAVA_TOOL_OPTIONS` is [documented](https://docs.oracle.com/javase/8/docs/technotes/guides/troubleshoot/envvars002.html) as being meant for agents and other diagnostic tools.

We only need these arguments because _stretch_ includes 1.8.0_181 which is oblivious to running within a container.  We're only considering *_TOOL_OPTIONS because Jib rewrites the entrypoint.  TBH I'd rather pull in 1.8.0_191 from unstable, or teach Jib to parse and pull in JVM flags.

If we were going to set anything, I'd prefer setting `_JAVA_OPTIONS` so that these options aren't inadvertently overridden by a consuming image, and result in a poorly-behaving container.

chanseokoh:
LGTM. We can thinkg about adding the environment variable later.

> Only the `adoptopenjdk` images specify `JAVA_TOOL_OPTIONS`. But its use leads to _Picked up JAVA_TOOL_OPTIONS_ output messages from the JVM. This message [cannot be suppressed](https://bugs.openjdk.java.net/browse/JDK-8039152):

I didn't get it. Is there a reason to suppress the message?

> We're only considering *_TOOL_OPTIONS because Jib rewrites the entrypoint.

I was considering any use cases of overriding the entrypoint, not just Jib. For example. someone may not want to have a runnable JAR?

> The other concern I have is that `JAVA_TOOL_OPTIONS` is [documented](https://docs.oracle.com/javase/8/docs/technotes/guides/troubleshoot/envvars002.html) as being meant for agents and other diagnostic tools.

> If we were going to set anything, I'd prefer setting `_JAVA_OPTIONS` so that these options aren't inadvertently overridden by a consuming image,

If that's the case, it sounds like `_JAVA_OPTIONS` is a better choice?

briandealwis:
> Is there a reason to suppress the message?

It's unsightly :-)  It's also a change from what we were doing before.  If somebody really doesn't like it though, they should be able to unset the variable.

PR #302, opened 2019-02-19
WORKSPACE: Latest Debian snapshot. Update libc6 to

evanj:
Update libc from 2.24-11+deb9u3 -> 2.24-11+deb9u4. This is a
"regular" update, and not a security update, so we still need the hack to
force the latest version to be resolved instead of the older version in
the stretch-security release.

However, this does fix some low priority CVEs that are reported by the
Google Cloud container vulnerability scanner. See:
https://metadata.ftp-master.debian.org/changelogs//main/g/glibc/glibc_2.24-11+deb9u4_changelog

This also updates the following packages to their latest versions:

base-files
curl
libc-bin
libc6
libcurl3
libldap-2.4-2
libldap-common
libsqlite3-0
openjdk-11-jre-headless
tzdata

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -28,8 +28,8 @@ dpkg_src(
     name = "debian_stretch",
     arch = "amd64",
     distro = "stretch",
-    sha256 = "4b981bd2445d85cf1d93ea9f2d2dc235d20543ecdadd0d8065a10793b94eab9b",
-    snapshot = "20190125T092538Z",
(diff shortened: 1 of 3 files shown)


chanseokoh:
Seems like this gets an updated version of OpenJDK. Need to fix the test.

chanseokoh:
Other than that, I think this looks good.

evanj:
Crap I was sure I ran these tests but I clearly screwed it up. Fixing now ...

PR #303, opened 2019-02-19
WORKSPACE: Use https for Jetty: https://repo1.maven.org

evanj:
This appears to be the new "maintained" Maven repository URL. At the very
least it is the URL that Bazel itself uses:

https://github.com/bazelbuild/bazel/blob/master/tools/build_defs/repo/maven_rules.bzl


chanseokoh:
Thanks for the contribution.

PR #305, opened 2019-02-19
Rework #295 to specify container-friendly args in entrypoint

briandealwis:
I have significant misgivings about using `_JAVA_OPTIONS`: as `java` outputs even when the variable is an empty string, and it's hard to unset an environment variable once set:
```sh
$ set _JAVA_OPTIONS=-Xms45m
$ _JAVA_OPTIONS="" java -version
Picked up _JAVA_OPTIONS: 
java version "1.8.0_192"
Java(TM) SE Runtime Environment (build 1.8.0_192-b12)
Java HotSpot(TM) 64-Bit Server VM (build 25.192-b12, mixed mode)
```
I fear this will break tests and tools that expect to produce useful output.

This PR reverts to my first approach in #295 and passes the container-friendly arguments on the command-line, fixing #289.  It's not as inheritable, but it doesn't change behaviour.

### DIFF
--- a/java/BUILD
+++ b/java/BUILD
@@ -17,40 +17,39 @@ cacerts_java(
         packages["zlib1g"],
         packages[jre_deb],
     ],
-    entrypoint = [
-        "/usr/bin/java",
-        "-jar",
-        # We expect users to use:
-        # cmd = ["/path/to/deploy.jar", "--option1", ...]
-    ],
-    env = java_env + {
+    entrypoint = ["/usr/bin/java"] + java_vm_args + ["-jar"],
+    # We expect users to use:
+    # cmd = ["/path/to/deploy.jar", "--option1", ...]
+    env = {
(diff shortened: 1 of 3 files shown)


chanseokoh:
> `$ _JAVA_OPTIONS="" java -version`

You would do `unset _JAVA_OPTIONS` normally.

> I fear this will break tests and tools that expect to produce useful output.

Just mentioning that, looks like printing out the message (when the variable is defined) has been the current behavior of OpenJDK. IMO, for any robust tools should handle this well from the beginning. I'm sure there will be people using the environment variable.

But I'm fine with reverting this.

briandealwis:
It's [hard to unset from a Dockerfile](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#env), and Jib doesn't support unsetting environment variables.

PR #307, opened 2019-02-22
Build Jetty-on-Java 11 images

chanseokoh:
[We install Jetty 9.4.14](https://github.com/GoogleContainerTools/distroless/blob/master/WORKSPACE#L162), and it should be stable on Java 11 runtimes. 

https://www.eclipse.org/lists/jetty-announce/msg00125.html
https://webtide.com/java-updates-jetty-and-the-future/

Verified its actual working with a tiny example.


loosebazooka:
Do we use the same jetty package for both java8 and java11?

chanseokoh:
Yeah, we do. https://github.com/GoogleContainerTools/distroless/blob/master/WORKSPACE#L162 I don't think we need to upgrade it now or use a different version. 9.4.14 should be stable on Java 11 as I said.

briandealwis:
Yeah: Jetty provides Automatic-Module-Name in its jars/bundles.

Issue #309, opened 2019-02-23
Unneeded multistage build in python example


dragetd:
https://github.com/GoogleContainerTools/distroless/blob/3507da42b7c2cbe2f1a17ca38f21659898623860/examples/python3/Dockerfile#L5

Why is a multistage build used here? Couldn't the python script be directly added to the distroless image?

lathama:
I created a PR with a better example or you can see https://lathama.net/git/lathama/Dockerfiles/src/branch/master/distroless/cherrypyexample/Dockerfile for more.

dragetd:
This makes a lot more sense! Thank you for this example. :D

But then again: If you copy *all* site-packages, (`/usr/local/lib/python3.6/site-packages`) you are also copying the (potentially unneeded?) pip-related packages if I understand correctly.

So one would basically have to isolate the dependencies in requirements.txt (or pipenv) and their deps from the rest of python packages to have 'minimal' container in the sense of the distroless project, or? (maybe I am missing something here)

lathama:
There could be some kilobytes of unused libraries copied over but a huge advance in minimal containers.

dragetd:
Alright!

I'd love to give it a spin myself with pipenv instead of venv + pip maybe if nobody else does, but sometime later.

Thanks for the efforts! -> Close

PR #310, opened 2019-02-23
Improve multi-stage Python example for #309

lathama:
Question raised about the example which was incomplete. This commit improves the example some. Based off my example at https://lathama.net/git/lathama/Dockerfiles/src/branch/master/distroless/cherrypyexample/Dockerfile

### DIFF
--- a/examples/python3/Dockerfile
+++ b/examples/python3/Dockerfile
@@ -1,8 +1,14 @@
-FROM python:3-slim AS build-env
+FROM python:3.6-stretch AS build-env
 ADD . /app
 WORKDIR /app
+RUN pip3 install --upgrade pip
+RUN pip3 install -r /requirements.txt
 
+# This is an example of how to install requirements, copy to the distroless
+# image and use them.
 FROM gcr.io/distroless/python3
 COPY --from=build-env /app /app
+COPY --from=build-env /usr/local/lib/python3.6/site-packages /site-packages
+ENV PYTHONPATH=/site-packages
 WORKDIR /app
 CMD ["hello.py", "/etc"]
--- a/examples/python3/Dockerfile
+++ b/examples/python3/Dockerfile
@@ -2,7 +2,7 @@ FROM python:3.6-stretch AS build-env
 ADD . /app
 WORKDIR /app
 RUN pip3 install --upgrade pip
-RUN pip3 install -r /requirements.txt
+RUN pip3 install -r requirements.txt
 
 # This is an example of how to install requirements, copy to the distroless
 # image and use them.


examples/python3/Dockerfile
```diff
@@ -1,8 +1,14 @@
-FROM python:3-slim AS build-env
+FROM python:3.6-stretch AS build-env
```
briandealwis: Why pin specifically 3.6?  And why `-slim` &rarr; `-stretch`?

examples/python3/Dockerfile
```diff
@@ -1,8 +1,14 @@
...
 ADD . /app
 WORKDIR /app
+RUN pip3 install --upgrade pip
```
briandealwis: Would it make sense to run `python -m compileall /app` too?

examples/python3/Dockerfile
lathama: I will have to check but distroless is still built off of stretch right. I was attempting to give some context here as it is only an example. Happy to make changes in the future.

examples/python3/Dockerfile
```diff
@@ -1,8 +1,14 @@
...
+RUN pip3 install -r requirements.txt
```
chanseokoh: (I am not versed in Python.) This directory does not contain `requirements.txt`. Does this work OK without the file?

lathama: Many assumptions here, yes. see the link in opening comment for context. Long term I suspect we need an actual example .py file for this demo along with tons of comments in the Dockerfile.

so adding everything from current project directory to the /app, change dir to the /app then use the projects requirements.txt

examples/python3/Dockerfile
lathama: That could be an option. I don't want to distract the user too much but I do see your point.

examples/python3/Dockerfile
chanseokoh: Sorry, I didn't get it. The example in the opening comment does have the `requirements.txt`: https://lathama.net/git/lathama/Dockerfiles/src/branch/master/distroless/cherrypyexample/requirements.txt

However, as you can see, our Distroless repo example doesn't have the file: https://github.com/GoogleContainerTools/distroless/tree/master/examples/python3

Can you confirm `docker built .` on this example won't fail on this `pip3 install -r requirements.txt` line?

lathama:
My initial change was a quick copy and paste from https://lathama.net/git/lathama/Dockerfiles/src/branch/master/distroless/cherrypyexample to explain the usage a little. If you would like to have a more complete example I can do that but will need to sit down and work it out later this week.

chanseokoh:
Sounds good. Just make sure anyone checking out the repo example and doing `docker build -t .` on this directory won't fail. And the example doesn't need to be complex. Try to keep it simple.

examples/python3/Dockerfile
lathama: I understand what you are asking. I will work up a more complete example later this week.

lathama:
closing for more complete example later

PR #314, opened 2019-02-26
Improve multi-stage Python example for #309 

lathama:
Example multistage build in Dockerfile to show simple byte compile and more complex usage of loading libraries from a requirements.txt

build with
`docker build . --target bytecompile`
`docker build . --target withrequirements`

I did this in a hurry so I may need to squash the merges, please advise @briandealwis and @chanseokoh

Special note for @dragetd I did change the pip command to isolate the libraries for a smaller install


lathama:
:( Travis CI error is

`Adding APT Sources
127.31s$ curl -sSL "https://storage.googleapis.com/bazel-apt/doc/apt-key.pub.gpg" | sudo -E apt-key add -
curl: (7) Failed to connect to storage.googleapis.com port 443: Connection timed out
gpg: no valid OpenPGP data found.
The command "curl -sSL "https://storage.googleapis.com/bazel-apt/doc/apt-key.pub.gpg" | sudo -E apt-key add -" failed and exited with 2 during .
Your build has been stopped.`

chanseokoh:
I think it's an occasional transient error. I've restarted the build (probably you can do it yourself too), and now it's working. However, I see now you have a code formatting issue:

```
************* Module hello
W: 21, 4: Unused import noop (unused-import)
W: 21, 4: Unused import noop (unused-import)
-----------------------------------
Your code has been rated at 9.93/10
```

lathama:
So I guess I need to do a little magic and get that module used. I will work on that now. Please have a look over the other parts.

lathama:
Yeah all built now! Open to any input on this. I could go into great detail and add a README.md in the future. I note that the other examples could use some love as well.

examples/python2.7/Dockerfile
```diff
@@ -1,8 +1,45 @@
 FROM python:2.7-slim AS build-env
```
chanseokoh: Is this stage necessary?

examples/python2.7/Dockerfile
```diff
@@ -1,8 +1,45 @@
...
+#  Add in the application code
+#  Byte-compile Python code to *.pyc
+
 ADD . /app
 WORKDIR /app
+RUN python -m compileall .
+
+# Copy the results into the distroless image.
 
 FROM gcr.io/distroless/python2.7
```
chanseokoh: Is this stage necessary?

lathama: Yes, this is an example of building the *.pyc to be dropped into the distroless image so they are not byte compiled on the fly each time it starts up.  I wanted a demonstration of dropping a *.py in and then creating an image with the byte compiled *.pyc as a simple fist step before showing off the pip install requirements step.

The comment is a little redundant but I felt like adding more comments.

chanseokoh: Hmm... I mean you have the same stages below as `FROM python:2.7-slim AS bytecompile` and `FROM gcr.io/distroless/python2.7 AS compiled`, so I thought this top two stages are unnecessary? They look identical. I don't see them in your Python 3 example either.

lathama: You might be right the github interface was confusing me, one second while I go to the source

lathama: You were correct, thanks for catching that. corrected and pushed, building now. Thanks!

briandealwis:
Thank you for your efforts on this @lathama — I think incorporating `pip` as part of the build is a great idea.  I *think* you've structured your `Dockerfile`s to demonstrate how the individual steps are required in combination to have a working system. But I fear that it's more confusing than helpful.  IMHO, these examples should be working examples that can be copied and pasted to produce a compact runnable image.

I simplified and refactored your python2.7 example below.  I split the copying steps to separate app code (which changes often) from the installed packages (which changes infrequently).  This has faster build times, since the layer with pip-based dependencies is only rebuilt and re-copied when they change.  This also avoids copying in extraneous build files like `BUILD`, `Dockerfiles`, and `requirements.txt`.
```
# Install dependencies
FROM python:2.7-slim AS pip
ADD requirements.txt /tmp
RUN pip install -r /tmp/requirements.txt -t .
RUN python -m compileall .

# Copy and compile the application code
FROM gcr.io/distroless/python2.7 AS compiled
ADD *.py /app/
RUN python -m compileall .

FROM gcr.io/distroless/python2.7 AS withrequirements
WORKDIR /app
# Place infrequently-changing dependencies in a separate layer from
# frequently-changing app code
COPY --from=pip /app /app
COPY --from=compiled /app /app
CMD ["hello.py", "/app"]
```

What would also be nice would be to have some tests to verify the image.  You can see the `examples/java/BUILD` and `examples/java/testdata/` for an example.

lathama:
@briandealwis I actually understand your example less. Maybe it is just my reading of it but I find it more confusing. My example is far from perfect. I did not want to make two separate Dockerfiles because that could be a mess too so I kept it in one. I am very open to ideas here but wanted to quickly offer an example for those new-comers to the project.

For today, would it be better to just keep the requirements example and forgo the byte compile stage?

I will look at the testing examples you mention. On a side note I am working on some Docker health-checks that work well with Distroless

Again, happy to help, keep the suggestions coming.

chanseokoh:
> What would also be nice would be to have some tests to verify the image. You can see the `examples/java/BUILD` and `examples/java/testdata/` for an example.

@briandealwis one thing to note is that the images built by Bazel (`BUILD`) and by `Dockerfile` are diverging. The test example in the Java case is for the image built by Bazel without `Dockerfile`. Not sure if `rules_docker` has a rule that can build an image from `Dockerfile`. (Maybe not?) So that example for testing may not fit nicely here.

briandealwis:
@lathama please read of https://cloud.google.com/solutions/best-practices-for-building-containers#optimize-for-the-docker-build-cache — separating files by the frequency-of-change can greatly improve Docker's ability to cache and reuse layers, with a dramatic decrease in build and push/pull times.  It's the strategy that is embodied in [Jib](https://github.com/GoogleContainerTools/jib/).

@chanseokoh oops, you're right.

lathama:
> A Dockerfile must start with a `FROM` instruction. The FROM instruction specifies the Base Image from which you are building. FROM may only be preceded by one or more ARG instructions, which declare arguments that are used in FROM lines in the Dockerfile.

As an educational example which I feel is the goal we would want to paint the whole picture very clearly for example. 

`RUN pip install -r /tmp/requirements.txt -t /app`
vs
`RUN pip install -r /tmp/requirements.txt -t .`
Advanced users are not the audience for this example in my opinion.

I would use:
`ADD . /app/`
vs
`ADD *.py /app/`
to establish a pattern if the user is interested in having a README or other non .py files installed in there app like static templates, css, js, or the like.

I also note the ticket that I am attempting to address has been closed.

PR #316, opened 2019-02-26
Actually publish Jetty Java 11 images

chanseokoh:
Forgot to add them to the image list at the end in #307.


cloudbuild.yaml
```diff
@@ -125,6 +125,8 @@ images:
   - 'gcr.io/$PROJECT_ID/java/jetty:java8'
   - 'gcr.io/$PROJECT_ID/java/jetty:debug'
```
loosebazooka: this order is a little weird

loosebazooka: eh never mind, I guess it matches the rest.

PR #335, opened 2019-03-25
Add a 'nobody' user and group

### LINKED ISSUE #332: Put a `nobody` user into /etc/passwd
Having a named, non-root user and group in the base image would make non-root use-cases easier.  It seems like a pretty common practice to say "USER nobody:nobody" in a Dockerfile.

thockin:
Fixes #332

### DIFF
--- a/base/testdata/base.yaml
+++ b/base/testdata/base.yaml
@@ -34,7 +34,7 @@ fileExistenceTests:
 fileContentTests:
 - name: 'known users'
   path: '/etc/passwd'
-  expectedContents: ['^root:x:0:0:user:/home:/bin/bash\n$']
+  expectedContents: ['^root:x:0:0:user:/home:/bin/bash\nnobody:x:65534:65534:user:/home:/bin/bash\n$']
 - name: 'known groups'
   path: '/etc/group'
-  expectedContents: ['^root:x:0:\ntty:x:5:\nstaff:x:50:\n$']
+  expectedContents: ['^root:x:0:\nnobody:x:65534:\ntty:x:5:\nstaff:x:50:\n$']
--- a/base/BUILD
+++ b/base/BUILD
@@ -15,10 +15,18 @@ passwd_entry(
     username = "root",
 )
 
+passwd_entry(
+    name = "nobody_user",
+    gid = 65534,
+    uid = 65534,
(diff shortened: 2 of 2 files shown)


chanseokoh:
Please fix the formatting error: https://github.com/GoogleContainerTools/distroless/pull/291#issuecomment-462479158

Let's use the number 65534 instead of 65535. That's the number on my machine, and it only beneficial to avoid 65535: https://superuser.com/questions/706338/why-on-some-systems-max-uid-gid-is-65534-an-not-65535

thockin:
I have changed numbers and run buildifier.

Separate bug: 

When I run `./buildifier.sh`, I get no output and a success error code.

When I run `buildifier -mode=check $(find . -name 'BUILD*' -o -name 'WORKSPACE*' -type f)`, I get a flag on base/BUILD but a success error code.

buildifier.sh is checking for a non-success error code, so it will never work.

As a casual contributor, it's scary to be told "Go wget and run this binary".  It would be nicer if that was more transparent (and in fact if it "just worked" from HEAD -- even a `docker run` in a script...)

chanseokoh:
Seems like a test failure. Please fix.

> Separate bug:

Probably there's something broken about that. Please file a bug. I maintain Java part of distroless, so I honestly don't know very well about the buildifier stuff.

thockin:
test fixed, I think.

base/testdata/base.yaml
```diff
@@ -34,7 +34,7 @@ fileExistenceTests:
 fileContentTests:
 - name: 'known users'
   path: '/etc/passwd'
-  expectedContents: ['^root:x:0:0:user:/home:/bin/bash\n$']
+  expectedContents: ['^root:x:0:0:user:/home:/bin/bash\nnobody:x:65534:65534:user:/hom:/bin/bash\n$']
```
chanseokoh: Test still failing. Probably `/hom:` --> `/home:`.

thockin:
all green

dlorenc:
@thockin there's an image with this change published here: gcr.io/distroless/base:5eb0f8a01ed8046cefc3a7ca9cf5b95bad296dcf

I can tag it as latest soon

thockin:
Thanks all!

PR #338, opened 2019-03-25
Make buildifier be helpful

briandealwis:
Change `buildifier.sh` to run with `-mode=diff` to report suggested changes.  Requires updating to 0.22.0 so `-mode=diff` errors on change.

Example of failure at https://travis-ci.org/GoogleContainerTools/distroless/builds/511137529 (raw log: 
https://api.travis-ci.org/v3/job/511137530/log.txt)

### DIFF
--- a/buildifier.sh
+++ b/buildifier.sh
@@ -16,8 +16,6 @@
 
 buildifier -mode=diff $(find . -name 'BUILD*' -o -name 'WORKSPACE*' -type f)
 if [ $? -ne 0 ]; then
-  if [[ $files ]]; then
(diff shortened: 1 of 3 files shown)


buildifier.sh
```diff
@@ -14,7 +14,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-files=$(buildifier -mode=check $(find . -name 'BUILD*' -o -name 'WORKSPACE*' -type f))
+buildifier -mode=diff $(find . -name 'BUILD*' -o -name 'WORKSPACE*' -type f)
 if [ $? -ne 0 ]; then
   if [[ $files ]]; then
```
jonjohnsonjr: This seems like it should be changed too

PR #341, opened 2019-03-25
Set -diff_command to use command-line diff tool when calling buildifier

chanseokoh:
On my machine, I have `DISPLAY=:0` defined, and seems like `buildifier` tries to be smart to bring up a GUI diff (`/usr/bin/tkdiff`), which is annoying.

```
$ ./buildifier.sh 
buildifier: selecting diff program with the BUILDIFIER_DIFF, BUILDIFIER_MULTIDIFF, and DISPLAY environment variables is deprecated, use flags -diff_command and -multi_diff instead
./base/BUILD:
/usr/bin/tkdiff: 37: exec: /usr/local/old_symlinks/tkdiff: Required key not available
exit status 2
Run 'buildifier -mode=fix $(find . -name 'BUILD*' -o -name 'WORKSPACE*' -type f)' to fix formatting
```

### DIFF
--- a/buildifier.sh
+++ b/buildifier.sh
@@ -14,7 +14,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-buildifier -mode=diff $(find . -name 'BUILD*' -o -name 'WORKSPACE*' -type f)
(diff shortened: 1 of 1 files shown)


chanseokoh:
I think it is better to always use the non-GUI diff rather than sometimes popping up a GUI diff. If someone really wanted a GUI diff, they could execute buildifier themselves, without this script.

PR #349, opened 2019-04-22
upgrade donet to 2.1

salrashid123:
Upgrade `dotnet` to `2.1`
* remove unnecessary `.so` files


ysma500:
Hi, 
Nice work on that PR. I was wondering why you took out the libraries in the build file but added them back in, in dockerfile?

salrashid123:
@ysma500   the PR is still pending since i'm not sure how to handle release versioning in this repo (i.,e if i merge this its going to abrupty upgrade anyone using `distroless/dotnet` from 2.0 to 2.1 

what i can do is atleast starting versioning

```
distroless/dotnet2.1
distroless/dotnet2.1:debug
distroless/dotnet2.0               << same as distroless/dotnet
distroless/dotnet2.0:debug      << same as distroless/dotnet:debug
```

then upgrade after some notification to users in a release guide or something

but how to set that up in bazel rules is getting me into deep water and would need assistence from the maintainers here

---

re: removing the .so files, yeah, i found out they weren't necessary according to the dotnet specs

dind't actually add them back into the dockerfile other than here where i mention you can static link per microsofts docs:

https://github.com/GoogleContainerTools/distroless/blob/3ddccddb4f2ea1f62259c48bcedda0fb9d9ff261/experimental/dotnet/README.md#example-hello-world

ysma500:
For the .so files, thanks for the explaining 😄, makes sense. For the versioning problem, I sadly don't have a solution but I'm hoping the activity here might bring some maintainers who might have insight on how to handle versioning. I'm happy that you've done this because I have a memory leak in an app running in dotnet core 2.0 that has been fixed in dotnet core 2.1. I'll be following this thread to be updated as it is an issue blocking me from using distroless in my workflow for now.

salrashid123:
i tried to rebuild the with the format specified above but unfortunately, there's been a bunch of changes in accepted formats for the `BUILD` and `WORKSPACE` files and it doesn't work anymore.   I've contacted the owners about this pr.  i'm closing it out here but the files remain in git incase maintainers want to pick it up

PR #356, opened 2019-05-10
Update to snapshot from 2019-05-10 to pick up OpenJDK 11.0.3

### LINKED ISSUE #354: Upgrade to Java 11.0.3 and close CVE-2019-2602
Please release an update on gcr.io/distroless-java:11 to upgrade to 11.0.3. There is a CVE out affecting 11.0.2.

briandealwis:
Fixes #354

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -28,8 +28,8 @@ dpkg_src(
     name = "debian_stretch",
     arch = "amd64",
     distro = "stretch",
-    sha256 = "79a66cd92ba9096fce679e15d0b5feb9effcf618b0a6d065eb32684dbffd0311",
-    snapshot = "20190322T151132Z",
+    sha256 = "da378b113f0b1edcf5b1f2c3074fd5476c7fd6e6df3752f824aad22e7547e699",
+    snapshot = "20190510T144732Z",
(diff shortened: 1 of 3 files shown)


WORKSPACE
```diff
@@ -96,9 +96,7 @@ dpkg_list(
         "python2.7-minimal",
         "libpython2.7-stdlib",
         "dash",
-        # Version required to skip a security fix to the pre-release library
-        # TODO: Remove when there is a security fix or dpkg_list finds the recent version
-        "libc-bin=2.24-11+deb9u4",
+        "libc-bin",
```
chanseokoh: Are we good now? Should we also update `"libc6=2.24-11+deb9u4"` above?

briandealwis: Trying to figure out which packages are resolved is tricky.  Turns out we're still pulling down `2.24-1deb9u1`, not `u4`.  I'll restore libc-bin.

briandealwis: The easiest way I've found to discover package versions is to look in `bazel-xxx/external/package_bundle/file/packages.bzl`

briandealwis: Ah can also look at the contents of the files in `/var/lib/dpkg/status.d/` within the image.

PR #367, opened 2019-05-22
Update for compatibility with bazel --incompatible_no_transitive_loads

pcj:
Newer versions of bazel don't transitively export functions from starlark files.  
This change explicitly exports the dpkg functions.


gertvdijk:
I think this was handled in the meantime in ff53751052293faff8efba18e5d3d3055fd6d442

pcj:
SGTM

Issue #235, opened 2018-08-29
RunAs unprivileged user


vic3lord:
Is there a plan to support a non-root user inside base? Usually I'm using `nobody`

mariusgrigoriu:
I believe it is already supported in the sense that you can run a Distroless-based image as some other user. Is your question that the image should default to a non-root user? If yes, then which UID is safe on all systems?

vic3lord:
@mariusgrigoriu yes, I meant as a default value. I think something above 1000 is safe, unless there's something I am missing.

mariusgrigoriu:
I've seen user accounts in the 5000s for GKE/COS. Other systems and their admins may decide on other UID ranges. Seems like a base image should be able to operate under any UID according to the sys admin's preferences.

dlorenc:
I don't think we'll be changing the default of this from 0. Most other official images use 0, and it's easily overrideable in the runtime or in a base image.

chanseokoh:
Opening for discussion from #306.

> Most other official images use 0,

I had a [similar argument](https://github.com/GoogleContainerTools/distroless/issues/306#issuecomment-466113438), but [@HazCod's argument](https://github.com/GoogleContainerTools/distroless/issues/306#issuecomment-476379259) was that those images keep the root mainly due to the existence of a package manager, which doesn't apply to Distroless.

In any case, if we do this, not sure if nobody would be a good fit, considering its nature. (Also nobody's home would be set to /nonexistent), so we may need to add a proper new user?

briandealwis:
I'll add that we recommend avoiding running as root in our [_Best Practices for Operating Containers_](https://cloud.google.com/solutions/best-practices-for-operating-containers#avoid_running_as_root).

My only argument against using `nobody` is that the container user should have a home directory so that there's a place for dotfiles (e.g., `~/.docker/config.json`) to be installed.

mattmoor:
I'd love to see us at a minimum publish `:nonroot` images, which consuming projects (e.g. ko, jib, bazel) can opt over to case-by-case.

cc @jonjohnsonjr Can you TAL?

jonjohnsonjr:
The disastrous python2 -> python3 migration may be our industry's greatest blunder.

pmorie:
I second @mattmoor's statements. Specific use case for me is publishing knative images that don't take root privs. Currently it appears that `ko` will have to inherit this from the base image.

chanseokoh:
[...]

PR #368, opened 2019-05-28
Create non-root images

dprotaso:
Fixes: #235

@mattmoor @dlorenc @jonjohnsonjr I don't have opinions about how things are tagged. Let me know and I can wrap this up

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -5,7 +5,10 @@ load("@io_bazel_rules_docker//container:container.bzl", "container_bundle")
 container_bundle(
     name = "all",
     images = {
+        "gcr.io/{PROJECT_ID}/static:nonroot": "//base:static-nonroot",
+        "gcr.io/{PROJECT_ID}/base:nonroot": "//base:base-nonroot",
         "gcr.io/{PROJECT_ID}/base:debug": "//base:debug",
+        "gcr.io/{PROJECT_ID}/base:debug-nonroot": "//base:debug-nonroot",
         "gcr.io/{PROJECT_ID}/cc:debug": "//cc:debug",
         "gcr.io/{PROJECT_ID}/java:latest": "//java:java8",
         "gcr.io/{PROJECT_ID}/java:8": "//java:java8",
--- a/base/BUILD
+++ b/base/BUILD
@@ -11,7 +11,7 @@ load("@io_bazel_rules_go//go:def.bzl", "go_binary")
 passwd_entry(
     name = "root_user",
     gid = 0,
-    home = "/home",
+    home = "/root",
     info = "root",
     shell = "/sbin/nologin",
     uid = 0,
@@ -29,11 +29,22 @@ passwd_entry(
     username = "nobody",
 )
 
+passwd_entry(
+    name = "nonroot_user",
+    gid = 65532,
+    home = "/home/nonroot",
+    info = "nonroot",
+    shell = "/sbin/nologin",
+    uid = 65532,
+    username = "nonroot",
+)
+
 passwd_tar(
     name = "passwd",
     entries = [
         ":root_user",
         ":nobody_user",
+        ":nonroot_user",
     ],
     passwd_file_pkg_dir = "etc",
 )
@@ -49,6 +60,12 @@ group_entry(
     name = "nobody_group",
     gid = 65534,
     groupname = "nobody",
+)
+
+group_entry(
+    name = "nonroot_group",
+    gid = 65532,
+    groupname = "nonroot",
 )
 
 group_entry(
@@ -70,6 +87,7 @@ group_file(
         ":nobody_group",
         ":tty_group",
         ":staff_group",
+        ":nonroot_group",
     ],
 )
 
@@ -135,6 +153,28 @@ container_image(
     entrypoint = ["/busybox/sh"],
     env = {"PATH": "$$PATH:/busybox"},
     tars = ["//experimental/busybox:busybox.tar"],
+)
+
+# Non-root base images
+container_image(
+    name = "static-nonroot",
+    base = ":static",
+    user = "nonroot",
+    workdir = "/home/nonroot",
+)
+
+container_image(
+    name = "base-nonroot",
+    base = ":base",
+    user = "nonroot",
+    workdir = "/home/nonroot",
+)
+
+container_image(
+    name = "debug-nonroot",
+    base = ":debug",
+    user = "nonroot",
+    workdir = "/home/nonroot",
 )
 
 container_test(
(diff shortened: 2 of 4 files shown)


chanseokoh:
My concern is that the home directory of `nobody` is pointing to `/nonexistent`. I think a non-root user (other than `nobody`) has to have a proper home. A lot of applications depend on the existence of a home directory.

loosebazooka:
Would it need to be an actual user instead of `nobody`? Perhaps `nonroot` with an actual home directory?

mattmoor:
cc @sharifelgamal This is what I was asking for at Kubecon!  thanks @dprotaso

mattmoor:
cc @loosebazooka may want these for Jib as well.

dprotaso:
I added a home directory for the 'nobody' user. I noticed running tests don't work on Mac

```
+ base/certs_test.image --norun
Loaded image ID: sha256:101cd50a552664199d987a2506fd3bcfecc301019c462adc1a4c98eab217309f
Tagging 101cd50a552664199d987a2506fd3bcfecc301019c462adc1a4c98eab217309f as basecheck_certs_image:intermediate
+ ../structure_test_linux/file/downloaded version
/private/var/tmp/_bazel_dprotasowski/d4d2a8e3e6c0766b8b5af3d4f8a07ed0/sandbox/darwin-sandbox/84/execroot/distroless/bazel-out/k8-fastbuild/bin/base/certs_test.runfiles/distroless/base/certs_test: line 7: ../structure_test_linux/file/downloaded: cannot execute binary file
```

briandealwis:
@dprotaso run with `--cpu=darwin`

Provocative question: why don't we make all images be non-root?  Is there any reason to have root-based images?

loosebazooka:
I'm still not sure if we should reuse `nobody` for this non-root user. I would expect nobody to have no file ownership. It seems odd to diverge from this convention.

dprotaso:
@loosebazooka what was the original purpose of the 'nobody' user?

chanseokoh:
AFAICT, it is sometimes advised to run some processes as `nobody`, but I do not think it is for running any general applications. Besides, I think there is no reason to be obsessed with `nobody` here. You can always create and use a proper user for non-root purposes.

loosebazooka:
@dprotaso mostly just historical context. People used it to isolate daemons, but then moved off that to other daemon-specific low-privilege users to isolate the daemons from eachother. 

I can't really speak to how users are using it now though in distroless, as it doesn't seem like anyone has ever needed a home directory for `nobody`. I think the convention of `nobody` has been for it to be the lowest of low privileged users though, without the ability to write to anything (how did they write logs?).

Maybe I'm being too pedantic about the purpose of `nobody` and this implementation is fine?

loosebazooka:
Sorry, I think I may have misunderstood the question: other context: nobody was added in #335 because of #332, and I do not have enough context but @thockin might have something to add here.

dprotaso:
So outstanding questions for me:

#### Which user/group do folks want to be default in non-root images

1) `nobody` user with a home directory
2) `nonroot` with their own home directory (will use uid:gid 65533 - one lower than nobody)

Note the `/root` home directory changes with both - I'm unaware if this haves any effects downstream. 

#### Image tags

How should we reference/upload non-root images to/from a registry. We can tweak tags or repo path. My preference is for change the repo path

ie. currently we have
```
gcr.io/distroless/static
gcr.io/distroless/base
gcr.io/distroless/base:debug
```

becomes (need to update the PR once consensus is made)

```
gcr.io/distroless/nonroot/static
gcr.io/distroless/nonroot/base
gcr.io/distroless/nonroot/base:debug
```

patflynn:
It seems sad to make the less secure option the default. I imagine the concern here is breaking existing users so I'm wondering if we have other options.

chanseokoh:
> `nonroot` with their own home directory (will use uid:gid 65533 - one lower than nobody

Others may have different opinions, but I prefer this. The home of `nobody` is currently `/nonexistent`, and `nobody` should mean "nobody", not a real user. I don't really like breaking the convention by having a usable home directory for it. And as I said, I think there is no reason to be obsessed with `nobody`.

For the uid:gid, I vote for `65432` for safety. I think `65533` has an increased chance of collision than `65432` (just like you picked it) and is too close to `65535`.

> Note the `/root` home directory changes with both - I'm unaware if this haves any effects downstream.

We could have `/nonroot` as a home directory for the new user instead of `/home/nonroot` to keep backward compatibility, but I'm not sure if that's a good idea either.

mattmoor:
@patflynn making nonroot the default would immediately break anyone depending on it (and there are some niche cases that need it today).  If we were starting from scratch (no pun intended), armed with the knowledge we have now, then this would be the default.

IMO we should use tags to distinguish these, not repos.  `:root` (same as `:latest`), and `:nonroot` (this PR).  For `:debug` and `:nonroot`, I'd just pick a blending: `:nonrootdebug`.

If we get these published, and folks (ko, bazel, jib, kubernetes, ...) adopt the explicit tagging by default, then we can see if `:latest` is still being pulled in a few months and flip the default.

cc @tallclair @lizrice will be interested in this thread as well

lizrice:
I love that you’re doing this! 

+1 to @chanseokoh’s comment preferring a real user rather than nobody. It feels wrong to have a home directory for the nobody account; and in the limited occasions where people deliberately use the nobody account they might even be relying on there not being a home for it.

I think for back-compatibility you're right to keep the :latest tag, but you could document it prominently as deprecated to encourage people to use the :root and :nonroot tags

loosebazooka:
I think we might still need to decide if we want to move user:`root`'s home dir to `/root` ?

briandealwis:
> @mattmoor: If we get these published, and folks (ko, bazel, jib, kubernetes, ...) adopt the explicit tagging by default, then we can see if :latest is still being pulled in a few months and flip the default.

Would it be worth setting up an distroless-announce@ mailing list?  Whether we change the default now or in several months, it's still going to break anybody depending on root who isn't aware of the change.

mattmoor:
@briandealwis I wasn't suggesting that we shouldn't socialize the transition, we should, but it probably makes sense to get any known things onto explicit tags first.

I'd love to get this in.  @dprotaso ping me if you need anything.

loosebazooka:
@mattmoor, any idea what to do with the user-home dir? do we expect users to explicitly be referencing user home instead of through the env?

mattmoor:
@loosebazooka `$HOME`

mattmoor:
Unless you are one of those `%USERPROFILE%` types 🙄

jeremyje:
Once this PR goes in how long will it take for the images to be available?

mattmoor:
@sharifelgamal ^^

dprotaso:
@donmccasland maybe?

sharifelgamal:
Sorry for the delay, no idea how this fell off my radar.

sharifelgamal:
These should be live in 15 minutes or so.

dprotaso:
@sharifelgamal thanks!

PR #370, opened 2019-05-29
fix container_tests on darwin

dprotaso:
Go binary was being compiled with `darwin` GOOS. We should use Linux since this binary runs in a Linux Container.

### DIFF
--- a/examples/nonroot/BUILD
+++ b/examples/nonroot/BUILD
@@ -40,6 +40,7 @@ container_image(
 go_binary(
     name = "user",
     srcs = ["testdata/user.go"],
+    goos = "linux",
     pure = "on",
 )
 
--- a/examples/nonroot/BUILD
+++ b/examples/nonroot/BUILD
@@ -40,6 +40,7 @@ container_image(
 go_binary(
     name = "user",
     srcs = ["testdata/user.go"],
+    # Test image is linux based
     goos = "linux",
     pure = "on",
 )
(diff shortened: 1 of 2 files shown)


dprotaso:
@briandealwis FYI

briandealwis:
Good find on the missing `goos`!  It looks like `examples/nonroot` needs a similar `goos`.

(I haven't been able to run the full test suite from my Mac due to some oddness around the Python tests involving `$TMPDIR`, but I realized explicitly setting `TMPDIR=/tmp` works around that.)

Can you back out the verbose changes?  I don't understand why the underlying `container-structure-tests` is run by default with `--quiet`, but that should be taken up with `bazel_build/rules_docker`.

briandealwis:
Filed bazelbuild/rules_docker#863 for verbose flag

examples/nonroot/BUILD
```diff
@@ -40,6 +40,7 @@ container_image(
 go_binary(
     name = "user",
     srcs = ["testdata/user.go"],
+    goos = "linux",
```
briandealwis: I think it's worth copying your comment here too.
```suggestion
    # Test image is linux based
    goos = "linux",
```

chanseokoh:
The CLA bot cannot handle this properly because there is a commit co-authored by @briandealwis. I will manually confirm this.

chanseokoh:
Huh, there is no `cla:yes` label in this repo. This feels like a deja vu. Can't remember how we resolved this kind of this.

chanseokoh:
Thanks @dprotaso.

dprotaso:
np

PR #375, opened 2019-06-22
Use numeric user

mattmoor:
cc @dprotaso


mattmoor:
Apparently PodSecurityPolicy only works with numeric users 🙄

mattmoor:
According to Knative users with PSP, here's what we see without this:
```
Error: container has runAsNonRoot and image has non-numeric user (nonroot), cannot verify user is non-root
```

Issue #225, opened 2018-08-01
uclibc BusyBox in Distroless debug images fails with "ls: Value too large for defined data type"


xinau:
*UPDATE by @chanseokoh: this issue is specifically about the pre-built, buggy BusyBox binary compiled using uClibc embedded in the Distroless `:debug` Docker images (e.g., [`gcr.io/distroless/java:debug`](https://gcr.io/distroless/java:debug)). The root cause is [Bug 11651](https://bugs.busybox.net/show_bug.cgi?id=11651) in BusyBox.*

---

It seems like LargeFileSupport is disabled in the debug images.
```console
$ docker run --entrypoint=/busybox/sh --rm -it gcr.io/distroless/base:debug 
/ # ls /
ls: can't open '/': Value too large for defined data type
```
According to the [coreutils docs](https://www.gnu.org/software/coreutils/faq/#Value-too-large-for-defined-data-type) this error is caused by not enabling LargeFileSupport when compiling.

The default docker image doesn't seem to have this problem
```console
$ docker run --entrypoint=/bin/sh --rm -it busybox
/ # ls
bin   dev   etc   home  proc  root  sys   tmp   usr   var
```

dlorenc:
cc @tejal29

tejal29:
hmm not able to reproduce this
```
docker run --entrypoint=/busybox/sh --rm -it gcr.io/distroless/base:debug 
Unable to find image 'gcr.io/distroless/base:debug' locally
debug: Pulling from distroless/base
8f125ded1b48: Pull complete 
8ca8724e5ffb: Pull complete 
Digest: sha256:333a18407c0e2d46b264a240175284b3bd2482892f4273bacb63a5d37a7adea2
Status: Downloaded newer image for gcr.io/distroless/base:debug
/ # ls
busybox  etc      lib      proc     tmp      var
dev      home     lib64    sys      usr
```

dlorenc:
Same here. @xinau do you have any other info on how you reproduced this?

xinau:
Since I've done several package updates (kernel, util-linux, ...)  in the mean time, they seem to have magically fixed the problem. I can't reproduce it anymore :( . 

Sry for bothering.

dgageot:
I have the same issue on Docker for Mac `Version 18.06.0-ce-mac69 (26398)`

```
docker run --entrypoint=sh --rm gcr.io/distroless/base:debug -c ls
ls: can't open '.': Value too large for defined data type
```

[...]

Issue #352, opened 2019-05-03
Use 64bit versions of binaries


steren:
gVisor is not able to run 32bit binaries. distroless images are using the 32bit version of busybox, as a consequence, images created with distroless are failing on gVisor.

chanseokoh:
Hmm... I get the impression that BusyBox intentionally provides a single pre-compiled binary (for simplicity and manageability for example?) as they say

> The "busybox-i686" version should run on both 32-bit and 64-bit x86 PCs

Technically we may be able to compile a 64-bit binary from source on our own, but I don't know if that is a good idea in terms of reliability and support. Maybe it doesn't matter much as a debug tool.

abergmeier:
The problem is not only gVisor. We currently encounter also https://github.com/GoogleContainerTools/distroless/issues/225 on our Kubernetes cluster.
Which makes the debug image to a large degree unusable.
IIUC #225 should be fixed by 64-bit build.

chanseokoh:
Fixed by #513.

PR #378, opened 2019-07-02
Make busybox package public

dprotaso:
Was experimenting creating a different distribution (ie. ubuntu) and realized in order to have debug images we need this package to be public

Right now I just copied the rules
https://github.com/dprotaso/distroless-ubuntu/tree/74094d7189c8b189208b867e72fb9e3252df4821/bionic/busybox

### DIFF
--- a/experimental/busybox/BUILD
+++ b/experimental/busybox/BUILD
@@ -1,4 +1,4 @@
-package(default_visibility = ["//base:__subpackages__"])
+package(default_visibility = ["//visibility:public"])
 
 load("@io_bazel_rules_docker//contrib:test.bzl", "container_test")
 load(":commands.bzl", "BUSYBOX_COMMANDS")


dprotaso:
Whoops - updated commit message to say busybox

chanseokoh:
I know there have been discussions around whether Distroless should ditch BusyBox and, for example, replace it with ToyBox. But I am not sure of the status or the decision.

dprotaso:
I have no preference - is there a corresponding issue for that discussion?

chanseokoh:
On the technical side, we have #225 and #352.

I don't see any problem replacing BusyBox with ToyBox. License seems better too. It is included in Android. (Actually, ToyBox is explicitly targeting Android to achieve Android's self-hosting goal.) If it resolves #225 and #352, I think we should go for it.

dprotaso:
>  I don't see any problem replacing BusyBox with ToyBox. 

Looks like there's some TODOs - http://landley.net/toybox/status.html

Does the discussion of BusyBox vs. ToyBox block this PR?

chanseokoh:
I don't know, honestly. You said you'll build Ubuntu images. Where do you plan to use them? Will you distribute them so that others can base on them? I'm wondering if we shouldn't stop spreading BusyBox.

dprotaso:
> Where do you plan to use them? Will you distribute them so that others can base on them?

I was just doing a proof of concept and had to copy the the busybox rules etc. which was less than ideal.

> I'm wondering if we shouldn't stop spreading BusyBox.

My suggestion is make the choice for the images that are created from this repository.  But leave the option open for creators of other distributions. Hence I'd open up the visibility of the busybox package and include a new toybox one (in another PR perhaps)

chanseokoh:
As you said, I think you still have the option to copy the rules as you see fit?

There is a non-technical reason that Android is bound to ToyBox. (Forgive me not explaining here what the reason is.) I'm just concerned that changing the visibility of the BusyBox rules might make the situation worse (_if the current situation is not what we think should be, I don't know_).

dprotaso:
> As you said, I think you still have the option to copy the rules as you see fit?

Sure but that's error prone & there's work maintaining the copies.

> There is a non-technical reason that Android is bound to ToyBox. (Forgive me not explaining here what the reason is.) I'm just concerned that changing the visibility of the BusyBox rules might make the situation worse (if the current situation is not what we think should be, I don't know).

I don't think it can get any worse - distroless debug images include busybox. Distroless images are used in the bazel docker_rules. These docker_rules are used everywhere.

chanseokoh:
> I don't think it can get any worse - distroless debug images include busybox. Distroless images are used in the bazel docker_rules. These docker_rules are used everywhere.

I can agree. Your reasoning looks sound and may be right. Most likely this won't make the current situation worse, I presume. However, I'm just concerned that a conclusion drawn out of engineers only may not justify some actions. Probably I'm just paranoid here. In any case, I don't think I am the right person to take the responsibility for approving this. I didn't intend to review this when I saw the title of this PR. I just chimed in to leave my opinion, just in case. I'm leaving this to someone else for review and approval.

dprotaso:
@chanseokoh are you able to assign somebody?

chanseokoh:
For this repo, I sometimes (not often) assign somebody if I definitely know who can/may review, but for this one, I'm afraid I am unsure. I'm mostly interested in the Java part of this repo, so I usually don't care much about other parts unless I got a sudden appetite.

loosebazooka:
I think in general a change like this supporting an external forks usecase seems odd, there seems no obvious reason to keep this public within the confines of *this* codebase. If you're already maintaining a fork, could you not just have this specific change in there?

Also, I believe @donmccasland and @sharifelgamal are the correct owners.

dprotaso:
For context I'm not maintaining a fork but composing my own distro using
rules in this repository

eikemeier:
> I don't see any problem replacing BusyBox with ToyBox.

AFAICS toybox has no `sh` yet, it's currently [in “pending”](https://github.com/landley/toybox/blob/master/toys/pending/sh.c).

Android [is using mksh](https://android.googlesource.com/platform/system/core/+/refs/heads/master/shell_and_utilities/README.md), so it seems that for the intended purpose toybox is not just or drop-in replacement for BusyBox.

chanseokoh:
While doing #437, I got to understand what this change means. I believe Distroless is in no way interested in making its own private Bazel rules public and maintain them so that other Bazel projects can import to use. That is not the purpose of this repo. Closing.

Issue #353, opened 2019-05-08
Update openssl to 1.1.1 to support TLS1.3


yegle:
Can we get openssl 1.1.1 in the image so that binaries that depend on OpenSSL can get TLS 1.3 support?

yegle:
Possibly related: https://github.com/GoogleContainerTools/distroless/issues/326

briandealwis:
openssl is still at 1.1.0 in stretch

https://tracker.debian.org/pkg/openssl

yegle:
Sounds like you should maintain additional versions that maps to Debian's stable/testing? I.e. `gcr.io/distroless/base:stable` and `gcr.io/distroless/base:testing`.

The current testing (Buster) is entering full freeze and the next stable release is imminent. It probably a good idea to at least start the planning of using packages from Debian Buster.

kalexmills:
It sounds like maintaining additional versions of the base image which point to stable/sid upstreams could potentially be used to resolve #364 as well.

yegle:
FYI Debian Burster was promoted from testing to stable today. When should we expect the distroless to be updated to the versions in Debian Burster?

donmccasland:
We have now released debian10 based images.  

openssl is still at 1.1.0 in stretch

https://tracker.debian.org/pkg/openssl

chanseokoh:
We just set up GitHub Actions to automate pulling in new deb packages updated in the Debian snapshot upstream: #537.

PR #380, opened 2019-07-09
Use libc based BusyBox from Debian distribution

eikemeier:
Debug images currently use a version of BusyBox which is statically linked with uClibc.

Firstly, debug images contain libc anyway, so there is no gain in using a statically compiled BusyBox variant.

Also, this results in issues #352 and #225, the later caused by [a bug](https://bugs.busybox.net/show_bug.cgi?id=11651#c13) in uClibc.

Any easy solution would be to use BusyBox from the Debian distribution, like most of the other parts installed in the images.

Things that might be improved in this patch:
- I've linked the aliases where `busybox --list-full` would place them, this might conflict with dependent debug images, but has the advantage that eventual debug scripts would work better
- I've commented out some commands that IMHO make no sense in a container. This should be reviewed
- Creating a tar just for the links might be inelegant, perhaps `symlinks` from `container_image` should be used

Note that this downgrades BusyBox from 1.27.1 to 1.22.1, which has less available commands (xxd for example). Upgrading could be done in concert with #353 by using a more recent Debian base.


dmoebius:
Please review and proceed with this CR. It also solves the problem of debug images not having `/bin/sh` available, which makes the OpenShift web terminal console useless.

WORKSPACE
```diff
@@ -183,14 +186,6 @@ http_archive(
...
 )
 
-# For the debug image
-http_file(
-    name = "busybox",
-    executable = True,
-    sha256 = "b51b9328eb4e60748912e1c1867954a5cf7e9d5294781cae59ce225ed110523c",
-    urls = ["https://busybox.net/downloads/binaries/1.27.1-i686/busybox"],
-)
-
```
Alcatros, line 193: Why do we remove the version checks? We seem to check for version 22 below, we should insert here the right sha256 check

chanseokoh: This PR just gets busybox from theh Debian snapshot repo rather than downloading the binary directory from busybox.net. And Distroless does verify all checksums when downloading Debian packages. Marking resolved.

floriankoch:
why not using toybox instead? Its used in android and seems better maintained

chanseokoh:
@floriankoch no specific reason. If ToyBox works, we can switch. Someone should contribute though.

chanseokoh:
Made obsolete by #513. #513 pulls in a 64-bit musl-libc BusyBox and fixes both #352 and #225. Thanks for your help @eikemeier.

PR #391, opened 2019-07-28
Experimental Debian 10 support

evanj:
This prototypes adding Debian 10 support for the following images:

//base:base-debian10
//cc:cc-debian10
//experimental/python3:python3-debian10

This is not complete.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -243,3 +243,134 @@ load(
 )
 
 _go_image_repos()
+
+dpkg_src(
+    name = "debian10",
+    arch = "amd64",
+    distro = "buster",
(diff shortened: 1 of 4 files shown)


evanj:
One random note: It would seem to me that it would be very good for this repository to have some tests that actually run code inside the built containers, to find things like "import ctypes" not working, or SSL not actually working. I understand this is a bit tricky and might make this repository build only work correctly on Linux, but boy it would sure help to make sure the images aren't broken.

JoostvDoorn:
@donmccasland Can `cloudbuild.yaml` be updated to push these new debian 10 packages?

donmccasland:
Opened https://github.com/GoogleContainerTools/distroless/issues/402 for this.

PR #396, opened 2019-08-09
Bazel 0.28 fixes

lukealonso:
Symbols loaded by load() are no longer visible externally, so they need to be re-exported
with an assignment.


lukealonso:
@googlebot I signed it!

PR #397, opened 2019-08-22
Upgrade node image to latest LTS v10

HofmannZ:
Node v8 has moved to maintenance LTS and v10 has moved to active LTS. Version `10.16.3` is now the latest active LTS version of Node js.

> It was [released](https://nodejs.org/download/release/latest-v10.x/) on the 15th of August. View the list of [deprecations](https://nodejs.org/en/blog/release/v10.0.0/#deprecations) from node 8 to node 10.


HofmannZ:
@googlebot I signed it!

Issue #385, opened 2019-07-23
Updating to Debian 10 (buster)


evanj:
Before I attempt to spend some time to implement this: Can a maintainer provide some guidance about what the "ideal" way to manage this should be? In particular: do we want to maintain both debian9 and debian10 based images? Do we want to switch the default? My proposal:

* Produce sets of images tagged with both `debian9` and `debian10` respectively. E.g. something like `gcr.io/distroless/base:debian9` and `gcr.io/distroless/base:debian10`.
* Leave the default as debian9, at least for now.

Any thoughts? This will also at least partially address #364 and possibly a few others.

chanseokoh:
@donmccasland @sharifelgamal Debian 10 "buster" released in July

chanseokoh:
Oh, wait. I heard Buster doesn't have `openjdk-8` anymore.

evanj:
Looks like it is only [`openjdk-11`](https://packages.debian.org/buster/openjdk-11-jre)? Is that problematic? With the current Java image naming convention, this could just add `gcr.io/distroless/java:11-debian10`?

loosebazooka:
I believe the distroless philosophy is to stay on a single latest version, so naming is something that'll probably have to be worked on a little.

evanj:
I have an example of the sort of thing I am thinking about in https://github.com/GoogleContainerTools/distroless/pull/391. Feedback welcomed!

donmccasland:
Hey folks,

I'm glad we've got some interest going on this.  I'd suggest we get a bug
going to track our progress and final decisions about it.

Because we don't have a solid deprecation plan and process for these
images, I'd like us to plan on producing both a Debian 9 and Debian 10 for
a period of time while clearly indicating somewhere that Debian 9 will
eventually be gone.

chanseokoh:
I guess we will not have distroless/java:8 for Debian 10, unless we get Java 8 from elsewhere.

evanj:
My experimental PR #391 was merged, so this is partially fixed. We now need to get them published. I've opened #398 which I think will cause the `//base` images to be published. If this approach is acceptable, I'll go ahead and do the same thing for the Python and Java images, and we can work on the other ones in a similar way. I don't use the other images, so I may need help with those.

evanj:
[...]

PR #398, opened 2019-08-25
base: Add -debian10 for all images; include in all to publish

evanj:
Experimental Debian 10 support was added in commit ee7342d305, but was
missing -debian10 variants for some of the images in base. This adds
these variants for all the images in the //base package. It also adds
them to the //:all target, which might mean they get published.

base.bzl: Use a function to define the repeated images/tests. This
  should be slightly easier to understand than a list comprehension.

Another step towards resolving #385. If this approach makes sense, I'll work on rolling it out to the other packages. We only really use the `base` and `python` images though, so I'll need help verifying that the others do in fact "work" as expected.


BUILD
```diff
@@ -5,10 +5,17 @@ load("@io_bazel_rules_docker//container:container.bzl", "container_bundle")
 container_bundle(
     name = "all",
     images = {
-        "gcr.io/{PROJECT_ID}/static:nonroot": "//base:static-nonroot",
-        "gcr.io/{PROJECT_ID}/base:nonroot": "//base:base-nonroot",
         "gcr.io/{PROJECT_ID}/base:debug": "//base:debug",
+        "gcr.io/{PROJECT_ID}/base:debug-debian10": "//base:debug-debian10",
```
yegle: Nit: it might be better to suffix all old images with `-debian9` (but keep non-suffixed labels). In the future the non-suffixed version can be deprecated.

donmccasland: I would prefer to leave the current tagging unchanged until debian10 moves out of experimental.

PR #411, opened 2019-10-08
Mention how to append the debug tag when an image has a tag already

shaobos:
Hi Distroless, 

Thanks for documenting the `:debug` option to allow troubleshooting with a shell, very helpful. It wasn't very obvious to me from the README.me how to apply the debug tag to an image that already has a tag, like `gcr.io/distroless/java:11`. 

I thought adding a dash would help and it worked, but I'm not sure if this is the recommended way to do this. Opening this PR to start some discussion.

### DIFF
--- a/README.md
+++ b/README.md
@@ -178,5 +178,6 @@ $ docker run --entrypoint=sh -ti my_debug_image
 /app # ls
 BUILD       Dockerfile  hello.py
 ```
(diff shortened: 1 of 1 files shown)


shaobos:
@googlebot I signed it!

shaobos:
@loosebazooka yeah it probably has not been updated in a while. I think adding this one-liner will help people like me save a few minutes figuring out how to construct the tag.

PR #412, opened 2019-10-12
package_manager.bzl: Use newest version to support Debian 10

evanj:
Corrects the VERSION variable in /etc/os-release

Commit c127f3e5f3 added support for Debian 10 (buster) to
package_manager. Due to the way WORKSPACE dependencies work, we
need this update to actually use the new version.

I also needed to update the version of rules_docker to avoid a Python3
error. This could be because I'm using the latest version of bazel,
but seems like a good idea also.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,7 @@ jdk:
 install:
   - export PATH=$PATH:$HOME/bin && mkdir -p $HOME/bin
   - wget https://github.com/bazelbuild/buildtools/releases/download/0.22.0/buildifier && chmod +x buildifier && mv buildifier $HOME/bin/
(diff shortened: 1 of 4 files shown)


evanj:
This build and the master build are failing due to a bug with debian-snapshot: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=941903

I've been able to work around it by adding an `/etc/hosts` record with the "good" IP from that bug report: 193.62.202.27 . Since one IP works and one doesn't, retrying a build until Bazel's cache "gets lucky" and manages to get all the files also sort of "works".

evanj:
Ha, I just rebased the change a bunch of times and I got lucky with the download so the tests passed. Annoying. It did reveal that the latest rules_docker seems to require a version of bazel newer than 0.28.1, so I am also including an update to Bazel 1.0.0.

PR #413, opened 2019-10-12
java: Add _debian9 and _debian10 package versions

evanj:
This adds Debian 10 support for the Java packages. I did not perform
any testing, beyond ensuring that bazel test //... works, since I do
not actually use Distroless for any Java packages myself.

Since Debian 10 (buster) only includes java11, this only changes it
for those packages.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -280,15 +280,14 @@ dpkg_list(
 
         #java
         "zlib1g",
-        # "libjpeg62-turbo",
-        # "libpng16-16",
-        # "liblcms2-2",
-        # "libfreetype6",
-        # "fonts-dejavu-core",
-        # "fontconfig-config",
-        # "libfontconfig1",
-        # "openjdk-8-jre-headless",
-        # "openjdk-11-jre-headless",
+        "libjpeg62-turbo",
+        "libpng16-16",
+        "liblcms2-2",
+        "libfreetype6",
+        "fonts-dejavu-core",
+        "fontconfig-config",
+        "libfontconfig1",
+        "openjdk-11-jre-headless",
 
         #python
         "dash",
--- a/java/BUILD
+++ b/java/BUILD
@@ -1,29 +1,43 @@
 package(default_visibility = ["//visibility:public"])
 
 load("@io_bazel_rules_docker//container:container.bzl", "container_image")
+load("@io_bazel_rules_docker//contrib:test.bzl", "container_test")
 load("@io_bazel_rules_docker//java:image.bzl", "java_image")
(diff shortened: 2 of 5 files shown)


evanj:
I was a bit sad to discover that the Java packages are using underscores to separate the name parts e.g. (`java8_debug`), and we are using a dash to separate the distribution version :( This would be a good chance to try to standardize this, if we wanted to.

I'm pretty sure this build and the master build are failing due to a bug with debian-snapshot: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=941903 ; causing not found errors while downloading debian packages

chanseokoh:
I believe it is conventional in Bazel you use underscores in names. All the names in this repo follows this convention, not just the rules for Java.

evanj:
Yeah ... I should probably fix that then :( I think I may have screwed this up since Docker typically uses dashes (e.g. `golang:1.13-buster`). I will prepare yet another PR with the appropriate edits, but it will likely conflict with basically every other PR :)

chanseokoh:
Sounds good, thanks. Note that Java uses dashes for Docker tags, while using underscores for Bazel names.

donmccasland:
Is this still valid in light of #416?

evanj:
I'll rebase everything to the new names.

evanj:
Okay, this is now rebased on top of master to use the new names.

stuartgunter:
It's great to see this has been done! Really appreciate it!

Is there a plan to publish the new images to gcr.io? (sorry if I missed where this has already been stated)

evanj:
They are there, with `-debian10` suffixes; e.g. https://gcr.io/distroless/base-debian10

See also a discussion about possibly reorganizing these image names: https://groups.google.com/d/msg/distroless-users/Uufl6cxWgMI/3AzFbONRAAAJ

stuartgunter:
Ah, I thought this was the PR for the `distroless/java` image (https://github.com/GoogleContainerTools/distroless/issues/385#issuecomment-541543975).

When I look at the gcr.io/distroless registry, I only see `-debian10` suffixes for `base`, `cc`, `python*` and `static` images at the moment - but `java` is still just `java`.

Sorry if this is a silly question or if I'm just getting confused between issues.

chanseokoh:
I planned to publish Java 11 Debian 10 images in #435. It has a minor technical issue that needs some help from the release team though. @donmccasland

donmccasland:
Hrmms, this looks like it broke the java8 build.  @evanj can you take a look?

chanseokoh:
I don't think it's broken. I can build locally at HEAD.

PR #415, opened 2019-10-12
README.md: Update to reference Debian 9 and 10

evanj:
The documentation refers to Debian 9 (stretch) as stable, but it is now
oldstable. Add a note that users should explicitly build with the right
"base" Debian distribution corresponding to their distroless images.


evanj:
This README update is a bit forward looking, since the Debian 10 images don't have as much testing, but this seems like the only path forward, so we might as well document it that way?

PR #416, opened 2019-10-13
Rename -debian(9|10) to _debian(9|10) to match Bazel conventions

evanj:
This does not change the Docker tag names in cloudbuild.yaml but does
change all the Bazel target names. We should repeat the same exercise
with -nonroot in a future change.


evanj:
This came up in discussion with PR #416, and will conflict with that change. Either this change or that change will need to be merged first, and then I'll need to rebase the other.

base/BUILD
```diff
@@ -113,73 +113,73 @@ go_binary(
...
 
 alias(
     name = "debug-nonroot",
-    actual = ":debug-nonroot-debian9",
+    actual = ":debug-nonroot_debian9",
 )
 
 container_test(
-    name = "base_release_test-debian9",
+    name = "base_release_test_debian9",
```
chanseokoh: Unrelated, but test names usually end with `..._debian9_test`.

cloudbuild.yaml
```diff
@@ -3,92 +3,92 @@ steps:
...
-  args: ['tag', 'bazel/base:debug-nonroot-debian9', 'gcr.io/$PROJECT_ID/base-debian9:debug-nonroot']
+  args: ['tag', 'bazel/base:debug-nonroot_debian9', 'gcr.io/$PROJECT_ID/base-debian9:debug-nonroot']
 - name: gcr.io/cloud-builders/docker
   args: ['tag', 'gcr.io/$PROJECT_ID/base-debian9:debug-nonroot', 'gcr.io/$PROJECT_ID/base:debug-nonroot']
 - name: gcr.io/cloud-builders/bazel
-  args: ['run', '--host_force_python=PY2', '//base:debug-nonroot-debian10']
+  args: ['run', '--host_force_python=PY2', '//base:debug-nonroot_debian10']
 - name: gcr.io/cloud-builders/docker
-  args: ['tag', 'bazel/base:debug-nonroot-debian10', 'gcr.io/$PROJECT_ID/base-debian10:debug-nonroot']
+  args: ['tag', 'bazel/base:debug-nonroot_debian10', 'gcr.io/$PROJECT_ID/base-debian10:debug-nonroot']
```
chanseokoh: These are tags, so they should use hyphens.

BUILD
```diff
@@ -6,16 +6,16 @@ container_bundle(
     name = "all",
     images = {
         "gcr.io/{PROJECT_ID}/base:debug": "//base:debug",
-        "gcr.io/{PROJECT_ID}/base:debug-debian10": "//base:debug-debian10",
+        "gcr.io/{PROJECT_ID}/base:debug_debian10": "//base:debug_debian10",
```
chanseokoh: The names in the front that starts with gcr.io are Docker image references, so tags should continue to use hyphens. But the names in the end are Bazel targets, so it is correct to use underscores.

base/BUILD
```diff
@@ -113,73 +113,73 @@ go_binary(
...
+distro_components("_debian9")
 
-distro_components("-debian10")
+distro_components("_debian10")
 
 # alias debian9 as the default images
 alias(
     name = "static",
-    actual = ":static-debian9",
+    actual = ":static_debian9",
```
chanseokoh: Do these aliases refer to Docker image names or Bazel target names?

cloudbuild.yaml
chanseokoh: Oh, but l think the tag names here are automatically derived from the Bazel targets. Then this makes sense, I guess?

evanj: That is my understanding: The bazel docker rules automatically generate those tag names; I'm assuming `cloudbuild.yaml` is what publishes the production packages, so the final `gcr.io` tag names are unchanged. I hope?

BUILD
evanj: Nice catch thank you will fix!

base/BUILD
evanj: Nice catch! I tried to correct some of the other ones in this PR so I'll add this one.

base/BUILD
evanj: These are Bazel aliases, which are there to ensure that if anyone is depending on these images from native Bazel rules, they don't need to update: `//base:static` is now an alias for `//base:static_debian9`. They are also used by the packages in this repository that have not been updated to be distribution-aware yet (Node, dotnet, Java).

BUILD
```diff
@@ -6,16 +6,16 @@ container_bundle(
...
         "gcr.io/{PROJECT_ID}/base:latest": "//base:base",
         "gcr.io/{PROJECT_ID}/base:nonroot": "//base:base-nonroot",
-        "gcr.io/{PROJECT_ID}/base:nonroot-debian10": "//base:base-nonroot-debian10",
-        "gcr.io/{PROJECT_ID}/static:debian10": "//base:static-debian10",
+        "gcr.io/{PROJECT_ID}/base:nonroot-debian10": "//base:base-nonroot_debian10",
+        "gcr.io/{PROJECT_ID}/static:debian10": "//base:static_debian10",
         "gcr.io/{PROJECT_ID}/static:latest": "//base:static",
         "gcr.io/{PROJECT_ID}/static:nonroot": "//base:static-nonroot",
-        "gcr.io/{PROJECT_ID}/static:nonroot-debian10": "//base:static-nonroot-debian10",
+        "gcr.io/{PROJECT_ID}/static:nonroot_debian10": "//base:static-nonroot_debian10",
```
chanseokoh: Missed these ones.

evanj: Damn it, after I managed to get lucky with the damned snapshot.debian.org bug ... :) Thanks.

donmccasland:
I've approved this, but I'll let you merge it when you're ready.

evanj:
I can't merge things in this repo! You'll have to do it. Why don't you go ahead and merge this one, and I'll rebase #413 .

donmccasland:
Sounds good.  I'll get to it later today.  (On vacation in New York)

evanj:
Ha if you are on vacation don't do it today at all :) (I really should not have been fiddling with this stuff this weekend either)

chanseokoh:
Looking into it.

PR #417, opened 2019-10-14
feat(go): add go module examples + refactor Dockerfile

hazcod:
- Adapt Go example to go modules
- Show a more secure example with `nonroot` user

### DIFF
--- a/examples/go/Dockerfile
+++ b/examples/go/Dockerfile
@@ -1,12 +1,28 @@
-FROM golang:1.12 as build-env
+# our builder image
+FROM golang:1.13.1-alpine AS builder
 
+# default app directory
 WORKDIR /go/src/app
+
+# copy-in our module info
+COPY go.mod go.sum /go/src/app/
+RUN go mod download
+
+# add our binary sources
 ADD . /go/src/app
 
-RUN go get -d -v ./...
+# build our binary
+RUN CGO_ENABLED=0 go build -a -installsuffix cgo -o /go/bin/app
 
-RUN go build -o /go/bin/app
+# ---
 
 FROM gcr.io/distroless/base
-COPY --from=build-env /go/bin/app /
+
+# copy our built binary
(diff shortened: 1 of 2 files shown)


hazcod:
@googlebot I signed it!

examples/go/Dockerfile
```diff
@@ -1,12 +1,28 @@
...
+# ---
+
+FROM gcr.io/distroless/static
+
+# copy our built binary
+COPY --from=builder --chown=nonroot /go/bin/app /app
 
-RUN go build -o /go/bin/app
+# run as unprivileged user
+USER nonroot
```
chanseokoh: Use `gcr.io/distroless/static:nonroot` instead?

hazcod: @chanseokoh I am not a big fan of using tags for build variants, I much prefer explicit security. As long as distroless does not have a clear update/break policy, I keep the tags plain.

examples/go/Dockerfile
```diff
@@ -1,12 +1,28 @@
...
 
-RUN go build -o /go/bin/app
+# ---
 
 FROM gcr.io/distroless/base
-COPY --from=build-env /go/bin/app /
-CMD ["/app"]
+
+# copy our built binary
+COPY --from=builder --chown=nonroot /go/bin/app /app
```
dolmen: `COPY` should be the last command to better use Docker's caching. `USER` and `CMD` never change so they should appear before.

hazcod: Fixed.

examples/go/go.mod
```diff
@@ -0,0 +1,3 @@
+module github.com/GoogleContainerTools/distroless/blob/master/examples/go
```
dolmen: Why `.../blob/master/...`? Why not `github.com/GoogleContainerTools/distroless/examples/go`?

hazcod: Seems like a copy mistake, fixed.

PR #429, opened 2019-10-30
dotnet: Add _debian9 and _debian10 variants

evanj:
These targets have zero tests, so this was not tested other than
ensuring the main command can start:

```
$ docker run -ti --rm bazel/experimental/dotnet:dotnet_debian10 /opt/dotnet/dotnet --info

Microsoft .NET Core Shared Framework Host

  Version  : 2.0.0
  Build    : e8b8861ac7faf042c87a5c2f9f2d04c98b69f28d
```

Someone who actually uses dotnet will need to check that this works.


evanj:
Good point! I thought there was someone in the issue discussing updating to Debian 10, but I was wrong. Looking through the history of this repository: @salrashid123 @Habikki @liguobao do any of use still use distroless for dotnet? How can we check that these images actually work?

I think it is a reasonable policy for us to sit on this until we find some dotnet users. Possibly this means we should just remove dotnet support?

(I'll also try to fix my PR so it actually passes on TravisCI)

Habikki:
I still use distroless for dotnet and seems to be working well for my dotnet core 2.1.400 apps.  I have yet to migrate to dotnet core 3.0 but can push that up if needed.

Let me know how I can help as I'm still dependent on distroless for a few projects I'm involved with.  Even if I need to be actively involved to support dotnet.  Complexity with my team and my AWS bill will thank you for continuing this support.

salrashid123:
ther'es also the dotnet (and other) language versioning scheme these  images shoudl use.  her'e was one proposal and pr  
https://github.com/GoogleContainerTools/distroless/pull/349#issuecomment-511611960

but frankly, i didn't see too much interest internally from the stakeholders and didn't invest in it further.  if we now want to adopt osvariants+runtimeversions, then i'd see these the nubmer distroless images profilerate....

donmccasland:
Anyone understand what's going on in the broken Travis test?

chanseokoh:
Fixed broken builds. Good to go now. Thanks @evanj!

PR #430, opened 2019-10-30
Add Java Debian 10 and fix wrong and missing repositories

chanseokoh:
@evanj see if this makes sense, since you added Debian 10 images.

What this does:
- Fixes inconsistent repositories: e.g., `base:debug-debian10` --> `base-debian10:debug`
- Added missing repositories: e.g., `base-debian9`, `static-debian9`, and a lot of `cc` repos
- Expands Java repositories to include Debian 10 + Java 11.

Of course, I should update `cloudbuild.yaml` too. Will try that later.


java/BUILD
```diff
@@ -53,18 +53,8 @@ cacerts_java(
...
-        "/usr/lib/jvm/java-8-openjdk-amd64/jre/bin/java",
-    ),
-    (
-        "_debian9",
-        "java8_debug",
-        "openjdk-8-jre-headless",
-        "/usr/lib/jvm/java-8-openjdk-amd64/jre/bin/java",
-    ),
+    ("_debian9", "java8", "openjdk-8-jre-headless", "/usr/lib/jvm/java-8-openjdk-amd64/jre/bin/java"),
+    ("_debian9", "java8_debug", "openjdk-8-jre-headless", "/usr/lib/jvm/java-8-openjdk-amd64/jre/bin/java"),
```
evanj: Nice I forgot about the trailing comma stuff for buildifier; much nicer!

chanseokoh:
Yeah, I think our backend uses `cloudbuild.yaml` to for final publication, but this rule is handy for local development because it is so much faster (since you've already built everything through Bazel) and does not require `gcloud`. But then, yeah, it's bad to have duplicates. I'm fine if we remove this, but perhaps later.

@donmccasland I think we can move forward with this.

PR #432, opened 2019-11-02
python3-requirements: Example that uses third-party dependencies

evanj:
This does not build with Bazel, so it is not currently covered by the
TravisCI test.


evanj:
See the README.md in the change for details: https://github.com/GoogleContainerTools/distroless/blob/a7156f6ff9d15892e726dd2368cb2383d40f9d04/examples/python3-requirements/README.md

donmccasland:
Hrmmms, looks like we're going to need to do _something_ for the travis tests here.

evanj:
Sorry for the delay: I just updated this to the latest master, and fixed the Travis test. The issue was that `test.sh` runs `pylint` on all Python scripts, so this was failing because of the missing `psutils` module, which is a third-party module.

PR #446, opened 2019-12-19
Include JVM instrumentation binaries in the distroless/java debug image.

### LINKED ISSUE #350: Include JVM instrumentation binaries in the distroless/java image
It would be useful to include the JVM instrumentation binaries in the distroless/java image. They are really small (approx. 10 KiB each). You cannot use them from outside of the container to connect to a JVM in the container, so you have to use `docker exec` to run them from within the container.

- `jcmd`
- `jinfo`
- `jmap`
- `jstack`
- `jstat`

sT331h0rs3:
Closes #350.

### DIFF
--- a/java/BUILD
+++ b/java/BUILD
@@ -37,8 +37,7 @@ cacerts_java(
         DISTRO_PACKAGES[distro_suffix]["fontconfig-config"],
         DISTRO_PACKAGES[distro_suffix]["libexpat1"],
         DISTRO_PACKAGES[distro_suffix]["libfontconfig1"],
-        DISTRO_PACKAGES[distro_suffix][jre_deb],
-    ],
+    ] + [DISTRO_PACKAGES[distro_suffix][deb] for deb in java_debs],
     # We expect users to use:
     # cmd = ["/path/to/deploy.jar", "--option1", ...]
     entrypoint = [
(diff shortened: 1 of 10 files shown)


sT331h0rs3:
@googlebot I signed it!

sT331h0rs3:
I do not understand why tests are failing. Could anyone please help with this?
[openjdk-8-jdk-headless](https://packages.debian.org/stretch/openjdk-8-jdk-headless) depends on [openjdk-8-jre-headless](https://packages.debian.org/stretch/openjdk-8-jre-headless), and java executable somes from the last one, so there should not be any difference for tests. The same applies for openjdk-11.
However java executable for pair debian9+openjdk-8 differs [in Debian package's file list page](https://packages.debian.org/stretch/amd64/openjdk-8-jre-headless/filelist) and [in tests](https://github.com/GoogleContainerTools/distroless/blob/22bd467/java/testdata/java8.yaml#L4): `/usr/lib/jvm/java-8-openjdk-amd64/jre/bin/java` vs `/usr/lib/jvm/java-8-openjdk-amd64/bin/java`; still tests are passing in this case.

chanseokoh:
It's because now you are unpacking only `openjdk-8-jdk-headless`. You need to unpack both `openjdk-8-jdk-headless` and `openjdk-8-jre-headless`.

>  However java executable for pair debian9+openjdk-8 differs [in Debian package's file list page](https://packages.debian.org/stretch/amd64/openjdk-8-jre-headless/filelist) and [in tests](https://github.com/GoogleContainerTools/distroless/blob/22bd467/java/testdata/java8.yaml#L4) ... still tests are passing in this case.

If you look carefully, the package file list does contain `/usr/lib/jvm/java-8-openjdk-amd64/jre/bin/java`.

sT331h0rs3:
@chanseokoh Thanks a lot for your help!
I think I should also add some tests to cover new cases.

java/BUILD
```diff
@@ -37,28 +37,66 @@ cacerts_java(
         DISTRO_PACKAGES[distro_suffix]["fontconfig-config"],
         DISTRO_PACKAGES[distro_suffix]["libexpat1"],
         DISTRO_PACKAGES[distro_suffix]["libfontconfig1"],
-        DISTRO_PACKAGES[distro_suffix][jre_deb],
-    ],
+    ] + [DISTRO_PACKAGES[distro_suffix][deb] for deb in j_debs],
```
chanseokoh: Could you rename `j_debs` to `java_debs`?

java/jetty/testdata/java11_debug.yaml
```diff
@@ -12,6 +12,10 @@ commandTests:
     command: "/usr/bin/java"
     args: ["-version"]
     expectedError: ['openjdk version "11\.*']
+  - name: javac
+    command: "/usr/lib/jvm/java-11-openjdk-amd64/bin/javac"
```
chanseokoh: Could you also add tests for non-debug images (`java8/11.yaml`) to test the `javac` does not exist? I think this can be done using `fileExistenceTests` with `shouldExist: false` like below.

PR #462, opened 2020-02-13
Rust Example

blhagadorn:
This adds an example of a simple Rust Hello World on STDOUT using the `cc` image


blhagadorn:
I felt compelled to do this because I wanted to show the usage of the `cc` image, FYI

blhagadorn:
@donmccasland

blhagadorn:
> We should probably get someone with some familiarity with rust to review this. But I might as well ask while I'm here:
> 
> 1. Are the libraries required for cc enough for rust?
That's a great question, other than needing libc to do OS, I/O calls, 
> 2. I think you need to include a BUILD file like: https://github.com/GoogleContainerTools/distroless/blob/master/examples/go/BUILD as these examples are built from source.



> We should probably get someone with some familiarity with rust to review this. But I might as well ask while I'm here:
> 
> 1. Are the libraries required for cc enough for rust?
That's a great question, I think libc is all it needs for its basic functions for interfacing with the system library, I/O, networking, etc. rather than having an abstract layer like Golang where syscalls are build into the language.
> 2. I think you need to include a BUILD file like: https://github.com/GoogleContainerTools/distroless/blob/master/examples/go/BUILD as these examples are built from source.
Ahh good point, I will work on a BUILD file for Bazel

examples/rust/BUILD
```diff
@@ -0,0 +1,13 @@
+# Public notice: this file is for internal documentation, testing, and
+# reference only. Note that repo maintainers can freely change any part of the
+# repository code at any time.
+package(default_visibility = ["//visibility:public"])
+
+# Rust
+load("@io_bazel_rules_docker//rust:image.bzl", "rust_image")
```
briandealwis: You need to reference [these from the WORKSPACE](https://github.com/bazelbuild/rules_rust/blob/master/README.md#setup)

README.md
```diff
@@ -88,6 +88,7 @@ You can find other examples here:
 * [Golang](examples/go/Dockerfile)
 * [Node.js](examples/nodejs/Dockerfile)
 * [dotnet](examples/dotnet/Dockerfile)
+* [rust](examples/rust/Dockerfile)
```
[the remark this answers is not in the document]
blhagadorn: Whoops, not it's not I'll change that!

blhagadorn:
Alright thanks for the tip @briandealwis , new to Bazel but the documentation helped and I think I have things correct. Ready for review when any of you can!

examples/rust/BUILD
```diff
@@ -0,0 +1,13 @@
...
+
+rust_image(
+    name = "rust_example",
+    srcs = ["src/main.rs"],
+    base = "//cc:cc",
+)
```
briandealwis: So this compiles for the current platform, which may be different from the container image's platform (e.g., compiled for macOS but deploying to Linux) which results in an odd error when launching the image:
```console
$ bazel run //examples/rust:rust_example
[...]
INFO: Build completed successfully, 1 total action
Loaded image ID: sha256:486acab411b4a2362f606e5167028cad10fcdcb02a549e18bc9068168b7061cd
Tagging 486acab411b4a2362f606e5167028cad10fcdcb02a549e18bc9068168b7061cd as bazel/examples/rust:rust_example
standard_init_linux.go:211: exec user process caused "exec format error"
```

Unfortunately the rust rules don't seem to support cross-compilation yet (bazelbuild/rules_rust#276, and bazelbuild/rules_docker#372), but there isn't a clean way to cause the build to fail (bazelbuild/bazel#3780).  So could you just add a note here that we require building on the same platform as the image.

blhagadorn: Argh that's annoying, I've added the note there as requested. Thanks for looking into that and explaining! I can imagine with WASM that will definitely be a pain point.

blhagadorn:
> Are the libraries required for cc enough for rust?
To re-answer this question after reserach, for this example I think `libc` is enough for most standard Rust procedures. If there are other dependencies, however, they might need other items.

briandealwis:
Thanks for your contribution!

blhagadorn:
No problem, thanks for reviewing!

CodesInChaos:
@blhagadorn is there a reason this needs `cc` and not just `base`?

blhagadorn:
@CodesInChaos For libcc1 which isn't inside of base, and is sparsely found in different Unix operating systems.  See this interesting discussion: https://github.com/rust-lang/rustup/issues/2213 or this masterpiece: https://wiki.alopex.li/LetsBeRealAboutDependencies. Otherwise, users might see this error: `error while loading shared libraries: libgcc_s.so.1: cannot open shared object file: No such file or directory`

However I might be overcomplicating things, is there an alternative?

CodesInChaos:
I didn't realize that rust applications link to libgcc at runtime. I thought that only the compiler depends on on it, while applications depend only on glibc.

blhagadorn:
Yeah, I think it's heavily dependent on operation system, too (Linux is of course a spectrum).  I might be wrong about this though.

PR #466, opened 2020-02-18
Clean up cloudbuild.yaml

chanseokoh:
Much more efficient and better readability.


chanseokoh:
Tested this with `gcloud builds submit --substitutions=COMMIT_SHA=foobarbaz`. GCB fails if any image specified in the last `images:` is not generated, so it proves that this PR still generates all the images we've been doing. Therefore, the only thing to check is that this change doesn't tag images incorrectly.

PR #469, opened 2020-02-19
Reduce visibility of package_manager

chanseokoh:
Turns out this is unnecessary.

I have confirmed building and publishing all images by
1. First building and uploading this PAR to a temporary bucket.
2. Updating WORKSPACE to use the new PAR, which required updating the SHA of the binary.
3. Build and publishing Distroless images on GCB (to my GCP project).

### DIFF
--- a/package_manager/BUILD
+++ b/package_manager/BUILD
@@ -4,7 +4,6 @@ par_binary(
     name = "dpkg_parser",
     srcs = glob(["**/*.py"]),
     main = "dpkg_parser.py",
-    visibility = ["//visibility:public"],
     deps = [
         ":parse_metadata",
         ":util",
--- a/package_manager/cloudbuild.yaml
+++ b/package_manager/cloudbuild.yaml
@@ -23,7 +23,7 @@ steps:
   args: [
     'cp',
     'bazel-bin/package_manager/dpkg_parser.par',
-    'gs://distroless/package_manager_tools/$COMMIT_SHA/dpkg_parser.par'
+    'gs://tmp-chanseok/package_manager_tools/$COMMIT_SHA/dpkg_parser.par'
   ]
 
 # We produce no Docker images.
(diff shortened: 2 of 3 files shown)


ianoc-stripe:
@chanseokoh Have you been making other changes to the GCP blob storage used here by chance?  a bunch of our CI infra got taken out with 403's a an older version of this par. and sha 256 failures against master

ianoc-stripe:
We use these utils to add back in packages that are missing in the distroless, like `liblcms2-2` which causes some image handling in the jdk to blowup

chanseokoh:
@ianoc-stripe maybe try https://storage.cloud.google.com/distroless/package_manager_tools/01579c3dd0f9b4397fe4f50769a9caea130b15e7/dpkg_parser.par Or just sync your distroless fork with the master.

But note that these binaries can be removed in the future.

chanseokoh:
> Or just sync your distroless fork with the master.

Ah, yeah, this won't work for now, but it will start working again after merging #469 and fixing an intermediate inconsistency.

ianoc-stripe:
For now i replaced the remote ref with the http_file and updated the sha to the right one, so its unblocked the ci flows.

wrt that the binaries can be removed, is this code we should fork/not depend on?

chanseokoh:
> wrt that the binaries can be removed, is this code we should fork/not depend on?

In principle, any fork of the repo should build successfully. What happened is that, when I prepared #469, I replaced the current dpkg_parser.par binary to have the Travis builds to pass. This changd the SHA, but now I find that was a mistake. Once #469 is merged, a new dpkg_parser.par will be built. After that, I will fix master to use the new binary.

ianoc-stripe:
Ah thats perfect, @chanseokoh . Thank you. I was just a little concerned if we were depending on something we shouldn't have been here. Appreciate the great work that goes into this!

PR #470, opened 2020-02-19
Upgrade Java to 8u242 and 11.0.6 (from 8u232 and 11.0.5)

### LINKED ISSUE #467: distroless:java-debian10 not using latest OpenJDK 11 release
Hello

I've just pulled the newly-released `distroless:java-debian10` image (thanks @chanseokoh for making this happen!) and noticed that it's running `openjdk 11.0.5 2019-10-15`.

The [latest OpenJDK 11 release](https://packages.debian.org/buster/openjdk-11-jre-headless) for Buster appears to be `11.0.6+10-1~deb10u1`.

I assumed this was specified somewhere in the build, but can't find any reference to it. Unfortunately that means I'm not sure how to help fix this.

Please could we update this to use the latest OpenJDK release? I'm happy to help make this change if someone can direct me to the right place in the repo.

Thanks

chanseokoh:
Fixes #467.

~~Travis build will fail to due my mistake in https://github.com/GoogleContainerTools/distroless/pull/469#issuecomment-588468371. #469 should be merged first.~~ UPDATE: resolved

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -27,8 +27,8 @@ dpkg_src(
     name = "debian_stretch",
     arch = "amd64",
     distro = "stretch",
-    sha256 = "2b13362808b7bd90d24db2e0804c799288694ae44bd7e3d123becc191451fc67",
-    snapshot = "20191230T150135Z",
+    sha256 = "56537cedf58e6f08bb3eafef514a20016fbfd227850ab810c43e5ffb00f57427",
+    snapshot = "20200219T150721Z",
     url = "https://snapshot.debian.org/archive",
 )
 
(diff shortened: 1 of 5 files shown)


WORKSPACE
```diff
@@ -362,7 +362,7 @@ dpkg_list(
 
 dpkg_src(
     name = "debian10_security",
-    package_prefix = "https://snapshot.debian.org/archive/debian-security/20191230T212302Z/",
-    packages_gz_url = "https://snapshot.debian.org/archive/debian-security/20191230T212302Z/dists/buster/updates/main/binary-amd64/Packages.gz",
-    sha256 = "20299ed97c71f26e6c34d948a358e3ac9bb9f4709a90b6f38e6094b688f5270b",
+    package_prefix = "https://snapshot.debian.org/archive/debian-security/20200219T150721Z/",
```
loosebazooka: somehow https://snapshot.debian.org/archive/debian-security/20200219T150721Z/ redirects to https://snapshot.debian.org/archive/debian-security/20200219T081806Z/

WORKSPACE
```diff
@@ -27,25 +27,25 @@ dpkg_src(
     name = "debian_stretch",
     arch = "amd64",
     distro = "stretch",
-    sha256 = "2b13362808b7bd90d24db2e0804c799288694ae44bd7e3d123becc191451fc67",
-    snapshot = "20191230T150135Z",
+    sha256 = "56537cedf58e6f08bb3eafef514a20016fbfd227850ab810c43e5ffb00f57427",
+    snapshot = "20200219T150721Z",
```
loosebazooka: If this value is always the same everywhere can we just put it in a variable at the top? Or is that not bazel-y?

WORKSPACE
chanseokoh: It doesn't have to be same. As you noted, it will redirect to the most recent timestamp if it doesn't exist, so perhaps it's ideal to set exact timestamps instead of having all same. I remember it used to the case at some point. However, I found it cumbersome, so now I'm just using the same timestamp everywhere. As long as it works, I think this is convenient. For this reason, I do not really want to put this in a variable.

WORKSPACE
chanseokoh: See the other comment.

WORKSPACE
loosebazooka: okay that makes sense, but I don't understand where the 20200219T150721Z came from then? Is it a point in time that we chose?

WORKSPACE
chanseokoh: I just go to (for example) https://snapshot.debian.org/archive/debian-security/ and select the most recent timestamp.

PR #483, opened 2020-03-10
Add debian9 and debian10 "latest" tagging for base, cc, and static images

donmccasland:


chanseokoh:
This won't work, unless we remove the internal tag syncing system.

Another problem is that `base:latest` and `base-debian9:latest` will not match. You need to add `base:latest` too.

And can we merge #466 first and base this on top of it? We need to remove the internal sync configuration anyways, so we have enough time.

PR #504, opened 2020-04-25
Simple cc example with Bazel test

erickduran:
Adding Dockerfile and Bazel build


erickduran:
@googlebot I signed it!

examples/cc/BUILD
```diff
@@ -0,0 +1,19 @@
+package(default_visibility = ["//visibility:public"])
```
chanseokoh, line 1:
```suggestion
# Public notice: this file is for internal documentation, testing, and
# reference only. Note that repo maintainers can freely change any part of the
# repository code at any time.
package(default_visibility = ["//visibility:public"])
```

examples/cc/BUILD
```diff
@@ -0,0 +1,19 @@
...
+
+load("@io_bazel_rules_docker//container:container.bzl", "container_image")
+
+container_image(
+    name = "hello",
+    base = "//cc:cc_debian9",
```
chanseokoh, line 7:
```suggestion
    base = "//cc:cc",
```

Issue #321, opened 2019-03-13
distroless/java:11 missing libfreetype.so.6


bentsolheim:
**Description of the issue:** The docker images build ok with Jib, but I get an exception thrown when the container runs. The application is a spring boot 2.1.x app that runs ok in the IDE against java 11.

**Expected behavior:** No exception

**Steps to reproduce:** Jib a spring boot 2.1.x app that uses the autoSizeColumn feature of apache-poi and attempt to run the container on your local docker.

Environment: Linux / Java 11 / Gradle / Spring boot 2.1.x

**jib-gradle-plugin Configuration:**

```
jib {
    from {
        image = 'gcr.io/distroless/java:11'
    }
    to {
        image = "theimage"
    }
    container {
        jvmFlags = ['-Djava.awt.headless=true']
    }
}
```

**Log output:**

```
Caused by: java.lang.UnsatisfiedLinkError: /usr/lib/jvm/java-11-openjdk-amd64/lib/libfontmanager.so: libfreetype.so.6: cannot open shared object file: No such file or directory
	at java.base/java.lang.ClassLoader$NativeLibrary.load0(Native Method) ~[na:na]
	at java.base/java.lang.ClassLoader$NativeLibrary.load(ClassLoader.java:2430) ~[na:na]
	at java.base/java.lang.ClassLoader$NativeLibrary.loadLibrary(ClassLoader.java:2487) ~[na:na]
	at java.base/java.lang.ClassLoader.loadLibrary0(ClassLoader.java:2684) ~[na:na]
	at java.base/java.lang.ClassLoader.loadLibrary(ClassLoader.java:2638) ~[na:na]
	at java.base/java.lang.Runtime.loadLibrary0(Runtime.java:829) ~[na:na]
	at java.base/java.lang.System.loadLibrary(System.java:1867) ~[na:na]
	at java.desktop/sun.font.FontManagerNativeLibrary$1.run(FontManagerNativeLibrary.java:57) ~[na:na]
	at java.base/java.security.AccessController.doPrivileged(Native Method) ~[na:na]
	at java.desktop/sun.font.FontManagerNativeLibrary.<clinit>(FontManagerNativeLibrary.java:32) ~[na:na]
	at java.desktop/sun.font.SunFontManager$1.run(SunFontManager.java:270) ~[na:na]
	at java.base/java.security.AccessController.doPrivileged(Native Method) ~[na:na]
[...]

PR #505, opened 2020-04-27
Adds libuuid1 to the Java distroless images.

digitalpardoe:
I switched over some builds for a project I'm involved with to use the Debian 10-based Java image and noticed that this issue: https://github.com/GoogleContainerTools/distroless/issues/321 seems to have reappeared.

After running with similar debugging flags to the original issue I managed to track it down to a missing `libuuid1` dependency under Debian 10.

This change will also add the dependency to the Debian 9-based image where it is not required but also has no effect.

I've tested this locally and with the project in question and it seems to have fixed the issue.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -84,6 +84,7 @@ dpkg_list(
         "fonts-dejavu-core",
         "fontconfig-config",
         "libfontconfig1",
+        "libuuid1",
         "openjdk-8-jre-headless",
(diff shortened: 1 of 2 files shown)


chanseokoh:
Merging now. You should see the images published on `gcr.io`, probably within an hour.

digitalpardoe:
Awesome, thanks for the quick resolution!

PR #513, opened 2020-05-21
fix: switch to 64-bit busybox instead of 32-bit uclibc busybox

donbowman:
Resolves https://github.com/GoogleContainerTools/distroless/issues/225

This moves to a 64-bit busybox, keeping the version the same.

Conceptually the same as https://github.com/GoogleContainerTools/distroless/pull/380 but a much simpler implementation.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -190,8 +190,8 @@ http_archive(
 http_file(
     name = "busybox",
     executable = True,
-    sha256 = "d922ebe9067d8ed1b2b9cf326776de40a9b23d4518d674e28a3c14181549d28b",
-    urls = ["https://busybox.net/downloads/binaries/1.31.0-i686-uclibc/busybox"],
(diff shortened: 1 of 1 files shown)


chanseokoh:
Thanks for doing this!

The PR title says glibc, but it looks like you are pulling in a musl version.

donbowman:
> Thanks for doing this!
> 
> The PR title says glibc, but it looks like you are pulling in a musl version.

its actually multiarch, i think they just named it badly. its static linked. i'll change the headline

chanseokoh:
> its actually multiarch, i think they just named it badly. its static linked. i'll change the headline

I think it means it's statically linked against musl instead of glibc. But I guess it doesn't matter if we use musl or glibc. I just wanted to confirm with you if this was intended. I'll approve and merge by EOD.

Thank you very much for your contribution!

PR #518, opened 2020-05-29
Add .dockerignore

asafyish:
I added .git/ and target/ folders to the .dockerignore file, it's also possible to add **/.git and **/target for recursive ignore, but it feels like it's too much intervention for the user.


asafyish:
@googlebot I signed it!

PR #526, opened 2020-06-12
WORKSPACE: Update to latest Debian snapshots

evanj:
This includes the Debian 10.4 point release. The following Debian 10
packages were updated:

    base-files       10.3+deb10u4
    libgnutls30       3.6.7-4+deb10u4
    libldap-2.4-2     2.4.47+dfsg-3+deb10u2
    libldap-common    2.4.47+dfsg-3+deb10u2
    tzdata            2020a-0+deb10u1

PR #527, opened 2020-06-12
WORKSPACE: Add buster-updates to debian10 to update ca-certificates

evanj:
This updates ca-certificates to 20200601~deb10u1, which was added to
buster-updates. This may cause problems when packages are added to
buster-updates, but then updated again into the stable point releases.
For example, tzdata version 2019c-0+deb10u1 was added to
buster-updates, then later tzdata version 2020a-0+deb10u1 was added
to buster for the 10.4 "point release". Since dpkg_list takes the first
package it finds, it finds the older tzdata version.

We should eventually fix the dpkg_list function to actually attempt to
order by version numbers to resolve this. For now, I've manually pinned
the tzdata version to ensure it gets the latest version.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -268,6 +268,22 @@ dpkg_src(
     url = "https://snapshot.debian.org/archive",
 )
 
+dpkg_src(
+    name = "debian10_updates",
+    arch = "amd64",
+    distro = "buster-updates",
+    sha256 = "24b35fcd184d71f83c3f553a72e6636954552331adfbbc694f0f70bd33e1a2b4",
+    snapshot = "20200612T083553Z",
+    url = "https://snapshot.debian.org/archive",
+)
+
+dpkg_src(
+    name = "debian10_security",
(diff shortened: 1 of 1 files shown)


evanj:
Hmm... I see that TravisCI is not running on these pull requests. It does appear to be running on Master though.

chanseokoh:
Hmm... I'm not sure what Travis is not firing. Happening on other PRs too. I suspect a temporary outage.

chanseokoh:
I suspect it's a temporary outage that Travis results are not reported. Happening on all PRs. In any case, I think #526 upgraded Java.

```
env var JAVA_VERSION value 8u252 does not match expected value: 8u242
```

Could you fix the test?
```
$ git grep 8u242
java/testdata/java8.yaml:      value: '8u242'
java/testdata/java8_debug.yaml:      value: '8u242'
```

chanseokoh:
Could you merge with master to run fixed tests on this PR?

PR #529, opened 2020-06-12
Version Bazel with Bazelisk; Add .bazelversion=2.2.0

evanj:
This should make it easier to update to newer Bazel releases: just
update the version in .bazelversion!

* Add .bazelversion with "2.2.0" since that is what is used on CI.
* Update .travis.yml and test.sh to download and use Bazelisk.

I can understand if this is not a "desirable" change so feel free to reject it, but I'm a fan of bazelisk, having started to use it for a few Bazel projects.


test.sh
```diff
@@ -22,7 +22,7 @@ set -o xtrace
...
 
 # Bazel build and test
-bazel clean --host_force_python=PY2 --curses=no
-bazel build --host_force_python=PY2 --curses=no //package_manager:dpkg_parser.par
-bazel build --host_force_python=PY2 --curses=no //...
-bazel test --host_force_python=PY2 --curses=no --test_output=errors //...
+bazelisk clean --host_force_python=PY2 --curses=no
+bazelisk build --host_force_python=PY2 --curses=no //package_manager:dpkg_parser.par
+bazelisk build --host_force_python=PY2 --curses=no //...
+bazelisk test --host_force_python=PY2 --curses=no --test_output=errors //...
```
chanseokoh, line 28: Ah, I like This. However, the [Bazelisk](https://github.com/bazelbuild/bazelisk) doc says renaming `bazelisk` to `bazel` and having it on `$PATH` is a viable option. Maybe we can leave this script to use `bazel`? That way, the user using their own Bazel can still work? I'm only suggesting this, because otherwise, I want you to update `CONTRIBUTING.md` to add that you need to download `bazelisk`. I've never heard of `bazelisk` before, so new comers may not know what it is and how to get it.

evanj: Great suggestion, since this will co-exist with existing workflows. I also think the `bazel` shell script has some `.bazelversion` detection, so at least in that case it will say something like "please use bazel version 2.2.0". Will fix.

PR #531, opened 2020-06-12
Update to Bazel 3.2.0; Update all Bazel dependencies

evanj:
This updates basically all the tools used to develop distroless to
the latest versions. This doesn't seem to change anything, but some
of these things were quite old. Some notable changes:

* We don't seem to need --host_force_python=PY2 unless the build
  environment has a broken version of Python3. Using the default
  (which I think is Python3 in recent versions of Bazel) seems to
  work. We should be using that now, since Python2 is now deprecated.

* Run CI on Ubuntu 18.04 instead of 14.04. This is a fairly recent
  LTS release.


.travis.yml
```diff
@@ -1,5 +1,5 @@
 # trusty beta image has jdk8, gcc4.8.4
```
chanseokoh, line 1:
```suggestion
```

evanj: I'll merge this change in a moment; I am still fixing some minor errors in this.

evanj:
Sorry for the noise on this, this is now ready to go. It took me a few passes through Travis to get all the things fixed (partly because I was hoping to avoid the slow bazel test //... on my machine ... which ironically just slowed me down MORE)

evanj:
> Yeah, Python tests (2 and 3) always take the most time. But I haven't bothered to figure out why.

Last I looked: they run a TON of commands in the container, and some combination of Bazel/Docker/the container structure test tool takes FOREVER to start/tear down the container each time. This would be substantially faster if we rewrote it so it started the container ONCE, then run a bunch of separate commands. I made this worse because as I found issues with Debian10, I added MORE command tests to those targets ... oops.

PR #532, opened 2020-06-12
Use supported Bazel 3.2.0 image in cloudbuild.yaml

chanseokoh:
@evanj gcr.io/cloud-builders/bazel cached on GCB wasn't even 2.2.0, so it was already broken after #529.

And it seems like they want to deprecate `gcr.io/cloud-builders/bazel`.
```
$ docker run --rm -it gcr.io/cloud-builders/bazel

                   ***** NOTICE *****

A supported `bazel` image, including multiple tagged versions,
is available at
http://gcr.io/cloud-marketplace-containers/google/bazel.

                ***** END OF NOTICE *****
```

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -2,7 +2,7 @@ timeout: 1800s
 steps:
 
 # Publishing base, cc, and static as latest is handled in a separate release process
-- name: gcr.io/cloud-builders/bazel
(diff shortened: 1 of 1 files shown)


evanj:
Oh man thanks for tracking this down and sorry for causing this!

PR #535, opened 2020-06-14
Fix slow Python tests

### LINKED ISSUE #534: Python tests taking too much time, often causing TIMEOUT on Travis builds
> Yeah, Python tests (2 and 3) always take the most time. But I haven't bothered to figure out why.

Last I looked: they run a TON of commands in the container, and some combination of Bazel/Docker/the container structure test tool takes FOREVER to start/tear down the container each time. This would be substantially faster if we rewrote it so it started the container ONCE, then run a bunch of separate commands. I made this worse because as I found issues with Debian10, I added MORE command tests to those targets ... oops.

_Originally posted by @evanj in https://github.com/GoogleContainerTools/distroless/pull/531#issuecomment-643453887_

chanseokoh:
Fixes #534.

It is now super fast.


chanseokoh:
Closing in favor of #536.

PR #536, opened 2020-06-15
python: Make tests faster: Import modules in a single container

evanj:
The previous version of the test imported a number of standard
library modules in seperate tests. This imports all the modules
in a single container. This is much faster.

Before:

    //experimental/python3:python3_debian10_test  PASSED in 306.6s

After:

    //experimental/python3:python3_debian10_test  PASSED in 14.8s

I also checked that this seems to find all the package errors.
I removed a few packages from BUILD and the expected test failed.
E.g. removed libbz2 makes "import bz2" fail.

This fixes:
https://github.com/GoogleContainerTools/distroless/issues/534

It is an alternative to:
https://github.com/GoogleContainerTools/distroless/pull/535


experimental/python2.7/testdata/python27.yaml
```diff
@@ -23,588 +23,48 @@ commandTests:
...
+        def onerror(failed_module_name):
+          # ignore any errors from modules we are explicitly skipping (e.g. tkinter)
+          if failed_module_name in skip_modules:
+            return
+          raise Exception('failed to import module: {}'.format(repr(failed_module_name)))
+
+        for module_info in pkgutil.walk_packages(onerror=onerror):
+          # TODO: use module_info.name when Debian9 support is dropped
+          module_name = module_info[1]
+          # idlelib on Python3.5 depends on tkinter which is not installed by default
```
chanseokoh, line 59: There's no mention of `idlelib` in the following condition.

experimental/python2.7/testdata/python27.yaml
```diff
@@ -23,588 +23,48 @@ commandTests:
...
+          if module_name in skip_modules or module_name.endswith('.__main__'):
```
chanseokoh, line 60: If we are ignoring an import error in `def onerror` for `skip_modules`, why do we exclude `skip_modules` here? (I just would like to remove any unnecessary code.)

experimental/python2.7/testdata/python27.yaml
```diff
@@ -23,588 +23,48 @@ commandTests:
...
+            continue
+
+          try:
+            __import__(module_name)
+          except ImportError as e:
+            # ignore tkinter import errors: there are so many
+            if 'No module named _tkinter' in str(e):
```
chanseokoh, line 67: Can this not be in `skip_modules` or handled in `onerror()`?

experimental/python3/testdata/python3.yaml
```diff
@@ -29,576 +29,51 @@ commandTests:
...
+          # ignore any errors from modules we are explicitly skipping (e.g. tkinter)
+          if failed_module_name in skip_modules:
+            return
+          raise Exception('failed to import module: {}'.format(repr(failed_module_name)))
+
+        for module_info in pkgutil.walk_packages(onerror=onerror):
+          # TODO: use module_info.name when Debian9 support is dropped
+          module_name = module_info[1]
+          # idlelib on Python3.5 depends on tkinter which is not installed by default
+          if module_name in skip_modules or module_name.endswith('.__main__') or module_name.startswith('idlelib.'):
```
chanseokoh, line 75: ditto. Check `skip_modules` only in either `onerror()` or here?

experimental/python2.7/testdata/python27.yaml
evanj: That was a stray comment from a previous approach. Removed (this was replace with the tkinter catch below). I edited the tkinter comment.

experimental/python2.7/testdata/python27.yaml
evanj: Great suggestion! This is actually the "ignore modules" thing we need to keep. The other one, in the error handler, can go away. I think the difference may be that walk_packages needs to import *packages*, which I guess is what Python calls modules that contain other modules? walk_packages does not actually import the modules themselves.

I did remove the check, and I renamed and shuffled things around a bit, thanks to this suggestion.

evanj: Ah, I spoke a bit too soon: On Python3, this error triggers for `tkinter`; on Python2.7, it does not. My intention was to make these tests as similar as possible. However, it turns out getting this to run with 4 different versions of Python is a bit tricky. I've simplified this for Python2, but Python3 still needs a bit of a gross hack.

experimental/python2.7/testdata/python27.yaml
evanj: See my previous comment: walk_packages does not actually import everything. It only imports "packages", so we need the error handler in two places effectively.

experimental/python2.7/testdata/python27.yaml
chanseokoh: Thanks for the explanation. I think we don't have to make these as similar as possible. It seems like Python 2 is really dead by now, so I'd prefer having simpler code for Python 3, if that's possible.

evanj:
Sorry for the delay. I just submitted a patch with some suggested improvements. Thanks as always for the helpful review!

Issue #537, opened 2020-06-21
Suggestion: better version and update management


petermylemans:
When going through the code, I've noticed that package manager is not always selecting the most up to date version with respect to the chosen snapshot. Currently this seems to be solved by pinning the version of specific libraries in the WORKSPACE file, but this looks a bit error prone as an approach.

My proposal:
 * Introduce a versions file per package bundle (deb9 and deb10) to clearly see what versions are being used. For example: https://github.com/petermylemans/distroless/blob/versions/package_bundle_debian10.versions
 * Introduce a better version selection algorithm based on https://github.com/theclimatecorporation/python-dpkg 

To fully complete the solution, I would suggest to make a kind of bot/github action that takes the latest snapshot from debian 9 / 10, regenerate the version files and see if there is a difference. If so, the bot can send a PR for review.

chanseokoh:
> When going through the code, I've noticed that package manager is not always selecting the most up to date version with respect to the chosen snapshot. Currently this seems to be solved by pinning the version of specific libraries in the WORKSPACE file, but this looks a bit error prone as an approach.

Exactly. You got it absolutely right.

I think your proposal is a great idea. @evanj @ LouisStAmouryou may also be interested in this. (PR #538 already filed.)

> Introduce a versions file per package bundle (deb9 and deb10) to clearly see what versions are being used.

I'm afraid this introduces another maintenance burden. I think you are saying these versions are auto-generated, but I'd like to confirm there is no manual update process for the version.

> To fully complete the solution, I would suggest to make a kind of bot/github action that takes the latest snapshot from debian 9 / 10, regenerate the version files and see if there is a difference. If so, the bot can send a PR for review.

This is a great idea. I love it.

chanseokoh:
> When going through the code, I've noticed that package manager is not always selecting the most up to date version with respect to the chosen snapshot. Currently this seems to be solved by pinning the version of specific libraries in the WORKSPACE file, but this looks a bit error prone as an approach.

Exactly. You got it absolutely right.

I think your proposal is a great idea. @evanj @LouisStAmour may also be interested in this. (PR #538 already filed.)

> Introduce a versions file per package bundle (deb9 and deb10) to clearly see what versions are being used.

I'm afraid this introduces another maintenance burden. I think you are saying these versions are auto-generated, but I'd like to confirm there is no manual update process for the version.

[...]

PR #538, opened 2020-06-21
Update version selection algorithm 

petermylemans:
First implementation for #537 

* Adds version files for each bundle of debs
* Added current version selection as described by debian guidelines (using pydpkg code).
* Removed the pinning of packages
* Added stretch_updates repo to fix ca-certificates version (was currently a version from 2016)

Note that adding pydpkg as a library introduced quite some transitive dependencies and I had issues running it on python2. So I ended up putting the relevant parts of their version comparison code (that is plain python) in its own file.


petermylemans:
@googlebot I signed it!

chanseokoh:
Great idea! Thank you very much for your contribution. I think I will not have time to review this until July, so I ask for your patience. Please check back with me in July.

petermylemans:
> Great idea! Thank you very much for your contribution. I think I will not have time to review this until July, so I ask for your patience. Please check back with me in July.

No problem! Looking forward to hear your feedback when you find the time.

package_manager/dpkg_parser.py
```diff
@@ -72,59 +75,46 @@ def main():
...
     package_to_version_map = {}
     package_file_to_metadata = {}
-    for pkg_vals in set(packages.split(",")):
-        pkg_split = pkg_vals.split("=")
-        if len(pkg_split) != 2:
-            pkg_name = pkg_vals
-            pkg_version = ""
-        else:
-            pkg_name, pkg_version = pkg_split
+    for pkg_name in set(packages.split(",")):
```
chanseokoh, line 89: Can you confirm it is no longer the case that items in `packages` are a "pkg_name=version" pair? Why do think the old code was attempting to parse the item?

package_manager/dpkg_parser.py
```diff
@@ -72,59 +75,46 @@ def main():
...
+        pkg = {}
         for package_file in package_files.split(","):
             if package_file not in package_file_to_metadata:
```
chanseokoh, line 92: I'm now wondering if this check as well as `package_file_to_metadata` are necessary. (I don't remember what values `package_files` hold though.) Can't we just always open and read the file, and then just get `metadata[pkg_name]`? It is really possible that `package_files` may have duplicate items and the optimization with this check is worth?

package_manager/dpkg_parser.py
```diff
@@ -45,6 +46,8 @@
                     help='A comma delimited list of packages to search for and download')
 parser.add_argument("--workspace-name", action='store',
                     help='The name of the current bazel workspace')
+parser.add_argument("--versionsfile", action='store',
+                    help='The target path of the versions file')
```
chanseokoh, line 50: "If set, the output path of the versions file to be generated"?

package_manager/dpkg_parser.py
petermylemans: I've removed all of them in the workspace file. 

The only reason I can imagine the code was in there, is to support pinning the versions. The comments above the pinned packages in the WORKSPACE file (see diff) seem to suggest the same. 

But this doesn't make any sense to me given that the most recent version should be resolved when using the snapshots. Pinning is already done through snapshot pinning. So pinning both the snapshot and the package version can only result in errors when the two mismatch.

The only argument to leave the code in, would be to support other project that import these rules through GIT?

chanseokoh: Ah, now I get it where these values come from (`dpkg_list`, that is). I agree we don't need this anymore.

package_manager/dpkg_parser.py
petermylemans: the check is there because of the two nested loops:
 1. first loop over all packages that we want to download
 2. loop over all repository url metadata to see if it has a version. If not available in the map, download it first
      2.1 **Added**: check if metadata found is more recent that other previous package metadata from another package_file

If you want; we can optimize it to fill the package_file_to_metadata before looping over all the packages. In that case the check you pointed out can be removed.

chanseokoh: Ah, I think I get it. This check is worth because (probably) `package_file` is a remote URL for a JSON file hence opening and reading it is expensive? If so, I think this check makes sense.

chanseokoh:
Thank you very much for your contribution! I will merge this next week.

petermylemans:
> This is awesome. Haven't tested the shell script, but other parts look good to me. I have a few comments.

Thx! I also have the shell script running with a github action on my fork to auto create pull requests like e.g. https://github.com/petermylemans/distroless/pull/2 

See also https://github.com/petermylemans/distroless/actions. It failed today due to a 404 on debian snapshots though. But the script is working now on my machine. Maybe let this run for a bit and I can send a second PR with the automation after stabilization?

package_manager/dpkg_parser.py
petermylemans: Fixed

package_manager/dpkg_parser.py
petermylemans: Correct, that's why I left it in the place it was originally written (works like a cache).

chanseokoh:
I think this is good to go. Really happy to have this setup. Thank you very much @petermylemans! Merging now.

PR #543, opened 2020-06-30
fix dpkg_parser in python3

jonnyyu:
by replacing getheader() with get().

getheader() function is in PY2 HTTPMessage but not in PY3 HTTPMessage. It seems to me a bug in six that getheader() function is available in PY2.


jonnyyu:
@googlebot I signed it!

chanseokoh:
Thank you for your contribution @jonnyyu!

PR #544, opened 2020-07-07
Automating deb package updates via GitHub Actions

### LINKED ISSUE #537: Suggestion: better version and update management

chanseokoh:
Does this look good @petermylemans?

Fixes #537.


petermylemans:
Looks good to me!

You might want to consider renaming the file from main.yml to something more specific like e.g. update-snapshots.yml or update-debs.yml.

.github/workflows/update-deb-package-snapshots.yml
```diff
@@ -0,0 +1,26 @@
+name: update-snapshots
+on:
+  schedule:
+    - cron:  '0 0 * * *'
```
briandealwis, line 4: Might be better to schedule this slightly off-peak?  Help Github avoid computation cascades at midnight?

.github/workflows/update-deb-package-snapshots.yml
```diff
@@ -0,0 +1,26 @@
...
+      uses: abhinavsingh/setup-bazel@v3
+      with:
+        version: 3.2.0
+
+    - name: Run update script
+      run: ./updateWorkspaceSnapshots.sh
+    
+    - name: Create Pull Request
+      uses: peter-evans/create-pull-request@v2
+      with:
```
briandealwis, line 22: Do you want to set an author and committer? https://github.com/peter-evans/create-pull-request#committer-and-author

.github/workflows/update-deb-package-snapshots.yml
chanseokoh: Good idea. Going with UTC 8:30am (EST 4:30 and 1:30 PST as of now).

.github/workflows/update-deb-package-snapshots.yml
chanseokoh: Oh, that's interesting. But I think it makes sense to have the author as the bot.

Issue #458, opened 2020-01-23
Build reproducibility: images depend on where they are built


chanseokoh:
UID and GID of some directories are build environment-dependent. See https://github.com/GoogleContainerTools/distroless/issues/456#issuecomment-577864699 and https://github.com/GoogleContainerTools/distroless/issues/456#issuecomment-577859858.

Is it possible that the build runs a tar zip-up on the environment where Bazel runs, and it doesn't clear the UID and GID information?

chanseokoh:
> See [#456 (comment)](https://github.com/GoogleContainerTools/distroless/issues/456#issuecomment-577864699) and [#456 (comment)](https://github.com/GoogleContainerTools/distroless/issues/456#issuecomment-577859858).

So I think these are from `cacerts/extrach.sh`, and looking at https://github.com/GoogleContainerTools/distroless/pull/551#discussion_r457647403, maybe it's an easy fix to just add `--owner=0 --group=0` when tarring up?

petermylemans:
The rabbit hole goes even deeper: 
Python 2 default for TAR format: GNU
Python 3.8 default for TAR format: PAX

==> we should probably fix those at the archive generation level tar.open(format=tarfile.GNU_FORMAT)

That is why building distroless with PY2 or PY3 results in different digests ;) 

Anyway, let's first fix the owner issues first.

PR #551, opened 2020-07-19
Fix java runtime default encoding to default to C.UTF-8

### LINKED ISSUE #253: distroless/java should set default output encoding to UTF-8
From https://github.com/GoogleContainerTools/jib/issues/1227 it seems that distroless/java's default file encoding is picked up as US-ASCII.  The Docker `openjdk` images [had the same problem](
https://github.com/docker-library/openjdk/issues/32) which was solved by setting `LANG=C.UTF-8`.  For some reason that doesn't make a difference with `distroless/java`:
```
$ docker run -e LANG=C.UTF-8 -it --rm issue-jib-utf8:0.0.1
env: {PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin, SSL_CERT_FILE=/etc/ssl/certs/ca-certificates.crt, HOSTNAME=825d06a20ca4, TERM=xterm, LANG=C.UTF-8, HOME=/home}
sun.jnu.encoding: ANSI_X3.4-1968
file.encoding: ANSI_X3.4-1968
Locale.getDefault(): en_US
Charset.defaultCharset: US-ASCII
Hello ???
```

petermylemans:
Finishes the work started by @chanseokoh  by creating a new locale build option that extracts the c locale from the appropriate debian package. This only extracts the `/usr/lib/locale` files and does not include the additional binaries included in the deb package.

I've added this to the java base image support as it impacts all java versions equally. I'm not sure whether it should be somewhere higher in the stack (libc is quite generic), but I guess we can move it further up the chain as issues come in from other software stacks with similar problems?

After merging this, the branch with the same name on upstream distroless can be cleaned up as well.

Bazel is quite new to me, so any feedback regarding syntax and naming conventions is much appreciated. I tried to be consistent with the cacerts approach.

Fixes #253


java/BUILD
```diff
@@ -20,6 +21,14 @@ DISTRO_VERSIONS = {
     "_debian10": versions_debian10,
 }
 
+[locale(
+    name = "locale_java" + distro_suffix,
+    deb = DISTRO_PACKAGES[distro_suffix]["libc-bin"],
+) for distro_suffix in [
+    "_debian9",
+    "_debian10",
+]]
```
chanseokoh, line 30: Probably this will work:
```suggestion
) for distro_suffix in DISTRO_SUFFIXES]
```
Could you also update the code in line 35-38?

locale/extract.sh
```diff
@@ -0,0 +1,18 @@
...
+DEB=$1
+OUT_TAR=$2
+
+ar -x "$DEB" data.tar.xz
+
+tar -xf data.tar.xz ./usr/lib/locale ./usr/share/doc/libc-bin/copyright
+
+rm data.tar.xz
+
+tar -c --owner=0 --group=0 -f locale.tar ./usr
```
chanseokoh, line 14: Ah, I don't see any usage of `--owner=0` or `--group=0` in other files. Now I suspect that's probably the reason for #458.

locale/extract.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+
+set -o errexit
+
+DEB=$1
+OUT_TAR=$2
+
+ar -x "$DEB" data.tar.xz
+
+tar -xf data.tar.xz ./usr/lib/locale ./usr/share/doc/libc-bin/copyright
```
chanseokoh, line 10: To be conservative, should we do
```suggestion
tar -xf data.tar.xz ./usr/lib/locale/C.UTF-8 ./usr/share/doc/libc-bin/copyright
```
?

locale/locale.bzl
```diff
@@ -0,0 +1,33 @@
...
+
+locale = rule(
+    attrs = {
+        "deb": attr.label(
+            allow_single_file = [".deb"],
+            mandatory = True,
+        ),
+        # Implicit dependencies.
+        "_extract": attr.label(
+            default = Label("//locale:extract_locale"),
```
chanseokoh, line 22: I'm really new to Bazel as well, but I guess perhaps it's not really difficult to make this portion general (e.g., by accepting another input attribute to the rule besides `deb` on line 16) and share code between this and `cacerts/cacerts.bzl`. However, let's not attempt to do this now. I'll just file a cleanup issue.

chanseokoh: Filed #552. This will be useful if we can generalize this.

locale/extract.sh
petermylemans: Correct, `dive` also shows this quite clearly. I also encountered the same issue on my own downstream extension that adds some extra certificates in a custom cacerts rule. 

Small gotcha: --owner and --group seem to be the preferred way to clear the user and group when packing the tar. However on Mac OS, you need to make sure that GNU Tar is chosen before MacOS' BSD Tar 🎉. I did that by linking it in in /usr/local/bin/tar to the gnu tar from homebrew. 

Do you want me to include the cacerts in this PR or do you prefer a separate PR for that issue?

locale/locale.bzl
petermylemans: Had the same reflection after writing the code. There should probably be a kind of pkg_tar_from_deb rule with an includes / excludes attribute.

locale/extract.sh
chanseokoh: > Do you want me to include the cacerts in this PR or do you prefer a separate PR for that issue?

I prefer a separate PR (if you decide to make anther contribution). Thanks for considering to do so in any case.

> However on Mac OS, you need to make sure that GNU Tar is chosen before MacOS' BSD Tar  due to differences in the accepted arguments. I did that by linking it in in /usr/local/bin/tar to the gnu tar from homebrew.

Thanks for the info. Could you add this as code comment in `extract.sh`? I'll update this info in #458 too.

chanseokoh:
Once minor feedback about adding a code comment, and we should be good to go with merging. Thanks!

locale/extract.sh
petermylemans: I've added the comment in the shell script.

For the PR: I'll see if I can put something together later this week.

PR #553, opened 2020-07-21
Cleanup and improve reproducibility of builds

### LINKED ISSUE #458: Build reproducibility: images depend on where they are built

### LINKED ISSUE #552: Eliminate code duplication between cacerts/cacerts.bzl and locale/locale.bzl
`locale/locale.bzl` is introduce in #551 and almost same as [`cacerts/cacerts.bzl`](https://github.com/GoogleContainerTools/distroless/blob/5a5744efd2be5d9edb84fabd407e0f883a27dc8b/cacerts/cacerts.bzl#L1). I think it's possible to share code by [making this rule general](https://github.com/GoogleContainerTools/distroless/pull/551#discussion_r457649864):

> ... perhaps it's not really difficult to make this portion general (e.g., by accepting another input attribute to the rule besides `deb` on line 16) and share code between this and `cacerts/cacerts.bzl`. However, let's not attempt to do this now. I'll just file a cleanup issue.

This can be useful, since there are open issues (such as #447 and #508) where we need to exclude unwanted files from a `.deb` when we install them.

petermylemans:
**Package Manager** package

- Replaced glob pattern from package manager par with explicit params to make room for additional utilities
- Added new utility:  *dpkg_extract* extracts a subset of files from a debian archive to the current directory 
* Fixed differences in owner of  `/etc/os-release` when building locally by using the build_tar utility.

**cacerts**
* Updated actions to use the dpkg_extract and build_tar for reproducibility
* Fixed differences in owner when building locally (by switching to build_tar)
* Added targets in the BUILD to build both cacerts and cacerts_java tar balls for the different distros
* Got rid of the extract_sh step and reset the visibility of the package to the workspace only.  @dlorenc  In August 2017 you set the cacerts to public for use in "debian-docker". Is this stil valid?
* Removed cacerts deb output. Isn't used internally in the repo anymore since the upgrade to debian 10 by @evanj . Before it was added for "tracking package status".  But this seems a bit strange to me as we are transforming the upstream package instead of "installing the package". 

**locale**
* Updated actions to use the dpkg_extract and build_tar for reproducibility
* Added targets in the BUILD to build locale tar balls for the different distros

Fixes #552
Fixes #458 with the exception of the python 3.8 issue


package_manager/util_test.py
```diff
@@ -66,5 +66,13 @@ def test_generate_debian_os_release(self):
             util.generate_os_release(distro, output_file)
             self.assertEqual(expected_output, output_file.getvalue())
 
+    def test_build_os_release_tar(self):
+        util.build_os_release_tar("buster", "etc/os-release", "etc", "release.tar")
+        os.remove("etc/os-release")
+        os.rmdir("etc")
```
petermylemans, line 72: This feels a bit strange to me. Anybody with better python skills got an idea on how to work with generated files? I've tried using temp files and folders; but the tar command is a bit picky on folder structures.

cacerts/BUILD
```diff
@@ -1,6 +1,15 @@
-package(default_visibility = ["//visibility:public"])
+package(default_visibility = ["//:__subpackages__"])
 
-sh_binary(
-    name = "extract_certs",
-    srcs = ["extract.sh"],
-)
+load("//:distro.bzl", "DISTRO_PACKAGES", "DISTRO_SUFFIXES")
```
petermylemans, line 3: @chanseokoh not so sure about this approach. I've tried adding the distro.bzl at the root to avoid writing the same basic loop setup for deb9 and deb10 all over the place.

There are other places where the same loop is still present; but I thought it better to validate the idea first before going all.

base/distro.bzl
```diff
@@ -0,0 +1,14 @@
+load("@package_bundle//file:packages.bzl", "packages")
```
evanj, line 1: This is smart, nice way to reduce this duplication.

cacerts/BUILD
evanj: I personally think this seems smart! I hacked the Debian10 support into distroless in a fairly ... unplanned way. I think this seems a bit more structured.

evanj:
Just to confirm I understand: This doesn't actually cause any problems for programs *using* Distroless, this seems to mostly be some simplification/cleanup of the build tools themselves, right?

I think this looks good to me, but I'm not an owner, just a sometime contributor and user :) To be honest I don't follow what exactly is happening with cacerts here. I do know that distroless at one point had somewhat "broken" public CA certificates, since it can't really run the Debian `update-ca-certs` tool which creates the normal symlink tree in `/etc/ssl/certs`, so some of this stuff is there to create the "aggregated" cert file. I think there are at least a few tests in the Java and Python tests to check that it works with google.com, because at some point it did not. Hopefully those tests do the trick :)

package_manager/dpkg_extract.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/sh
+
+set -o errexit
+
+DEB=$1
+shift
```
chanseokoh, line 6: I get `line 6: shift: shift count out of range` when I don't provide any argument to the script. (You may see different behaviors when using a `sh` vs `bash`.)

package_manager/dpkg_extract.sh
```diff
@@ -0,0 +1,21 @@
...
+if [ -z ${DEB+x} ]; then
```
chanseokoh, line 7: From my testing, this doesn't work, as `DEB` is always defined. (That is, even if `$1` is undefined, the script ends up seeing `DEB=`, resulting in defining `DEB` .) I prefer `[ -z "${DEB}" ]` (i.e., just surrounding `DEB` in quotes).

package_manager/util.py
```diff
@@ -1,9 +1,12 @@
 import hashlib
 import base64
 import collections
-import tarfile
 import os
 
+# Distroless's test.sh runs pylint on all python files, but this module will not exist
+# pylint: disable=import-error
+from build_tar import TarFile
```
chanseokoh, line 8: Question as a Python newbie: it's just using a different tar library to create a tar, right?

chanseokoh:
> * Fixed differences in owner of  `/etc/os-release` when building locally by using the build_tar utility.

I don't think I saw any changes about this?

Again, thanks for the awesome contribution!

package_manager/util.py
petermylemans: build_tar comes from  "@rules_pkg//:build_tar" . It is the same utility as used by the container layer building tooling.

Internally they also use tarfile as provided by Python, but they add a reproducibility sauce onto that. (such as stripping  user ids by default, resetting mtimes, sanitizing path names...).

petermylemans:
> I vaguely remember that when building the entire images, I saw output of multiple executions of the cacert extraction script. I just think if it is still the case with this new rule, it may increase the overall build time when we start to expand the usage in multiple places. Can it be that case? For example, if I use some extraction rule in `java/BUILD`, will it unpack a deb and create the same tar for each of the dozens of the Java images?

Given that they are now regular targets, it should work like any other dependency. So I would expect cacerts extraction to run twice (once for deb 9 and another time for deb 10). Same for cacerts_java, resulting in 4 targets in total.

petermylemans:
> Just to confirm I understand: This doesn't actually cause any problems for programs _using_ Distroless, this seems to mostly be some simplification/cleanup of the build tools themselves, right?
> 
> I think this looks good to me, but I'm not an owner, just a sometime contributor and user :) To be honest I don't follow what exactly is happening with cacerts here. I do know that distroless at one point had somewhat "broken" public CA certificates, since it can't really run the Debian `update-ca-certs` tool which creates the normal symlink tree in `/etc/ssl/certs`, so some of this stuff is there to create the "aggregated" cert file. I think there are at least a few tests in the Java and Python tests to check that it works with google.com, because at some point it did not. Hopefully those tests do the trick :)

@evanj Correct. My worry relates more to people "highjacking" some of the public packages of this repo.  In theory they should have pinned the repo by commit hash, but they might run into issues when updating to a newer hash due to change of build tooling that should have been internal to distroless (imho).

package_manager/dpkg_extract.sh
petermylemans: I also have a python version of this script in another personal project that includes a basic form of filtering (excluding) files. That script is also using the build_tar instead of the unix tooling. 

It was developed after I opened this PR, so it might need some polish (an hour or so).

Might be better to switch to that one instead of fixing this shell script? Or do you have a preference for shell scripts?

petermylemans:
> > * Fixed differences in owner of  `/etc/os-release` when building locally by using the build_tar utility.
> 
> I don't think I saw any changes about this?
> 
> Again, thanks for the awesome contribution!

@chanseokoh  package_manager/util.py#generate_os_release was changed for this.

package_manager/dpkg_extract.sh
chanseokoh: I think a shell script is fine. Not that a script is a good thing, but I don't think we want over-engineering. I mean, the goal is to not increase complexity and code lines to reduce maintenance burden. When I first came to this repo, everything was overwhelming.

chanseokoh: But the excluding feature will probably be a nice thing that we may need anyway. But let's not do it here to keep the PR small. We can always do it later.

package_manager/dpkg_extract.sh
petermylemans: Fixed.

petermylemans:
@chanseokoh let me know if you're happy with the code. Then I can do a final check to make sure the digests of the unchanged layers match the ones on gcr.

package_manager/dpkg_extract.sh
petermylemans: Should be fixed

chanseokoh:
Looks good to me! Thanks for checking them. Really appreciate it.

petermylemans:
Had to do a small fix for file permissions on /etc/os-release (is back to 0644 as on gcr).

I've got two diffs after local build: 
  1. first layer in base (adds certificates). ./usr/share/doc/ca-certificates was added without defining permissions for parent directories (as is standard in the layer building tooling). So now the ./usr, ./usr/share are defined earlier in the tar entry list. This doesn't change the output, but does change the digest.  --> one time change to fix the order, but remains stable after this.

``` 
< is upstream
> is new
13a14,17
> drwxr-xr-x 0/0               0 1970-01-01 01:00 ./usr/
> drwxr-xr-x 0/0               0 1970-01-01 01:00 ./usr/share/
> drwxr-xr-x 0/0               0 1970-01-01 01:00 ./usr/share/doc/
> drwxr-xr-x 0/0               0 1970-01-01 01:00 ./usr/share/doc/ca-certificates/
40d43
< drwxr-xr-x 0/0               0 1970-01-01 01:00 ./usr/
47d49
< drwxr-xr-x 0/0               0 1970-01-01 01:00 ./usr/share/
72d73
< drwxr-xr-x 0/0               0 1970-01-01 01:00 ./usr/share/doc/
```
2. 4th layer in java:base (adds locale c-bin and some other generic java dependencies). Now the entries are always added in **alphabetic order** (not some random order as before).
```
0a1,2
> drwxr-xr-x 0/0               0 1970-01-01 01:00 ./
> drwxr-xr-x 0/0               0 1970-01-01 01:00 ./etc/
13,14c15,16
< -rw-r--r-- 0/0              62 1970-01-01 01:00 ./usr/lib/locale/C.UTF-8/LC_NAME
< -rw-r--r-- 0/0            2498 1970-01-01 01:00 ./usr/lib/locale/C.UTF-8/LC_TIME
---
> -rw-r--r-- 0/0             131 1970-01-01 01:00 ./usr/lib/locale/C.UTF-8/LC_ADDRESS
> -rw-r--r-- 0/0         1515838 1970-01-01 01:00 ./usr/lib/locale/C.UTF-8/LC_COLLATE
15a18,19
> -rw-r--r-- 0/0             243 1970-01-01 01:00 ./usr/lib/locale/C.UTF-8/LC_IDENTIFICATION
> -rw-r--r-- 0/0              23 1970-01-01 01:00 ./usr/lib/locale/C.UTF-8/LC_MEASUREMENT
18d21
< -rw-r--r-- 0/0              50 1970-01-01 01:00 ./usr/lib/locale/C.UTF-8/LC_NUMERIC
20,24c23,24
< -rw-r--r-- 0/0             243 1970-01-01 01:00 ./usr/lib/locale/C.UTF-8/LC_IDENTIFICATION
< -rw-r--r-- 0/0              47 1970-01-01 01:00 ./usr/lib/locale/C.UTF-8/LC_TELEPHONE
< -rw-r--r-- 0/0             131 1970-01-01 01:00 ./usr/lib/locale/C.UTF-8/LC_ADDRESS
< -rw-r--r-- 0/0              23 1970-01-01 01:00 ./usr/lib/locale/C.UTF-8/LC_MEASUREMENT
< -rw-r--r-- 0/0         1515838 1970-01-01 01:00 ./usr/lib/locale/C.UTF-8/LC_COLLATE
---
> -rw-r--r-- 0/0              62 1970-01-01 01:00 ./usr/lib/locale/C.UTF-8/LC_NAME
> -rw-r--r-- 0/0              50 1970-01-01 01:00 ./usr/lib/locale/C.UTF-8/LC_NUMERIC
26c26,27
< drwxr-xr-x 0/0               0 1970-01-01 01:00 ./
---
> -rw-r--r-- 0/0              47 1970-01-01 01:00 ./usr/lib/locale/C.UTF-8/LC_TELEPHONE
> -rw-r--r-- 0/0            2498 1970-01-01 01:00 ./usr/lib/locale/C.UTF-8/LC_TIME
86d86
< drwxr-xr-x 0/0               0 1970-01-01 01:00 ./etc/
```

petermylemans:
@chanseokoh looks good to me. I just had to fix 2 file permissions from 755 to 644 (see last 2 commits after your review).

Two layers will be changes (one in base and one in java base) as described in the comment above for future reference.

PR #563, opened 2020-08-07
php-fpm distroless image

dlen:
This is a php-fpm debian10 based distroless image. It features:
- php and php-fpm 7.3 interpreters
- Basic PHP extensions ( all the php-common provided extensions plus: mysql, json, jd, xml, readline, opcache).
- a www.conf pool that sets php-fpm to listen on port 9000 and forwards errors to stderr
- It adds the www-data user to the container as it is the default user for php-fpm forks
- The container runs as root

We thought this could be of use for the community that wants to run php on a distroless fashion.


dlen:
@googlebot I signed it!

dlen:
@googlebot I signed it!

chanseokoh:
This repo is primarily maintained and driven by the community (the users) and thus welcomes external contributions. I would be interested in seeing how the community decides the direction and reaches on a specific implementation, or if this will be implemented eventually.

I usually see it takes time before a sufficient number of people gather to amass a sizable momentum to make this kind of work go forward.

I have no expertise in PHP, but looking at the code, one concern is that there seem too many static properties (file paths and version strings) that I think will break as time goes and require constant maintenance.

EraYaN:
Due to the way the debian packages are done, I think you can't really get around those paths. At least the debian maintainers don't change stuff around often.

cawolf:
@chanseokoh can you elaborate which paths and versions are the most critical for you? As far as I can see, everything contained in the `.versions` files is already requiring constant maintenance, as we can see with the conflicting files in this PR. Unfortunately, php-fpm depends on a lot of libraries to work properly, but maybe we can find a way to limit the maintenance overhead.

From the perspective of a PHP developer: the smallest current debian based images are really big and require the same maintenance. I'm looking forward to having such a distroless image for php-fpm.

chanseokoh:
> As far as I can see, everything contained in the .versions files is already requiring constant maintenance, as we can see with the conflicting files in this PR.

Just wanted to correct this, nothing else. The reason you see conflicting `.version` files is exactly because we automated generating them; they require no maintenance. It's just that the files in this PR were generated in Aug, 2020. Re-generating the files will get rid of the conflicts.

q464pp:
不太懂

q464pp (commented):
不太懂PHP拉取合併

pentago:
Any news on this? I'd love to use it too.

q464pp:
可以拉取合併嗎？

pentago:
> 可以拉取合併嗎？

Šta?

gpetrov:
This last oks really good. Could you add also php 8 and Debian 11 support as well?

q464pp:
可以拉取合併嗎？

pentago:
> 可以拉取合併嗎？

Ne kontam

piotrkardasz:
@gpetrov I've created a separate PR to support multiple versions of PHP based on the most recent package repository `packages.sury.org`

pentago:
I think PHP distroless is kot going to happen anytime soon, or at all :(

PR is open for almost 2 years now..

dlen:
Hi! 

What @pentago says is the feeling I got when I published this PR. I could update the PR and solve the conflicts would that be enough to get it merged?

loosebazooka:
We don't really want to add more language runtimes without a better extensibility model so someone else can host and maintain the php image. A php image is out of scope for this repository.

Issue #583, opened 2020-09-26
distroless/static multi-arch


mattmoor:
The distroless/static image is essentially just data: `cacerts`, `tzdata`, `/etc/passwd`, `/tmp`. (I do see a single binary pulled in: `tzconfig`, but I think ideally we'd even strip this)

In `ko`, we (@jonjohnsonjr 👀 ) just landed support for multi-arch, which is based around the set of platforms available in the base image (when you pass `--platform=all`).  I hear that `Jib` is looking into multi-arch as well.

By default, `ko` uses `gcr.io/distroless/static:nonroot`, which is just `linux/amd64`, but because the image itself is basically just data it actually works fine on `arm64` (if forced).  As a workaround for this, today I whipped up the script below to effectively "cast" the `amd64` image to an `arm64` image and produce a manifest list.

In principle, the way most of distroless builds, it should be possible to "properly" produce the image for each architecture from the packages on the appropriate architecture mirror, but I'm a little worried about the lack of multi-arch support in rules_docker being a major blocker.

I'm sort of wondering about shipping an initial band-aid like the trick below 🥺 

```
package main

import (
	"log"
	"github.com/google/go-containerregistry/pkg/authn"
	"github.com/google/go-containerregistry/pkg/name"
	v1 "github.com/google/go-containerregistry/pkg/v1"
	"github.com/google/go-containerregistry/pkg/v1/empty"
	"github.com/google/go-containerregistry/pkg/v1/mutate"
	"github.com/google/go-containerregistry/pkg/v1/remote"
	"github.com/google/go-containerregistry/pkg/v1/types"
)
func main() {
	ref, err := name.ParseReference("gcr.io/distroless/static:nonroot")
	if err != nil {
		log.Fatalf("ParseReference() = %v", err)
	}
	tag, err := name.NewTag("docker.io/mattmoor/distroless-static:nonroot")
	if err != nil {
		log.Fatalf("NewTag() = %v", err)
	}
	ropt := []remote.Option{
		remote.WithAuthFromKeychain(authn.DefaultKeychain),
	}
	desc, err := remote.Get(ref, ropt...)
	if err != nil {
		log.Fatalf("Get() = %v", err)
	}
	switch desc.MediaType {
	case types.OCIImageIndex, types.DockerManifestList:
		log.Fatalf("ref: %v, is an index! %q", ref, desc.MediaType)
	}
	img, err := desc.Image()
	if err != nil {
[...]

PR #586, opened 2020-09-26
Add arm64 variants for distroless static

mattmoor:
Related: https://github.com/GoogleContainerTools/distroless/issues/583

### DIFF
--- a/base/base.bzl
+++ b/base/base.bzl
@@ -3,52 +3,74 @@ load("@io_bazel_rules_docker//container:container.bzl", "container_image")
 load("@io_bazel_rules_docker//contrib:test.bzl", "container_test")
 load(":distro.bzl", "DISTRO_PACKAGES", "DISTRO_REPOSITORY")
 load("//cacerts:cacerts.bzl", "cacerts")
+load("//:checksums.bzl", "ARCHITECTURES")
 
 NONROOT = 65532
 
 # Replicate everything for debian9 and debian10
 def distro_components(distro_suffix):
-    cacerts(
+    for arch in ARCHITECTURES:
+        cacerts(
+            name = "cacerts_" + arch + distro_suffix,
+            deb = DISTRO_PACKAGES[arch][distro_suffix]["ca-certificates"],
+        )
+
+        for (user, uid, workdir) in [("root", 0, "/"), ("nonroot", NONROOT, "/home/nonroot")]:
+            container_image(
+                name = "static_" + user + "_" + arch + distro_suffix,
+                debs = [
+                    DISTRO_PACKAGES[arch][distro_suffix]["base-files"],
+                    DISTRO_PACKAGES[arch][distro_suffix]["netbase"],
+                    DISTRO_PACKAGES[arch][distro_suffix]["tzdata"],
+                ],
+                architecture = arch,
+                env = {
+                    "PATH": "/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin",
+                    # allows openssl to find the certificates by default
+                    # TODO: We should run update-ca-certifaces, but that requires "openssl rehash"
+                    # which would probably need to be run inside the container
+                    "SSL_CERT_FILE": "/etc/ssl/certs/ca-certificates.crt",
+                },
+                tars = [
+                    ":passwd",
+                    ":group_tar",
+
+                    # Create /tmp, too many things assume it exists.
+                    # tmp.tar has a /tmp with the correct permissions 01777
+                    # A tar is needed because at the moment there is no way to create a
+                    # directory with specific permissions.
+                    ":tmp.tar",
+                    ":nsswitch.tar",
+                    DISTRO_REPOSITORY[arch][distro_suffix] + "//file:os_release.tar",
+                    ":cacerts_" + arch + distro_suffix + ".tar",
+                ],
+                user = "%d" % uid,
+                workdir = workdir,
+            )
+
+    for (user, uid, workdir) in [("root", 0, "/"), ("nonroot", NONROOT, "/home/nonroot")]:
+        native.alias(
+            name = "static_" + user + distro_suffix,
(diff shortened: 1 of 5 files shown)


mattmoor:
cc @jonjohnsonjr @chanseokoh @loosebazooka @ImJasonH 

WDYT about starting with a pattern like this, and then starting to work on stitching the images together as we work to plumb through better architecture awareness both above (e.g. `java`, `python`, ...) and below (e.g. actually fetching `arm64` packages)?

base/base.bzl
```diff
@@ -8,47 +8,68 @@ NONROOT = 65532
...
     container_image(
         name = "base" + distro_suffix,
         base = ":static" + distro_suffix,
         debs = [
-            DISTRO_PACKAGES[distro_suffix]["libc6"],
-            DISTRO_PACKAGES[distro_suffix]["libssl1.1"],
-            DISTRO_PACKAGES[distro_suffix]["openssl"],
+            DISTRO_PACKAGES["amd64"][distro_suffix]["libc6"],
+            DISTRO_PACKAGES["amd64"][distro_suffix]["libssl1.1"],
+            DISTRO_PACKAGES["amd64"][distro_suffix]["openssl"],
```
mattmoor, line 72: cc @dekkagaijin 

I hear you want a distroless/base (with glibc) for use with K8s stuff?  In this PR I'm faking `arm64` support, but if we can plumb through reading the package lists, then we can extend my "arch" refactor to include at least this, and probably some of the java and python targets.

If we push through a second architecture all the way, I can volunteer to test ARM64 on a mix of my RPi4 and a Graviton2 EKS cluster.

mattmoor:
Alright, so I've been refactoring some of the logic to make it more regular, and adjust how it's generated to make it more amenable to becoming a matrix of stuff.

I have it to the point where I'm pulling in what I think we'd need in this PR to use real arm64 libs instead of faking it with the amd64 libs: https://github.com/GoogleContainerTools/distroless/pull/589

My next step will likely be to rebase this on that and do this "properly".  I'll probably just stick to "static" still, but this should let us get (at least) "base" pretty easily.

mattmoor:
I copied an image build with this here: `docker.io/mattmoor/distroless-static:arm64` I'm going to test things now.

mattmoor:
Alright, here it is on a Graviton cluster: https://pizzaplanetv2-test-image.default.mattmoor.io

base/base.bzl
```diff
@@ -3,52 +3,74 @@ load("@io_bazel_rules_docker//container:container.bzl", "container_image")
...
+                workdir = workdir,
+            )
+
+    for (user, uid, workdir) in [("root", 0, "/"), ("nonroot", NONROOT, "/home/nonroot")]:
+        native.alias(
+            name = "static_" + user + distro_suffix,
+            actual = ":static_" + user + "_amd64" + distro_suffix,
+        )
+
+    native.alias(
```
dekkagaijin, line 57: would be good to add a comment on things like this mentioning that there's nothing architecture-specific in the deb

mattmoor: That's not why this is here.  I'm following the pattern established in `BUILD`, which seems to be so that folks relying on the old target names aren't broken.

That said, I'm kinda inclined to rip them all out as part of this refactor, but without understanding the history better didn't want to jump the gun.

cloudbuild.yaml
```diff
@@ -43,15 +48,20 @@ steps:
...
+    docker tag bazel/base:static_root_amd64_debian9     gcr.io/$PROJECT_ID/static:${COMMIT_SHA}
+    docker tag bazel/base:static_root_amd64_debian9     gcr.io/$PROJECT_ID/static:latest
+    docker tag bazel/base:static_root_amd64_debian9     gcr.io/$PROJECT_ID/static-debian9:${COMMIT_SHA}
+    docker tag bazel/base:static_root_amd64_debian9     gcr.io/$PROJECT_ID/static-debian9:latest
+    docker tag bazel/base:static_root_amd64_debian10    gcr.io/$PROJECT_ID/static-debian10:${COMMIT_SHA}
+    docker tag bazel/base:static_root_amd64_debian10    gcr.io/$PROJECT_ID/static-debian10:latest
+    docker tag bazel/base:static_nonroot_amd64_debian9  gcr.io/$PROJECT_ID/static:nonroot
+    docker tag bazel/base:static_nonroot_amd64_debian9  gcr.io/$PROJECT_ID/static-debian9:nonroot
+    docker tag bazel/base:static_nonroot_amd64_debian10 gcr.io/$PROJECT_ID/static-debian10:nonroot
+
```
dekkagaijin, line 61: Wanna sprinkle some `docker manifest create` in here? :D

base/base.bzl
dekkagaijin: fair enough

cloudbuild.yaml
mattmoor: cc @jonjohnsonjr 

as I was hoping he'd do the multi-arch stitching, but given how this is publishing the images, maybe rules_docker work is wholly unnecessary? Regardless, out of scope for this change 😉

mattmoor:
Alright, I was able to `gcloud builds submit --substitutions=COMMIT_SHA=deadbeef` and it is publishing images now.

cloudbuild.yaml
mattmoor: Looks like I still need to add these to the images like to publish though.

mattmoor: Also, I have a PoC where I just enabled the experimental docker features and published a manifest list with these from GCB 🤩 , so expect another PR on top of this soon.  I'll fix `images:` here first, but I don't think we can use `images:` to publish the manifest lists, unfortunately.

mattmoor: ok, added to `images:` going to start on a PR that publishes manifest lists next.

mattmoor:
This is now rebased

dlorenc (approved):

Issue #585, opened 2020-09-26
Avoid rewriting WORKSPACE for updates


mattmoor:
I notice the SHAs (and other constants) checked into WORKSPACE, and have a suggestion:
https://github.com/GoogleContainerTools/distroless/blob/3e15f614453ff1923441576fba635259bd456f22/WORKSPACE#L23-L39

Currently this is patched in bash, which is fraught:
https://github.com/GoogleContainerTools/distroless/blob/3e15f614453ff1923441576fba635259bd456f22/updateWorkspaceSnapshots.sh#L43-L53

The basic idea is to have the script own generating an entire `.bzl` file defining constants that are included into WORKSPACE, so instead of running a bunch of seds, you can use a "here file":

```shell
cat > foo.bzl <<EOF

FOO = "$bar"
BAZ = "$blah"
...

EOF
```

I'm looking at this through the lens of what it might take to refactor things for multi-arch, so I'd plug putting some forward thought into how that might affect a forward-looking design (define maps, or arch-per-file?).

cc @loosebazooka @chanseokoh @jonjohnsonjr 

I may play with this later, but need to set up a suitable environment to start doing dev on these again 😅

mattmoor:
FWIW, I think the sort of separation I'm suggesting is what I did here: https://github.com/bazelbuild/rules_docker/blob/master/repositories/repositories.bzl

mattmoor:
see: https://github.com/GoogleContainerTools/distroless/issues/583

PR #587, opened 2020-09-27
Normalize rule naming across versions

mattmoor:


mattmoor:
Alright, I think this gets things about as regular as I'd want to bite off in a single change.  With this in hand, I think the next chunk I'd want to TAL at is https://github.com/GoogleContainerTools/distroless/issues/585, which should make expanding to include a new architecture quite a bit simpler.

I'm going to start on that in a separate PR based on this one.

mattmoor:
@chanseokoh I rebased this on the update change you landed this morning.

dlorenc (approved):

PR #592, opened 2020-09-28
Remove Aliasing

mattmoor:
Make everything use explicit references to avoid confusion.  It's more verbose, but hey...  it's Bazel after all.

Issue #286, opened 2019-01-24
Remove cloudbuild.yaml


chanseokoh:
It looks like the file was introduced as a workaround because the legacy `docker_push` rule stopped working. (Correct me if I'm wrong.) With the new rules (PR: #280), it works again in my testing.

chanseokoh:
But looks like there's an automatic publish hook using it (https://github.com/GoogleContainerTools/distroless/pull/283#issuecomment-457736388), so I guess `cloudbuild.yaml` should stay for a while.

chanseokoh:
A deviation that `cloudbuild.yaml` producing additional tags: https://github.com/GoogleContainerTools/distroless/pull/297

donmccasland:
We are using cloudbuild and triggers to automatically build and publish images.  We will not be changing that.

PR #595, opened 2020-09-29
[multi-arch] Make distroless/base:{latest,nonroot,debug,debug-nonroot} multi-arch

mattmoor:
Based on: https://github.com/GoogleContainerTools/distroless/pull/592


loosebazooka:
Kinda has all the alias stuff still in here? Is that intentional?

mattmoor:
@loosebazooka yeah, I've been stacking things for my own sanity, see the PR note.

I can throw a WIP in the title until they are ready, I just haven't been because it's not semantic like with Prow.

loosebazooka:
Yeah, I think I wasn't sure if you intended on merging  #592 soon or not.

mattmoor:
@loosebazooka I was, I was waiting to see if anyone screamed about the prior change, but we can always rollback both if you want it sooner?

I figured as long as it was only blocking me, I'd wait and keep stacking things against it 😅

loosebazooka:
should we use a feature branch?

mattmoor:
@loosebazooka we can do that 👍 

Lemme set that up quickly and I'll repoint this PR.

mattmoor:
@loosebazooka ok, I created the `multi-arch` feature branch off master, and this has both the changes in it.  I'll leave the other PR alone for now, but we can land this into the feature branch if folks want to come party on things in parallel and send PRs there.

mattmoor:
Actually, don't merge this yet, I think it still has some cloudbuild.yaml issue 🤔

mattmoor:
Trying another GCB run with this latest version.  I hadn't reworked the `docker push`, just the `manifest push`.

mattmoor:
(kicking travis)

BUILD
```diff
@@ -61,12 +61,12 @@ container_bundle(
         "gcr.io/{PROJECT_ID}/java-debian10/jetty:java11": "//java/jetty:jetty_java11_debian10",
         "gcr.io/{PROJECT_ID}/java-debian10/jetty:debug": "//java/jetty:jetty_java11_debug_debian10",
         "gcr.io/{PROJECT_ID}/java-debian10/jetty:java11-debug": "//java/jetty:jetty_java11_debug_debian10",
-        "gcr.io/{PROJECT_ID}/python3:latest": "//experimental/python3:python3",
-        "gcr.io/{PROJECT_ID}/python3:nonroot": "//experimental/python3:python3_nonroot",
-        "gcr.io/{PROJECT_ID}/python3:debug": "//experimental/python3:debug",
-        "gcr.io/{PROJECT_ID}/python3:debug-nonroot": "//experimental/python3:debug_nonroot",
-        "gcr.io/{PROJECT_ID}/python2.7:latest": "//experimental/python2.7:python27",
-        "gcr.io/{PROJECT_ID}/python2.7:debug": "//experimental/python2.7:debug",
+        "gcr.io/{PROJECT_ID}/python3:latest": "//experimental/python3:python3_debian9",
```
loosebazooka, line 64: maybe @chanseokoh knows, did we not ever do debian10 images for python? Is it because experimental?

mattmoor: AFAIK this file isn't actually used for anything real, since `cloudbuild.yaml` enumerates all the targets explicitly, so I'd take this with a grain of salt.

mattmoor:
I have `//cc` mostly done, so once GCB is happy with the changes here, I'll start to stage that.  Do you want to TAL at Java on top of that once it is up?  I'm happy to test a simple arm64 image for you if you don't have hardware.

I tested this with `//examples/go` tweaked to target/use arm64, I'll need to sort out how best to test `//cc`, since I'm guessing cross-compilation in Bazel will be trickier.

base/base.bzl
```diff
@@ -126,17 +96,17 @@ def distro_components(distro_suffix):
...
         configs = ["testdata/" + distro_suffix[1:] + ".yaml"],
-        image = ":debug" + distro_suffix,
+        image = ":debug_root_amd64" + distro_suffix,
     )
 
     container_test(
         name = "static_release" + distro_suffix + "_test",
         configs = ["testdata/" + distro_suffix[1:] + ".yaml"],
-        image = ":static" + distro_suffix,
+        image = ":static_root_amd64" + distro_suffix,
```
loosebazooka, line 111: do we need to add `container_test` blocks for all the `arm64` images?

mattmoor: I didn't because then we'd need arm64 hardware to test them on presubmit, which starts to complicate things.

I've been cross-compiling this stuff and testing samples against a real arm64 cluster.  I hear that Bazel just started shipping arm64 binaries, so you might see if they have a solution for CI that this repo can piggyback on 🤔

BUILD
chanseokoh: Yeah, `container_bundle` is to build a local tarball with all the images for `docker save` and might be useful for some people during development. (I don't use it.) I tried to sync this with `cloudbuild.yaml`, but I guess I missed Python Debian 10 here.

chanseokoh:
> Do you want to TAL at Java on top of that once it is up? I'm happy to test a simple arm64 image for you if you don't have hardware.

We don't have arm64 hardware, but I believe it's easy to set up arm64 Travis builds. Distroless has Java tests that actually run images (including accessing `https://google.com` through java code and verifying Java versions by running `java --version`). I think that should be enough.

BUILD
mattmoor: I believe the original intent was to allow `cloudbuild.yaml` to use the bundle in conjunction with a `container_pushall` rule so that the `cloudbuild.yaml` could just becomes `bazel run //:pushall`, however, that would make us beholden to the absence of multi-arch support in rules_docker, so I'm sort of glad we're not stuck with that.

I pointed @jonjohnsonjr at this, so perhaps if/when we adds multi-arch to rules_docker, we can consolidate a bit here and `cloudbuild.yaml` can simplify dramatically 🤩

chanseokoh: Oh yeah, I remember opening #286.

mattmoor:
@chanseokoh Awesome, yeah I've heard they may have qemu support for more obscure hardware platforms like s390x as well, so perhaps this will be useful as we expand the matrix further.  Do we want to merge this into the feature branch, and then we can explore that on top of this?  the GCB build completed successfully this time.

I'm doing one final check of the busybox bits on arm64 hardware now.

mattmoor:
The other unknown (to me) is the right way to filter tests by architecture with Bazel, since not all of these will pass all places

mattmoor:
Alright, I found one more bug, rerunning GCB with the latest.

mattmoor:
This is the latest issue that just bit me: https://github.com/bazelbuild/rules_docker/issues/1636

mattmoor:
I think that this is almost exactly the matrix we are going to want in the fullness of time: https://docs.travis-ci.com/user/multi-cpu-architectures/#example-multi-architecture-build-matrix

chanseokoh:
I'd just merge the alias PR. As of now, given the current state of Distroless, all the Bazel rules are internal code and not for public consumption. Only those images published to `gcr.io/distroless/**` are considered public. We (and anyone) should freely restructure or refactor the internal Bazel code at this point. Also we provide zero support for anyone depending on the internal Bazel rules.

mattmoor:
Alright, I have verified that `debug` works properly, so the busybox bits are validated.

I merged the alias PR, so this is just the `//base:{base,debug}` changes.

I'm going to stage the `//cc` changes and then come back here to experiment with travis on arm64

mattmoor:
I'm closing this in favor of: https://github.com/GoogleContainerTools/distroless/pull/600

PR #597, opened 2020-09-29
Add a buildifier github action

mattmoor:


mattmoor:
I added some errors in a prior commit and you can annotations like this on the PR code:
image

chanseokoh:
Basically this seems like a nice idea, but because we are already running buildifier on Travis, I am not sure why we want to add another duplicate check? Is this really necessary? Given the current state of Distroless, I'm trying to not increase code complexity and maintenance burden as much as possible.

mattmoor:
@chanseokoh I am trying to eliminate the travis check, because we don't need this check running on multiple architectures.

chanseokoh:
Sounds good. Just make sure to keep it in `./test.sh` or document running buildifier locally so that it makes contributors run it before putting up a PR that has a style issue.

mattmoor:
I removed the old way and documented both this and `pylint` in CONTRIBUTING.md

mattmoor:
I'm going to look at pylint next.

chanseokoh (approved):

PR #600, opened 2020-09-29
Run tests on arm64 as well

mattmoor:
Enable travis builds of arm64 in addition to amd64, and extend the multi-arch support to include everything under `//base`, including the `container_test` rules.

This required quite a few changes across a number of repos.  Thanks to @nkubala for the quick turnaround on PRs to the structure test repo, and the 1.9.1 release.  The one outstanding PR is to pull the 1.9.1 release into rules_docker here: https://github.com/bazelbuild/rules_docker/pull/1639.

Unfortunately, Bazel's support for filtering targets by architecture is still pretty weak.  I tried a half dozen options from constraints (`restricted_to`) to config settings (`select`) and at best I could make the targets on mismatched platforms emit an error, but Bazel doesn't yet support filtering them (and the issue is >3 years old).

The workaround I was able to find is similar to how Google runs integration testing internally, which is to tag the tests as `tags = ["manual"]` and then to run specific targets, however, I also included the architecture the test is expected to run on, e.g. `tags = ["manual", "amd64"]`.  Then in Travis I use `bazel query` to pull out all of the appropriate tests for each architecture and run them explicitly.  This required me to tag all of the existing tests (that haven't moved) as `amd64` explicitly, but as we move these to loops over `ARCHITECTURES` we would use the appropriate architecture and the `arm64` support should "just work" 🤞 

We shouldn't merge this until I'm able to rebase on the merged rules_docker PR, but I'd appreciate it if folks would take a look and leave any other comments they might have. 🙏


examples/rust/BUILD
```diff
@@ -8,8 +8,8 @@ load("@io_bazel_rules_docker//rust:image.bzl", "rust_image")
 
 # NOTE: Bazel Rust rules don't support cross-compilation yet,
 # so at this time it's required that you build on the same platform as the image
-rust_image(
-    name = "rust_example",
-    srcs = ["src/main.rs"],
-    base = "//cc:cc_debian10",
-)
+# rust_image(
```
mattmoor, line 11: I forgot about this.  I'll have to see if tags solves the problem here as well :(

WORKSPACE
```diff
@@ -370,18 +370,25 @@ http_archive(
...
+
 # Docker rules.
 http_archive(
     name = "io_bazel_rules_docker",
-    sha256 = "4521794f0fba2e20f3bf15846ab5e01d5332e587e9ce81629c7f96c793bb7036",
-    strip_prefix = "rules_docker-0.14.4",
-    urls = ["https://github.com/bazelbuild/rules_docker/archive/v0.14.4.tar.gz"],
+    sha256 = "cf53839c398e464b10ec2fbeb11aedb446f078c28e3b4ce372461bb105ef435c",
+    strip_prefix = "rules_docker-f8478e57ab7457e403fda474f06ac0bb120d92a7",
+    urls = ["https://github.com/mattmoor/rules_docker/archive/f8478e57ab7457e403fda474f06ac0bb120d92a7.tar.gz"],
```
chanseokoh, line 391: Is this the URL we want?

cc/BUILD
```diff
@@ -4,15 +4,19 @@ load("//base:distro.bzl", "DISTRO_PACKAGES", "DISTRO_SUFFIXES")
...
+            DISTRO_PACKAGES["amd64"][distro_suffix]["libgcc1"],
+            DISTRO_PACKAGES["amd64"][distro_suffix]["libgomp1"],
+            DISTRO_PACKAGES["amd64"][distro_suffix]["libstdc++6"],
+        ],
+    )
+    for mode in [
+        "",
+        ":debug",
+    ]
+    for distro_suffix in ("_debian9", "_debian10")
```
chanseokoh, line 21:
```suggestion
    for distro_suffix in DISTRO_SUFFIXES
```

cpus/BUILD
```diff
@@ -0,0 +1,11 @@
+load("//:checksums.bzl", "ARCHITECTURES")
```
chanseokoh, line 1: Is this `load(...)` necessary?

And didn't you say in your PR description that you're not going to use `config_setting`?

cloudbuild.yaml
```diff
@@ -62,21 +70,26 @@ steps:
...
+    docker tag bazel/base:debug_root_amd64_debian9       gcr.io/$PROJECT_ID/base-debian9:debug
+    docker tag bazel/base:debug_root_amd64_debian10      gcr.io/$PROJECT_ID/base-debian10:debug
+    docker tag bazel/base:debug_nonroot_amd64_debian9    gcr.io/$PROJECT_ID/base:debug-nonroot-amd64
+    docker tag bazel/base:debug_nonroot_amd64_debian9    gcr.io/$PROJECT_ID/base-debian9:debug-nonroot
+    docker tag bazel/base:debug_nonroot_amd64_debian10   gcr.io/$PROJECT_ID/base-debian10:debug-nonroot
+
+    docker tag bazel/base:base_root_arm64_debian9        gcr.io/$PROJECT_ID/base:latest-arm64
+    docker tag bazel/base:base_root_arm64_debian9        gcr.io/$PROJECT_ID/base:nonroot-arm64
+    docker tag bazel/base:debug_root_arm64_debian9       gcr.io/$PROJECT_ID/base:debug-arm64
+    docker tag bazel/base:debug_root_arm64_debian9       gcr.io/$PROJECT_ID/base:debug-nonroot-arm64
```
chanseokoh, line 92:
```suggestion
    docker tag bazel/base:base_root_arm64_debian9        gcr.io/$PROJECT_ID/base:latest-arm64
    docker tag bazel/base:base_nonroot_arm64_debian9     gcr.io/$PROJECT_ID/base:nonroot-arm64
    docker tag bazel/base:debug_root_arm64_debian9       gcr.io/$PROJECT_ID/base:debug-arm64
    docker tag bazel/base:debug_nonroot_arm64_debian9    gcr.io/$PROJECT_ID/base:debug-nonroot-arm64
```

And likewise, eventually we should also tag these in `-debian9`.

cloudbuild.yaml
```diff
@@ -62,21 +70,26 @@ steps:
...
-    docker tag bazel/base:debug_debian9           gcr.io/$PROJECT_ID/base:debug
-    docker tag bazel/base:debug_debian9           gcr.io/$PROJECT_ID/base-debian9:debug
-    docker tag bazel/base:debug_debian10          gcr.io/$PROJECT_ID/base-debian10:debug
-    docker tag bazel/base:debug_nonroot_debian9   gcr.io/$PROJECT_ID/base:debug-nonroot
-    docker tag bazel/base:debug_nonroot_debian9   gcr.io/$PROJECT_ID/base-debian9:debug-nonroot
-    docker tag bazel/base:debug_nonroot_debian10  gcr.io/$PROJECT_ID/base-debian10:debug-nonroot
+    docker tag bazel/base:base_root_amd64_debian9        gcr.io/$PROJECT_ID/base:${COMMIT_SHA}
+    docker tag bazel/base:base_root_amd64_debian9        gcr.io/$PROJECT_ID/base:latest-amd64
+    docker tag bazel/base:base_root_amd64_debian9        gcr.io/$PROJECT_ID/base-debian9:${COMMIT_SHA}
+    docker tag bazel/base:base_root_amd64_debian9        gcr.io/$PROJECT_ID/base-debian9:latest
```
chanseokoh, line 76: Ideally and eventually, I think this should be a manifest list. We've maintained that `base` and `base-debian9` are basically identical. (And same for `-debian10`.) But I guess you're adding manifest lists only to `base` as a start?

chanseokoh (commented):
This is cool stuff!

I don't have much expertise in Bazel, so Unfortunately I cannot comment on what would be the best approach to build and test based on the environment. (Strange, I feel like this kind of feature should be a basic component in any testing frameworks.) But as long as it works, this is a great progress.

WORKSPACE
mattmoor: whoops, no 🤦

cpus/BUILD
mattmoor: Indeed, this is unused, I just forgot to clean it up.

cc/BUILD
mattmoor: Sorry, merge conflict.  I had a similar change in my later PRs 😅

cloudbuild.yaml
mattmoor: Good catch!  re: `-debian9` I think we need a better strategy for scripting the variants, but totally agree.  This is quite tedious to keep straight as-is 😅

cloudbuild.yaml
mattmoor: Yeah, I wanted to plumb through the main things for each first.  I think we need a bigger discussion around how we (likely) script these variations to cut down on the boilerplate (and sorts of errors I had above)

mattmoor:
I pushed a commit with the couple edits called out.  I added the `debian9/10` variant topic to my list.  Once I land the open chain I'll open issues for the remaining pieces of work on my radar (java, node, dotnet, this, other architectures like ppc and s390x).

base/base.bzl
```diff
@@ -48,79 +49,107 @@ def distro_components(distro_suffix):
...
-        entrypoint = ["/busybox/sh"],
-        env = {"PATH": "$$PATH:/busybox"},
-        tars = ["//experimental/busybox:busybox.tar"],
-    )
+            # A debug image with busybox available.
+            container_image(
+                name = "debug_" + user + "_" + arch + distro_suffix,
+                architecture = arch,
+                base = ":base_" + user + "_" + arch + distro_suffix,
+                directory = "/",
```
chanseokoh, line 68: Maybe `directory` is unnecessary, as it's [default](https://github.com/bazelbuild/rules_docker#container_image-1) is `/`?

base/base.bzl
```diff
@@ -48,79 +49,107 @@ def distro_components(distro_suffix):
...
-        configs = ["testdata/base.yaml"],
-        image = ":base" + distro_suffix,
-    )
+        ##########################################################################################
+        # Check that we can invoke openssl in the base image to check certificates.
+        ##########################################################################################
+        container_test(
+            name = "openssl_" + arch + distro_suffix + "_test",
+            configs = ["testdata/certs.yaml"],
+            image = ":base_root_" + arch + distro_suffix,
```
chanseokoh, line 109: Is it meaningful to parameterize user (here and tests below) given that we're in a loop, or it is not worth?

chanseokoh (commented):
Looks good to me.

base/base.bzl
mattmoor: We could look at doing this more, it wasn't clear to me if/which of these tests might do something privileged, so I tried to avoid creeping the scope too much.

The one thing I did make a point to do was make the pure-go test use static, which I believe previously had zero direct test coverage.

mattmoor:
@chanseokoh if it's alright I am going to punt on further churn here (since both are pre-existing limitations, and I want to maintain sanity rebasing the other changes), but I agree with the cleanups / experiments you suggest.

I can try to follow up with these on top of the fuller chain of changes 🙏

mattmoor:
Note: there are some other funny warts I haven't addressed, like plumbing root/nonroot through all the variants (I don't think //cc or python2.7 have it, but python3 does).

chanseokoh (approved):
Sounds good! Thanks for the awesome work.

PR #604, opened 2020-09-30
Test things with rules_docker at HEAD

mattmoor:


mattmoor:
I staged this because I might need to get some fixes upstreamed to rules_docker, and I wasn't sure how far behind HEAD we were.  Looks like things pass at HEAD. 😌

Issue #222, opened 2018-07-24
docker multiarch image - add ppc64le/s390x architectures to docker image


grooverdan:
I was looking to use the distroless as a base for skydive (https://hub.docker.com/r/skydive/skydive/) on all architectures but it was only accessible on x86_64.

It seems bazel has a cross-compilation capability to make this happen.

Thanks the great contribution to container security.

dlorenc:
We have no way to test images on these architectures, so I'm a bit hesitant to ship binaries for them. Would you be interested in forking/hosting your own binaries in a different project?

grooverdan:
There is a travis `os: linux-ppc64le` for running builds/tests on real hardware. I've got an example multiarch builds/deployment with https://github.com/grooverdan/hello-world/ / https://travis-ci.org/grooverdan/hello-world.

If the package `qemu-user-static` (tested Fedora and Ubuntu) is installed on the running OS, and `/usr/bin/qemu-${TARGET_GOARCH}-static` is copied inside the container, then you can have a fully emulated hardware test container. I certainly understand if this doesn't meet the reassurance criteria.

I'm not yet ready to run my own project (first exposure to Bazel) yet but I'll probably get time eventually to hack on the dependencies (https://github.com/bazelbuild/rules_docker/issues/300? / https://github.com/bazelbuild/bazel/issues/5657? others?). I certainly can help out with testing (running/writing).

donmccasland:
I'm not sure we can ever support this, but if someone wants to take a crack at it, I'm ready to review PRs and build results.

mattmoor:
It should be relatively straightforward to add additional architectures following what I'm doing for arm64 here (and it's baseline changes): https://github.com/GoogleContainerTools/distroless/pull/586

PR #612, opened 2020-10-08
Build s390x distroless base and static images

knm3000:
Issue https://github.com/GoogleContainerTools/distroless/issues/222
- Added build scripts for s390x distroless base and static images for debian9 and debian10
- do not include security updates for Debian9 s390x (s390x debian9 security update repo does not exist)
- excluded "libunwind8" package for Debian9/10 s390x (s390x repos do not have this package)


barthy1:
@mattmoor this is part of Tekton on `s390x` initiative. 
We need to add `s390x` support to distroless to be able to use current base Tekton `gcr.io/distroless/..` [images](https://github.com/tektoncd/pipeline/blob/master/.ko.yaml) with `ko resolve --platform=all`. See discussion [here](https://github.com/tektoncd/community/pull/211#discussion_r496701364).

It would be great if you can provide your input already, while we are fixing Travis build error.

nodejs/BUILD
```diff
@@ -6,6 +6,8 @@ load("//:checksums.bzl", "ARCHITECTURES")
 
 NODEJS_MAJOR_VERISONS = ("10", "12", "14")
 
+ARCHITECTURES = [arch for arch in ARCHITECTURES if arch != "s390x"]
```
mattmoor, line 9: Perhaps instead of this we should just special case static?

mattmoor: or all of base

WORKSPACE
```diff
@@ -63,6 +63,7 @@ load(
     )
     for arch in ARCHITECTURES
     for (name, distro) in VERSIONS
+    if "security" in SHA256s[arch][name]
```
mattmoor, line 66: There's no security mirror...?

mattmoor:
Not thrilled about the `if arch != "s390x"` everywhere, I'd like to understand the path to removing these

WORKSPACE
knm3000: security.debian.org does not have s390x for debian stretch:
http://security.debian.org/debian-security/dists/stretch/updates/main/
and there are no any other mirrors that include s390x.

knm3000:
@mattmoor thanks for the review, I've removed if `arch != "s390x"` and defined BASE_ARCHITECTURES instead, can you please take a look?

WORKSPACE
```diff
@@ -79,222 +81,256 @@ load(
...
-            "libgcrypt20",
-            "libgpg-error0",
-            "libacl1",
-            "libattr1",
-            "libselinux1",
-            "libpcre3",
-            "libbz2-1.0",
-            "liblzma5",
-        ],
+        packages = PACKAGES_DEBIAN9,
```
mattmoor, line 187: I think it would actually simplify the logic to make this:
```
   packages = PACKAGES_DEBIAN9 + if arch in BASE_ARCHITECTURES: ["libunwind8"] else [],
```

You could even leave `PACKAGES_DEBIAN9` inlined here, and make this diff *much* smaller.

examples/nodejs/BUILD
```diff
@@ -5,7 +5,7 @@ package(default_visibility = ["//visibility:public"])
 
 load("@io_bazel_rules_docker//container:container.bzl", "container_image")
 load("//base:distro.bzl", "DISTRO_SUFFIXES")
-load("//:checksums.bzl", "ARCHITECTURES")
+load("//:checksums.bzl", "BASE_ARCHITECTURES")
```
mattmoor, line 8:
```suggestion
load("//:checksums.bzl", ARCHITECTURES="BASE_ARCHITECTURES")
```

much smaller diff

updateWorkspaceSnapshots.sh
```diff
@@ -34,7 +36,8 @@ cat > checksums.bzl <<EOF
 # DO NOT MODIFY THIS FILE DIRECTLY.
 # TO GENERATE THIS RUN: ./updateWorkspaceSnapshots.sh
 
-ARCHITECTURES = ["amd64", "arm64"]
+ARCHITECTURES = ["amd64", "arm64", "s390x"]
+BASE_ARCHITECTURES = ["amd64", "arm64"]
```
mattmoor, line 40:
```suggestion
BASE_ARCHITECTURES = ["amd64", "arm64"]
ARCHITECTURES = BASE_ARCHITECTURES + ["s390x"]
```

mattmoor (commented):
Few comments inline about how we can reduce the churn of this a bunch.

namrata-ibm:
@knm3000 Thank you for this PR. 
Noticed that cloudbuild.yaml does not include `docker push 'gcr.io/$PROJECT_ID/static:latest-s390x'`. Isn't it required during `manifest create`?

knm3000:
@mattmoor I've updated ARCHITECTURES / packages code as you suggested (and also included missing 'docker push' lines from @namrata-ibm comment), can you please review? Should I should squash my commits into one?

mattmoor:
This looks about as clean as we can make it.  @chanseokoh WDYT?

chanseokoh (approved):
@mattmoor I'm not versed in Bazel and Python, so I trust your opinion.

Merging this now. Those images will be auto-published within an hour. If they don't, please update here, as it would mean this PR broke `cloudbuild.yaml`.

chanseokoh:
Oh, Travis is still running.

chanseokoh:
Too bad, a step in `cloudbuild.yaml` became too long, and GCB failed to publish the images.

```
Your build failed to run: generic::invalid_argument: invalid build: invalid .steps field: build step 1 arg 1 too long (max: 4000)
```

@knm3000 would you fix `cloudbuild.yaml` by perhaps splitting the modified step into two?

Issue #602, opened 2020-09-30
Suggestion: better approach to package and version management


petermylemans:
Also relates to #153 and the clean up already done by @mattmoor (thanks btw!). During further investigation I've stumbled on some work in rules_pkg github repository that looks like an improved version of the deb management that is currently in the distroless workspace.

Advantages:
 - It has a nice tool [2] that automates rewriting the workspace using buildozer instead of regex expressions, which is a poor man's solution.
 - A lot more readable compared to the current approach in distroless
 - Performance of downloading dependencies is much better and more stable due to the use of builtin bazel workspace rules instead of a custom python program.
 - Support for 3rd party debian repositories when required. This could open up distroless to support more recent versions of the software stacks included in the scope, while still using the stable foundation of debian for the base packages.

Things to improve:
 - it is not possible at the moment to have multiple sources (e.g. debian, debian-updates and debian-security) while evaluating the most recent version, so the code needs to be adapted to support that. The work required is not too difficult. A syntax proposal is outlined below.
 - from what I can tell it is not packaged in rules_pkg, but instead lives as a separate workspace in the same git repo. 

@mattmoor @chanseokoh @aiuto feel free to review this idea. I'm happy to start a PR in either repository if we agree this is a good way forward.

References:
[1] https://github.com/bazelbuild/rules_pkg/blob/main/deb_packages/WORKSPACE 
[2] https://github.com/bazelbuild/rules_pkg/tree/main/deb_packages/tools/update_deb_packages

```
http_file(
    name = "buster_archive_key",
    sha256 = "9c854992fc6c423efe8622c3c326a66e73268995ecbe8f685129063206a18043",
    urls = ["https://ftp-master.debian.org/keys/archive-key-10.asc"],
)
http_file(
    name = "buster_security_archive_key",
    sha256 = "4cf886d6df0fc1c185ce9fb085d1cd8d678bc460e6267d80a833d7ea507a0fbd",
    urls = ["https://ftp-master.debian.org/keys/archive-key-10-security.asc"],
)

deb_packages(
    name = "debian",
    arch = "amd64",
    distro = "buster",
    packages = {
        "base-files": "http://ftp.debian.org/debian/pool/main/b/base-files/base-files_10.3+deb10u6_amd64.deb",
        "busybox": "http://ftp.debian.org/debian/pool/main/b/busybox/busybox_1.30.1-4_amd64.deb",
[...]

Issue #153, opened 2017-12-14
Download packages more robustly


drigz:
Bazel's downloader is fairly robust: when you use a 'mirror.bazel.build' URL with `http_archive`, you can be confident it will not error out when fetching. However, as dpkg_parser.py uses urllib2 directly, it doesn't take advantage of this reliability.

From skimming #72, it seems like work is underway to allow fixing a package version, and hosting it in a mirror. It would be good to take advantage of Bazel's downloader as part of this redesign.

For example, one error we've seen in CI:

```
ERROR: error loading package 'src/docker': Encountered error while reading extension file 'file/packages.bzl': no such package '@package_bundle//file': dpkg_parser command failed: Traceback (most recent call last):
  File "/usr/lib/python2.7/runpy.py", line 162, in _run_module_as_main
    "__main__", fname, loader, pkg_name)
  File "/usr/lib/python2.7/runpy.py", line 72, in _run_code
    exec code in run_globals
  File "/home/kbuilder/.cache/bazel/_bazel_kbuilder/external/dpkg_parser/file/dpkg_parser.par/__main__.py", line 188, in <module>
  File "/home/kbuilder/.cache/bazel/_bazel_kbuilder/external/dpkg_parser/file/dpkg_parser.par/__main__.py", line 79, in main
  File "/home/kbuilder/.cache/bazel/_bazel_kbuilder/external/dpkg_parser/file/dpkg_parser.par/__main__.py", line 107, in download_dpkg
  File "/usr/lib/python2.7/socket.py", line 351, in read
    data = self._sock.recv(rbufsize)
  File "/usr/lib/python2.7/httplib.py", line 578, in read
    return self._read_chunked(amt)
  File "/usr/lib/python2.7/httplib.py", line 632, in _read_chunked
    raise IncompleteRead(''.join(value))
httplib.IncompleteRead: IncompleteRead(4096 bytes read)
 (/home/kbuilder/.cache/bazel/_bazel_kbuilder/external/dpkg_parser/file/dpkg_parser.par --package-files /home/kbuilder/.cache/bazel/_bazel_kbuilder/external/debian_stretch/file/Packages.json,/home/kbuilder/.cache/bazel/_bazel_kbuilder/external/debian_stretch_backports/file/Packages.json --packages libpython2.7 --workspace-name package_bundle)
```

MarkusTeufelberger:
[...]

PR #614, opened 2020-10-10
Update package manager

petermylemans:
Goal: get rid of the python package manager and switch to a more stable solution that uses bazel builtin functions as much as possible.

Updating packages is as easy as running `bazel run update_deb_packages`. This will update the workspace with the latest available versions for the respective distribution and/or sources. I've also added PGP validation for the release files to reject any corrupted files.
Third party sources (e.g.  AdoptOpenJDK deb repo) are possible, but not in use at the moment.

- [X] Archive dependencies should live in the WORKSPACE file, but due to the sheer number of packages, it is getting big... Need another approach?
- [X] packages_amd64_debian10 and packages_arm64_debian10 are very similar in terms of packages. The reason I can't join them into a single rule is that bazel doesn't support nested dicts as attributes as far I know. 
I'm also not sure whether all packages exist in all archs, so might be best to keep them separate
` deb_packages(packages = { "some_pack": { "amd64": "url...", "arm64": "another url" } )` is not allowed.
- [X] Have a look at JAVA_VERSION FIXME
- [X] Have a look at the /etc/os-release FIXME. Should be included in base-files.
- [X] Clean up updater and version files from "old solution"
- [X] Clean up python package manager. Only dpkg_extract.sh remains (due to usage in locale en cacerts rules).

Relates to #602 #153


chanseokoh:
Cool stuff! Haven't looked into it seriously, but it appears to clean things up a lot as it depends on an external rule. I just hope the rule is well supported in the long run.

> I'm also not sure whether all packages exist in all archs,

This is not the case. I've seen it in https://github.com/GoogleContainerTools/distroless/pull/612#issue-500157699:

> excluded "libunwind8" package for Debian9/10 s390x (s390x repos do not have this package)

petermylemans:
Yes I also saw that PR, so I've opted to use a .bzl file per repo per arch for that reason. 

With respect to maintaining the external rules: I guess that's the story of all OSS projects especially the small one's. 

In this case it is a limited code base, but useful beyond distroless. You could e.g. use it to extend a distroless base image with additional debs in a reproducible way. 

I'm happy to move the code towards an organisation if that alleviates some of your concerns. In this case I look to some of the other rules packagers for guidance?

chanseokoh:
> I'm happy to move the code towards an organisation if that alleviates some of your concerns.

Looks like no one at Google wants to be receive it. Do you know some organization to move the code to?

petermylemans:
Not at the moment apart from creating a new one?

But one could argue that it is probably better in that case to make it internal to distroless? 
I guess the reasoning of other projects is "too specific to distroless' use case"?

petermylemans:
I'll have to rethink the approach a bit based on the fact that packages can disappear from ftp master. 
Snapshots (as we are currently using it) seems to be the only way to have fully reproducible builds.

PR #616, opened 2020-10-13
added common build & publish script

morlay:
## workflow

* common build `./hack/build.sh`
  * support switch `USER=${{ matrix.user }} DISTRO=${{ matrix.distro }} ./hack/build.sh ${{ matrix.target }}`
  * with a local registry service for cache bazel docker build for single arch
* use buildx to build and push to `ghcr.io`

## build example
https://github.com/morlay/distroless/runs/1248863511 

published images: https://github.com/morlay?tab=packages
image

## setup
https://docs.github.com/en/free-pro-team@latest/packages/getting-started-with-github-container-registry/enabling-github-container-registry-for-your-organization

* create a persional token for member of GoogleContainerTools https://github.com/settings/tokens/new
   * enable `write:packages` only
* add REPO or ORG secrets `CR_PAT` with the created persional token.

### DIFF
--- a/.github/workflows/publish-debug.yaml
+++ b/.github/workflows/publish-debug.yaml
@@ -1,0 +1,57 @@
+name: publish-debug
+
+on:
+  push:
+    branches: [ 'master' ]
+    paths:
+      - .github/workflows/publish-debug.yaml
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Install bazelisk
+        run: |
+          curl -LO "https://github.com/bazelbuild/bazelisk/releases/download/v1.7.1/bazelisk-linux-amd64"
+          mkdir -p "${GITHUB_WORKSPACE}/bin/"
+          mv bazelisk-linux-amd64 "${GITHUB_WORKSPACE}/bin/bazel"
+          chmod +x "${GITHUB_WORKSPACE}/bin/bazel"
+
+      - name: login ghcr.io
+        uses: docker/login-action@v1
+        with:
+          registry: ghcr.io
+          username: ${{ github.repository_owner }}
+          password: ${{ secrets.CR_PAT }}
+
+      - name: setup
+        run: |
+          bazel build --host_force_python=PY2 //package_manager:dpkg_parser.par
+          sed -i 's/^{/{"experimental": "enabled",/g' ~/.docker/config.json
+
+      - name: build //base:*
+        env:
+          HUB: ghcr.io/${{ github.repository }}
+          LANG: base
+          USERS: root nonroot
+          DISTROS: debian9 debian10
+          TOPICS: base debug static
+          ARCHS: amd64 arm64 s390x
+        run: |
+          ./hack/bazel_run.sh
+          PUSH=1 MANIFEST=1 ./hack/docker_tag_or_push.sh
+
+      - name: build //cc:*
+        env:
+          HUB: ghcr.io/${{ github.repository }}
+          LANG: cc
+          USERS: root nonroot
+          DISTROS: debian9 debian10
+          TOPICS: cc debug
+          ARCHS: amd64 arm64
+        run: |
+          ./hack/bazel_run.sh
+          PUSH=1 MANIFEST=1 ./hack/docker_tag_or_push.sh
--- a/hack/bazel_run.sh
+++ b/hack/bazel_run.sh
@@ -1,0 +1,21 @@
+#!/bin/bash
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+LANG=${LANG:-"base"}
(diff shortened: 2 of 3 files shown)


.github/workflows/publish-nodejs.yaml
```diff
@@ -0,0 +1,72 @@
...
+          - nodejs/nodejs10_debug
+          - nodejs/nodejs12
+          - nodejs/nodejs12_debug
+          - nodejs/nodejs14
+          - nodejs/nodejs14_debug
+        distro:
+          - debian9
+          - debian10
+        user:
+          - root
```
morlay, line 25: nonroot not support for nodejs

.github/workflows/publish-base.yaml
```diff
@@ -0,0 +1,70 @@
...
+          - debian10
+        user:
+          - root
+          - nonroot
+
+    runs-on: ubuntu-latest
+
+    services:
+      registry:
+        image: registry:2
```
morlay, line 29: local hub

.github/workflows/publish-base.yaml
```diff
@@ -0,0 +1,70 @@
...
+        ports:
+          - 5000:5000
```
morlay, line 31: local hub `localhost:5000`

chanseokoh:
Thanks for putting up the PR, but I think we don't have to migrate over to GitHub Actions now. That can wait. I vaguely remember you are trying to add more tags to the `*-debian9` and `*-debian10` repos? If so, please take an incremental step to fix that in `cloudbuild.yaml` (and one thing at a time).

morlay:
@chanseokoh 

i got `*-debian10` by custom build. 
my purpose of building istio-distroless work well https://hub.docker.com/layers/querycapistio/proxyv2/1.7.3-distroless/images/sha256-8c87dee4d4dd8f7d361a122ed4bd034d397d0d4b7e852267c8fc31084dabe8c2?context=explore

i just hope create a common build & publish script to make `cloudbuild.yaml` clean (i may change the pr title)
and it could be useful to add more arch support.

but i have a bad network issue to google cloud. Github workflow just for debug (removed when the scripts ready)

Follow your suggestion, new script could be used like.
```shell
export USERS="root nonroot"
export ARCHS="amd64 arm64"
export DISTROS="debian9 debian10"

export LANG=base
export TOPICS="base static debug"
./hack/bazel_run.sh

export LANG=cc
export TOPICS="cc debug"
./hack/bazel_run.sh

export LANG=nodejs
export TOPICS="nodejs10 nodejs10_debug nodejs12_debug"
./hack/bazel_run.sh
```
logs https://github.com/morlay/distroless/runs/1251417776

new image tag rule could be change later. 
just hope keep current rule for now.

chanseokoh:
FYI, the plan is to have `cloudbuild.yaml` just run `bazel run //:publish`.

chanseokoh:
https://github.com/GoogleContainerTools/distroless/pull/624

morlay:
Cool. 

is bazel support create oci image too？

We could using `cc-debian:latest-${TARGETARCH}` instead.

could commit_sha tagged images with suffix too？

like `cc-debian:${commit_sha}-amd64`，`cc-debian:${commit_sha}-nonroot-amd64`

chanseokoh:
Bazel supports pushing OCI images. (OCI images are almost identical to Docker actually. The difference is just setting a different `mediaType` in JSON files.)

> could commit_sha tagged images with suffix too？

Technically possible, but we don't have a plan to have such suffixes. I actually think publishing images tagged with some `${commit_sha}` is unnecessary.

morlay:
if bazel support oci，
hope //publish could publish oci images.

commit_sha images is a fallback when oci images missing

downstream like istio 
not using the latest tag，but the image sha.

if oci images not published at same time,
i think they will not accept a pr with `latest`

chanseokoh:
Bazel supports publishing OCI images, as I said.

morlay:
I got that info.

Just notice `//:publish` on https://github.com/GoogleContainerTools/distroless/pull/624 have not support that.

chanseokoh:
Ah, I see what you were saying, sorry. Just notice that Distroless has always been publishing Docker images, not OCI. I'm not sure if we will change that going forward.

morlay:
seems OCI images have to be created by `docker manifest create & push`

may consider publishing commit_sha with suffix images too. 
it will be helpful for version lock usage. (before OCI images ready.) 

``` Dockerfile
# https://github.com/istio/istio/blob/master/docker/Dockerfile.distroless#L1
FROM gcr.io/distroless/static-debian10@sha256:4433370ec2b3b97b338674b4de5ffaef8ce5a38d1c9c0cb82403304b8718cde9 as distroless_source
```
to 
``` Dockerfile
FROM gcr.io/distroless/static-debian10:${COMMIT_SHA}-${TARGETARCH} as distroless_source
```

PR #618, opened 2020-10-13
Revert "Build s390x distroless base and static images"

chanseokoh:
Reverts GoogleContainerTools/distroless#612 to unblock image publishing.


loosebazooka (approved):

knm3000:
The timeout problem is caused by cloudbuild.yaml changes. Can you please revert cloudbuild.yaml changes only and keep the others?

knm3000:
I have fixed the error, please see https://github.com/GoogleContainerTools/distroless/pull/619
Is it possible to postpone the revert and check if the fix works?

PR #619, opened 2020-10-14
Fixed 4000-character limit error

knm3000:
This should fix `Your build failed to run: generic::invalid_argument: invalid build: invalid .steps field: build step 1 arg 1 too long (max: 4000)` error from https://github.com/GoogleContainerTools/distroless/pull/612.
In this case no revert changes https://github.com/GoogleContainerTools/distroless/pull/618 will be required.


chanseokoh (approved):
Thanks for the fix. Let me try this.

Issue #615, opened 2020-10-12
Add a nonroot Java image


pierDipi:
Having a nonroot Java image would be great. 
I can help if needed.

chanseokoh:
Makes sense. Probably just a matter of creating new rules using some `:nonroot` base image. I can review PRs.

PR #617, opened 2020-10-13
Add non-root Java images

pierDipi:
Add non-root Java images

Fixes #615

PR #620, opened 2020-10-15
Publish non-root Java images

pierDipi:
Images haven't been published since the cloudbuild file doesn't reference them in the images section.

Fixes https://github.com/GoogleContainerTools/distroless/issues/615
Context https://github.com/GoogleContainerTools/distroless/pull/617

cc @chanseokoh


chanseokoh:
Ah, that's what we've missed. However, there's another problem. GCB failed to run, as a step arg became too long.

> Your build failed to run: generic::invalid_argument: invalid build: invalid .steps field: build step 2 arg 1 too long (max: 4000)

Cloud you also break down the step? Like this recent #619.

pierDipi:
Sure!

chanseokoh (approved):
Thanks. The images should be published within an hour (typically much earlier than that) if nothing goes wrong again.

chanseokoh:
All good now. The images should have been published.

PR #622, opened 2020-10-19
Build powerpc (ppc64le) distroless base and static images

dims:
To support k8s use case

Signed-off-by: Davanum Srinivas <davanum@gmail.com>


WORKSPACE
```diff
@@ -288,7 +288,6 @@ load(
             "liblzma5",
         ] + (["libunwind8"] if arch in BASE_ARCHITECTURES else []),
         sources = [
-            "@" + arch + "_debian10_security//file:Packages.json",
```
dims, line 291: FIXME!

package_manager/dpkg_parser.py
```diff
@@ -70,6 +70,8 @@
 def main():
     """ A tool for downloading debian packages and package metadata """
     args = parser.parse_args()
+    if args.arch == "ppc64le":
```
dims, line 73: Separate this into another commit!

dims:
@chanseokoh Can you please take a look?

mattmoor (approved):

mattmoor:
This still LGTM.  Can you please confirm things work with a GCB dry run?

checksums.bzl
```diff
@@ -3,54 +3,66 @@
 # TO GENERATE THIS RUN: ./updateWorkspaceSnapshots.sh
 
 BASE_ARCHITECTURES = ["amd64", "arm64"]
-ARCHITECTURES = BASE_ARCHITECTURES + ["s390x"]
+ARCHITECTURES = BASE_ARCHITECTURES + ["s390x", "ppc64le"]
```
chanseokoh, line 6: Why can't this be in `BASE_ARCHITECTURES`?

dims: @chanseokoh i was just following the pattern set by the s390x :) we can revisit this in a subsequent PR.

dims:
@mattmoor thanks for the tip on running GCB by hand to verify stuff trying it now - `gcloud builds submit --substitutions=COMMIT_SHA=fake`

checksums.bzl
mattmoor: I think that like `s390x` this is a starting point to get base & static in place?

chanseokoh: I believe the only reason that we distinguish `BASE_ARCHITECTURES` and `ARCHITECTURES` is to include or exclude `libunwind8` because Debian doesn't have the library package on s390x.

So, the consequence of this would be that, Distroless won't be able to build ppc64le dotnet images (dotnet requires `libunwind8`, as it seems).

chanseokoh: Not that this is a major issue. We just need to be clear on why we have both `BASE_ARCHITECTURES` and `ARCHITECTURES` for what reason.

mattmoor: Yeah, but `s390x` doesn't publish python, node, or other images yet, so practically I think the "base" signifies what we publish `//base` for, but this may be back-filling the naming a bit 😅

chanseokoh (approved):

checksums.bzl
chanseokoh: Yeah, but I do think `ppc64le` (basically any arch that doesn't need special handling such as x390x) should be in `BASE_ARCHITECTURES`. I can tolerate a brief transition state if this is to be fixed in a subsequent PR.

dims:
```
[dims@dims-a02 14:21] ~/junk ⟩ crane manifest gcr.io/kubernetes-development-244305/static:nonroot
{
   "schemaVersion": 2,
   "mediaType": "application/vnd.docker.distribution.manifest.list.v2+json",
   "manifests": [
      {
         "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
         "size": 526,
         "digest": "sha256:f602a35415ff9e447e3f2a4634e6cece03c93c3939779453f01b811e8683a0b9",
         "platform": {
            "architecture": "amd64",
            "os": "linux"
         }
      },
      {
         "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
         "size": 526,
         "digest": "sha256:9977744d96d4ab58b6bc888d97d185f96007eca43b4d9dca36ac4db15917787a",
         "platform": {
            "architecture": "arm64",
            "os": "linux"
         }
      },
      {
         "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
         "size": 526,
         "digest": "sha256:5957cad095455053c34c9e5531f41e0032d5a4d25168a7f3880a2671aeeb2c18",
         "platform": {
            "architecture": "ppc64le",
            "os": "linux"
         }
      },
      {
         "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
         "size": 526,
         "digest": "sha256:71f90c8d46025d2b6d3d2f67e57dcd0061dd770b9110d9f55dc0d3d38f3007d8",
         "platform": {
            "architecture": "s390x",
            "os": "linux"
         }
      }
   ]
```

dims:
@mattmoor  got a clean run on GCB. This is now ready to merge.

```
---------------------------------------------------------------------------------------------------------------------------------------------------------------

ID                                    CREATE_TIME                DURATION  SOURCE                                                                                                       IMAGES  STATUS
ea8996d8-5a71-44d2-aded-8dbba248d830  2020-10-20T17:40:59+00:00  35M38S    gs://kubernetes-development-244305_cloudbuild/source/1603215658.303978-e8732b4d6ade479b8df52633df4dcc0f.tgz  -       SUCCESS
```

chanseokoh:
Thanks for checking. Merging it now.

bahetiamit:
I was trying the base image on power ppc64le arch machine and it fails with exec error as show below

[ ~]# uname -m
ppc64le
[ ~]# docker run --rm gcr.io/distroless/base@sha256:92720b2305d7315b5426aec19f8651e9e04222991f877cae71f40b3141d2f07e
standard_init_linux.go:190: exec user process caused "exec format error"

The busybox binary that is getting used is from https://busybox.net/downloads/binaries/1.31.0-defconfig-multiarch-musl/busybox-powerpc64 which is for big endian power arch and not little endian - https://bugs.busybox.net/show_bug.cgi?id=13491

Further, it seems that debian repo has the required power little endian arch binary & that seems to be working fine - https://travis-ci.com/github/Siddhesh-Ghadi/busybox-binary/builds/219571877#L175. 

So for this image to work properly on ppc64le the binary should be installed from debian repos.

chanseokoh:
@bahetiamit I have no experience with powerpc, so bear with me. Why should distroless target big-endian arch and not little-endian? Is little-endian defunct, and no one is using the CPU?

bahetiamit:
@chanseokoh Will it make more sense to file an issue for better tracking of this issue?

PR #556, opened 2020-07-28
Add support for Node v12 and v14

michael-gillett:
# Overview
This PR introduces Node v12 and v14 distroless images. I read through the thread at #442 and believe this meets all of the [criteria outlined](https://github.com/GoogleContainerTools/distroless/pull/442#issuecomment-613402437).

Note: For backwards compatibility I've continued to use Node v10 for the `latest` and `debug` tags. I imagine once it reaches end of life we can switch to have `latest` be the current Node LTS version.

Shout out to @craigthompsonhive @HofmannZ for getting the ball rolling on this

# Testing
I generated the Node v14 image locally using Bazel, dropped it into one of my projects that was using the Node v10 distroless image, and everything ran as expected

# Future Work
Setting up automatic minor and patch version bumps using GitHub actions

PR #623, opened 2020-10-20
Split cloudbuild.yaml scripts out into files

ImJasonH:
This avoids the annoying GCB arg length limit, and results in all Bazel
steps being run before all Docker steps.


chanseokoh:
For consistency, can you rework `java/cloudbuild.yaml` too? Note GCB has other limits (e.g., [max number of images](https://github.com/GoogleContainerTools/distroless/pull/556#issuecomment-665330724)). They increased the limit on the number of images to 200, but we've also added a lot of images. So I think it's safe to keep them separate, as we may be approaching the 200 limit.

chanseokoh:
Ah, forget about the max number of images. I forgot that the root `cloudbuild.yaml` pushes images directly instead of letting GCB do so. (OTOH, `java/cloudbuild.yaml` still relies on GCB to push.) So I guess we can merge them into a single build?

ImJasonH:
Added java/cloudbuild.yaml behavior

chanseokoh (approved):

chanseokoh:
Let's try this out.

PR #624, opened 2020-10-20
Use bazel to push all images in cloudbuild.yaml

ImJasonH:
Also update our rules_docker dependency to the latest release

This moves us toward a single source of truth for the list of all images, and when Bazel supports manifest lists, we can replace cloudbuild_docker.sh and push all of those with a single Bazel invocation too.

The list in BUILD was generated based on the `docker tag` invocations previously in cloudbuild_docker.sh, so it should be up to date.


BUILD
```diff
@@ -5,54 +5,160 @@ load("@io_bazel_rules_docker//container:container.bzl", "container_bundle")
 container_bundle(
     name = "all",
     images = {
-        "gcr.io/{PROJECT_ID}/base:latest": "//base:base_root_amd64_debian9",
-        "gcr.io/{PROJECT_ID}/base:nonroot": "//base:base_nonroot_amd64_debian9",
-        "gcr.io/{PROJECT_ID}/base:debug": "//base:debug_root_amd64_debian9",
-        "gcr.io/{PROJECT_ID}/base:debug-nonroot": "//base:debug_nonroot_amd64_debian9",
+        "gcr.io/{PROJECT_ID}/static:{COMMIT_SHA}": "//base:static_root_amd64_debian9",
```
mattmoor, line 8: I think that the vast majority of this could be computed using a "dictionary comprehension", see: https://docs.bazel.build/versions/master/skylark/lib/dict.html

That would seriously boost my confidence that we've covered everything (and would probably fill in some gaps we've had previously).

I think that the main pieces it wouldn't "just work" for would be the aliases we submit as sugar, but I think enumerating those explicitly would be very nice.

mattmoor: All told, it will probably also be shorter!

ImJasonH: Yeah I thought about going that route in this PR but I thought it might make a better follow-up since we were already kind of unsure about the list being out of sync. Seeing the whole list explicitly spelled out is easier to compare to the previous list.

Let me know if you'd prefer this to be in this PR and I can hack on it, otherwise it'll be a follow-up.

ImJasonH: Yeah, after hacking on this for a few minutes I'm not sure it's cleaner to use dict comprehensions, versus just spelling it all out. For one, it [doesn't seem possible to add two dicts](https://github.com/bazelbuild/skydoc/issues/10), and [`update` returns `None`](https://docs.bazel.build/versions/master/skylark/lib/dict.html#update). 😖 

Maybe I'm just not good enough at Skylark to make it happen. If people have better luck making this expressive I'll gladly defer to them. I agree it probably would help make sure we cover (and keep covering!) all the combinations. In the meantime I think this is an improvement we can build on in future changes.

chanseokoh:
The Travis ARM64 build is failing (consistently I think). Looks like an mismatched architecture issue. Not sure what's causing this.

ImJasonH:
The arm failure seems to be due to me actually _rolling back_ the `rules_docker` dep (I didn't realize it was pulling from an unreleased version of rules_docker...)

Without that, I'm hitting https://github.com/bazelbuild/rules_docker/issues/525 caused by too many targets being pushed by push-all. I'll investigate whether I can fix that in rules_docker...

ImJasonH:
Confirmed that the build passes with https://github.com/bazelbuild/rules_docker/pull/1658, and only takes a few minutes longer to push all 178 images in sequence.

ImJasonH:
This should be ready for a look now, https://github.com/bazelbuild/rules_docker/pull/1658 is merged and we now use the sequential push option to avoid instability around pushing so many images.

cloudbuild_bazel.sh
```diff
@@ -4,162 +4,4 @@ set -o errexit
...
-    bazel run --host_force_python=PY2 //java:java11_debug_root_debian10
-    bazel run --host_force_python=PY2 //java:java11_debug_nonroot_debian10
-
-    bazel run --host_force_python=PY2 //java/jetty:jetty_java8_debian9
-    bazel run --host_force_python=PY2 //java/jetty:jetty_java8_debug_debian9
-    bazel run --host_force_python=PY2 //java/jetty:jetty_java11_debian9
-    bazel run --host_force_python=PY2 //java/jetty:jetty_java11_debug_debian9
-    bazel run --host_force_python=PY2 //java/jetty:jetty_java11_debian10
-    bazel run --host_force_python=PY2 //java/jetty:jetty_java11_debug_debian10
+bazel run --host_force_python=PY2 //:publish
```
chanseokoh, line 7: Do need this script? What is we embed these two commands into `cloudbuild.yaml`? Or, was it that you will eventually remove `cloudbuild.yaml`? In any case, rename the file to `build_and_publish.sh`? And is it safe to run `cloudbuild_bazel.sh` locally? (I guess it will fail because `{PROJECT_ID}` is undefined?) Maybe now it's worth adding some comments into this script?

chanseokoh (commented):
Cool! Thanks for cleaning this up.

cloudbuild_bazel.sh
ImJasonH: Nope, we could just move cloudbuild_bazel.sh into cloudbuild.yaml now. 👍 

When/if rules_docker supports manifest lists we can assemble and publish those using Bazel too and remove that script 🙏

chanseokoh (approved):
LGTM. I've programmatically verified that all the images are correctly covered. I'll merge this soon.

PR #630, opened 2020-11-06
Publish base & cc with disto suffix

morlay:
Publishing multi arch images of base & cc with `*-debian9` `*-debian10` (have to re order images in `BUILD`, `{COMMIT_SHA}` as group splitter)

Downstream need `*-debian10`
https://github.com/istio/istio/issues/26652#issuecomment-714858976


cloudbuild_docker.sh
```diff
@@ -12,58 +12,30 @@ set -o xtrace
...
-docker manifest push gcr.io/$PROJECT_ID/cc:latest
-
-docker manifest create gcr.io/$PROJECT_ID/cc:debug \
-   gcr.io/$PROJECT_ID/cc:debug-amd64 \
-   gcr.io/$PROJECT_ID/cc:debug-arm64
-docker manifest push gcr.io/$PROJECT_ID/cc:debug
-
+docker_manifest() {
+  image=$1
+  from_images=""
```
chanseokoh, line 17:
```suggestion
  local image=$1
  local from_images=""
```

cloudbuild_docker.sh
```diff
@@ -12,58 +12,30 @@ set -o xtrace
...
+
+  for arch in $ARCHS; do
```
chanseokoh, line 19: Why is it that you pass "image" as a function arg while "arch" as a variable? I'd go for function args, e.g.,
```
docker_manifest gcr.io/$PROJECT_ID/static${distro_suffix}:nonroot "amd64 arm64 s390x ppc64le"
```

chanseokoh (commented):
Awesome. Thanks for you contribution!

cloudbuild_docker.sh
morlay: my IDE warn me drop the `local` https://github.com/koalaman/shellcheck/wiki/SC2039

chanseokoh: I wouldn't bother, but if you want, change the shebang to `#!/bin/bash`.

cloudbuild_docker.sh
morlay: Sure. I updates it.

cloudbuild_docker.sh
morlay: Sure. I updated both.

chanseokoh: Ah, sorry. The `docker` image doesn't have `/bin/bash`. Could you revert the shebang? And let's add the `_` prefix to these variables (without `local`).

```sh
   _image=$1
   _archs=$2
   ...
```
That'll reduce the risk of variable conflicting at the global level.

chanseokoh: @morlay still pending

chanseokoh: Never mind. I fixed it.

chanseokoh:
Oops, sorry. I broke something.

chanseokoh:
Couldn't restore the PR, so I put to #631. Sorry for the trouble. If you want to restore this PR anyway, you can go ahead, and I will review it again.

Issue #606, opened 2020-10-01
multi-arch support for Java/Jetty


mattmoor:
Following in the style of https://github.com/GoogleContainerTools/distroless/pull/605 and its predecessors.

cc @loosebazooka @chanseokoh

cablespaghetti:
Watching with interest.  :+1: 

If I get time I might have a go given there is already an example to take inspiration from. Unless somebody more qualified is already working on this?

reppners:
I'd also like to see this happen as I'm trying to have a distroless java with buildx to have support for Dockerfile statements like `FROM --platform=$BUILDPLATFORM gcr.io/distroless/java11` and then build with ` docker buildx build . -t my-distroless-java-app`.

Is there any timeline or way to support this?

PR #670, opened 2021-01-08
Make Java/Jetty multi-arch (amd64, arm64).

digitalpardoe:
The project I'm working on is starting to take a closer look at ARM servers (and beginning to use ARM development machines) but we'd like to stick with distroless where possible and we were missing Java images.

For our purposes we only really needed the base image but given https://github.com/GoogleContainerTools/distroless/issues/606 is open it seemed worthwhile to try and migrate everything Java-based.

I've only gone for amd64 and arm64 as it's what I was able to easily test but I suspect the the other architectures may work fine and as such changed the `BUILD` files so `ARCHITECTURES` is an alias to `BASE_ARCHITECTURES` and can be easily switched out.

As part of the testing there were direct references to directories with architectures in `command` statements, I thought duplicating this might get out of hand quickly given the number of images and architectures so I added an additional alias of `/usr/lib/jvm/openjdk` into the built containers which allowed the test cases to be simplified somewhat - it will probably also help people that are building multi-arch pipelines from these images.

My general feeling is that this is a large PR and has made some of the Java-related sections a little messy - for example the `JAVA` section of the main `BUILD` file - initially this was a list of `("","")` and was more readable but `buildifier` had other opinions.

Overall it works and can probably be made neater but that may be best for one of the maintainers to take a run at as I fear further modification from me may make it worse rather than better.


digitalpardoe:
A selection of the arm64 images were manually tested on an M1 Mac and the amd64 images on an Intel Mac. I also set up the whole build to build to an internal Docker repo and can verify the multi-arch manifests for the Java images are working as expected.

digitalpardoe:
The ARM CI build seems to have failed with `Disk quota exceeded `.

digitalpardoe:
I've removed all the Debian 9 references from the Java side of things and run all the container tests in isolation, they appear to be working as expected - the other builds around that however still seem to be broken.

dlorenc:
> I've removed all the Debian 9 references from the Java side of things and run all the container tests in isolation, they appear to be working as expected - the other builds around that however still seem to be broken.

Thanks! I'll have to defer to @chanseokoh to review/merge the Java changes here.

PR #674, opened 2021-01-27
Add Erlang base image

magnetised:


magnetised:
Sorry - didn't mean to make a PR here 😳

Nezteb:
@magnetised Out of curiosity, is there a reason you didn't want to add Erlang to the official `GoogleContainerTools` repo? I was looking at adding an Erlang/Elixir image but found this first. 😄

loosebazooka:
We don't really intend of supporting the world. The build and maintenance is expensive so we try to keep our support scoped pretty tightly. Although there are offerings from chainguard or ubuntu or other vendors that might satisfy your needs.

dlorenc:
If you're interested we have erlang packaged in wolfi and would add an image for it!
https://github.com/wolfi-dev/os/blob/main/erlang.yaml

The image would go here: https://github.com/chainguard-images/images/tree/main/images

Nezteb:
@loosebazooka Totally understandable; I figured I'd ask. 😄 

@dlorenc Neat! Thanks for the info! 🥳

PR #689, opened 2021-03-03
Disable parallel runs when building package bundles when updating snapshots

chanseokoh:
The daily GitHub Actions workflow to upgrade packages started to fail at some point recently. Disabling parallel runs as we did for Cloud Build:

https://github.com/GoogleContainerTools/distroless/blob/1e4a8bb3ad03f71b572cbcb3bbc25f3fd8d0ff14/cloudbuild.yaml#L20-L33


mpeddada1 (approved):

chanseokoh:
Hmm... this might have helped to some degree, but unfortunately, the workflow keeps failing. :(

PR #699, opened 2021-03-24
Bumping packages to latest stable versions

priyawadhwa:


dlorenc:
Thanks @priyawadhwa ! The cron appears to be broken :(

dlorenc:
This looks like the same issue causing the action to fail :( Let's see if it works in GCB. We have all 3 CI systems!

chanseokoh:
FYI, when they fail, it's always due to a network-related issue when downloading .deb files from snapshots.debian.org. I believe the repo employs some kind of DDoS prevention.

- [GitHub Actions daily update-snapshots workflow](https://github.com/GoogleContainerTools/distroless/actions/workflows/update-deb-package-snapshots.yml): this had been very very stable until Mar 3rd. But since then, it never passes. I suspect some kind of DDoS prevention or rate limiting started to kick in from snapshots.debian.org against GitHub Actions?
- Travis: this has been OK, except the ARM build. The AMD build usually passes after restarting at most a few times when it fails. OTOH, the ARM build almost all the time fails.
- GCB image publication pipeline: this is pretty stable.

Issue #719, opened 2021-04-10
The example Dockerfile uses ADD, and COPY is appropriate.


bootjp:
The official Docker documentation states.

> Although ADD and COPY are functionally similar, generally speaking, COPY is preferred. That’s because it’s more transparent than ADD.
https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#add-or-copy

Therefore, we propose a change as follows.
Can you accept this pull request?
https://github.com/bootjp/distroless/commit/f7ad33ad5b1084bc557f5a28c488fe26e0316d34

PR #726, opened 2021-04-24
It is appropriate to use COPY in this case.

bootjp:
#719


loosebazooka:
uh sorry no one looked at this, can you update it and we'll merge it?

bootjp:
@loosebazooka conflict resolved.
I would appreciate it if you could merge them.

PR #762, opened 2021-07-16
Use python3 for builds

loosebazooka:


test.sh
```diff
@@ -18,20 +18,29 @@ set -o errexit
 set -o xtrace
 
 # Linting
-buildifier -mode=fix $(find . -name 'BUILD*' -o -name 'WORKSPACE*' -o -name '*.bzl' -type f)
-find . -name "*.py" | xargs pylint --disable=R,C
+#buildifier -mode=fix $(find . -name 'BUILD*' -o -name 'WORKSPACE*' -o -name '*.bzl' -type f)
```
loosebazooka, line 21: This was failing for some reason, I don't know why? Or it passed but had error code 123

loosebazooka: never mind, I just fixed the error.

dlorenc (approved):

PR #769, opened 2021-07-25
Add size comparisons to README.md

imjasonh:
Compare distroless/static sizes to alpine and debian, as well as a moderately sized jpeg, and https://github.com/GoogleContainerTools/distroless.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,6 +13,14 @@ Restricting what's in your runtime container to precisely what's necessary for y
 and other tech giants that have used containers in production for many years.
 It improves the signal to noise of scanners (e.g. CVE) and reduces the burden of establishing provenance to just what you need.
 
+Distroless images are _very small_.
+The smallest distroless image, `gcr.io/distroless/static`, is around 650 kB.
+That's about 25% of the size of `alpine` (~2.5 MB), and less than 1.5% of the size of `debian` (50 MB).
+
+For example, `gcr.io/distroless/static` is a container image that's much smaller than [this image of a shipping container](https://unsplash.com/photos/bukjsECgmeU).
+It's about 1/3rd the size of all the resources on [this page you're reading right now](https://github.com/GoogleContainerTools/distroless).
+It's very small.
+
 ## How do I use distroless images?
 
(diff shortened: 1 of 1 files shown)


dlorenc (approved):

README.md
```diff
@@ -13,6 +13,14 @@ Restricting what's in your runtime container to precisely what's necessary for y
 and other tech giants that have used containers in production for many years.
 It improves the signal to noise of scanners (e.g. CVE) and reduces the burden of establishing provenance to just what you need.
 
+Distroless images are _very small_.
```
chanseokoh, line 16: This only considers non-language image. The Distroelss Java 11 is 200MB, which is not much different from other images. (The Java 8 image, which is defunct, is of size 130MB while `openjdk:8-jre-alpine1 is 84.9MB.)

imjasonh: It's a fair point. Is it maybe better to say "Distroless images *aim to be* very small"?

It looks like gcr.io/distroless/java-debian10:11 is quite a bit smaller than openjdk:11 though (71 MB vs 314 MB), so I think we could still make a blanket statement that distroless images are smaller than their distroful counterparts.

chanseokoh: `openjdk:11` is a full JDK. It's not a fair comparison.
```
gcr.io/distroless/java-debian10   11             8775134e5ec2   51 years ago    200MB
adoptopenjdk                      11-jre         ca04986fae0a   12 days ago     244MB
openjdk                           11             e782567c0965   4 days ago      648MB
```

chanseokoh: And alpine is much smaller.
```
adoptopenjdk/openjdk11            alpine-jre              7af79a5acb07   22 hours ago    148MB
```

imjasonh: Fair enough. Is there some way we can document the relative smallness of distroless images (even though there might exist smaller alternatives) clearly? If you think the comparison isn't worth mentioning that's fine too, mainly I don't want to have the README take a long tangent covering the relative sizes of all possible alternatives. I just wanted to point out that distroless/static is smaller than the GitHub repo's front page 😄

chanseokoh: I don't have a strong opinion, so don't block on me. I'm not a maintainer after all. Do what you think is appropriate.

PR #773, opened 2021-07-26
remove binary build artifact

loosebazooka:

#771 

~~I can't imagine this was intentionally added in.~~ lol it was my bad


loosebazooka:
this is a little more involved than I thought

chanseokoh:
I'm curious if tar files count as binary artifacts. 
```
$ find . -name '*.tar'
./base/tmp.tar
./base/nsswitch.tar
```

loosebazooka:
I'm not sure, @jeffmendoza? are tars considered binaries?

jeffmendoza:
It's only `examples/cc/bin/hello` here I believe.

dlorenc:
I think they should be considered binary artifacts.

jeffmendoza:
https://github.com/ossf/scorecard/blob/main/checks/binary_artifact.go#L50 fyi

jeffmendoza:
https://github.com/ossf/scorecard/issues/759

dlorenc (approved):

PR #777, opened 2021-07-28
Remove all debian9 references

loosebazooka:
Cleaning up old debian9 references


experimental/python2.7/generate_ldconfig_cache.py
```diff
@@ -1,55 +0,0 @@
-#!/usr/bin/env python
```
loosebazooka, line 1: @evanj not sure exactly what this did. Looks like you touched it last? Was this only required for debian9?

dlorenc (approved):

experimental/python2.7/generate_ldconfig_cache.py
evanj: Hmm... If I remember, this predates Debian10, so I'm not actually sure. The change that introduced this was:

https://github.com/GoogleContainerTools/distroless/pull/228

This was necessary to make the Python2.7 standard library's `ctypes.util.find_library` function work to find `c` and `rt` libraries, which are loaded by gunicorn and monotonic (and maybe other popular third-party Python libraries). It is possible that Python3 may not need this any more?

I added a test when I added this. If this still works, then this is no longer needed:

```
  # libraries like gunicorn and monotonic call glibc functions using ctypes
  - name: use_ctypes_glibc
    command: ["/usr/bin/python3", "-c", "import ctypes.util; ctypes.CDLL(ctypes.util.find_library('rt')).timer_create"]
```

PR #781, opened 2021-08-06
Add support for Packages.xz

loosebazooka:
Add support for Packages.xz

- Also update workspace to use Packages.xz instead of Packages.gz
- Users may chose to still use Packages.gz if they want if expliclty specify packages_url
- this is a breaking change to dpkg_src rule
  - rename input: packages_gz_url -> packages_url
  - when using snapshots will only look for Packages.xz, the sha256 must reference
    a Packages.xz, it will not use the Packages.gz.


loosebazooka:
There is one issue that this mandates the `dpkg_src` rule when using `snapshots` will have to use `Packages.xz`, so all sha256s must be for `Packages.xz`, not `Packages.gz`

dlorenc (approved):

PR #791, opened 2021-08-18
package_manager dpkg_parser: Use /usr/bin/env python3

evanj:
The dpkg_parser.py file imports the lzma system package, which only
exists in Python 3. On my system (Debian 10), the default "python"
is still Python 2. As a result, the built dpkg_parser.par does not
execute. Update the interpreter to use "python3" explicitly.

The python_version attribute does nothing at the moment, but it seems
to me that is likely a bug in the subpar bazel rules. Let's set it
since that seems to be the documented way to change the Python
version for a py_binary.

### DIFF
--- a/package_manager/BUILD
+++ b/package_manager/BUILD
@@ -1,5 +1,6 @@
 load("@subpar//:subpar.bzl", "par_binary")
 
+# Requires python3 for lzma system library
 par_binary(
     name = "dpkg_parser",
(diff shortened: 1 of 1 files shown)


loosebazooka:
oh this is convenient. I was going around forcing python to point to python3 in our builds.

loosebazooka (approved):

PR #797, opened 2021-08-24
Update update-deb-package-snapshots.yml

loosebazooka:
Add a comment so I can get failure messages

I need to be the one merging too I think


dlorenc (approved):
I approve this message

PR #806, opened 2021-08-29
Add origin information for filtered Packages.xz

loosebazooka:
Packages.xz.orig: the full original Packages.xz file from debian sources
InRelease.orig: the InRelease file with checksums and GPG signature inline


dlorenc:
Lint error, otherwise lgtm

dlorenc (approved):

PR #813, opened 2021-09-02
Make registry configurable with env var

priyawadhwa:
This way we can use an insecure registry for getting digests when generating provenance


loosebazooka:
Can bazel do default values instead or us injecting the default everywhere?

dlorenc (approved):

PR #814, opened 2021-09-03
Add Tekton task for generating provenance 

priyawadhwa:
Adds a task which:
1. pulls in the distroless repo at the specified commit
2. builds the images and pushes to an insecure registry to the cluster
3. gets the digests for six distroless images, and saves them in the IMAGES result

This six images will be picked up by Tekton Chains to generate provenance

If this works, we can definitely add in more images. we can probably have up to 30 per Task before we hit the maximum size for Tekton results.

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -60,3 +60,22 @@ steps:
   entrypoint: sh
   args:
   - ./cloudbuild_cosign.sh 
+
+# Start task in k8s to generate provenance
+- name: gcr.io/cloud-builders/gcloud
+  args:
+  - container
+  - clusters
+  - get-credentials
+  - provenance
+  - --zone=us-central1-c
+  - --project=distroless
+
+- name: gcr.io/tekton-releases/dogfooding/tkn@sha256:f69a02ef099d8915e9e4ea1b74e43b7a9309fc97cf23cb457ebf191e73491677
+  args:
+  - task
+  - start
+  - -f=provenance/provenance-task.yaml
+  - --param
+  - CHAINS-GIT_COMMIT=${COMMIT_SHA}
+  - --use-param-defaults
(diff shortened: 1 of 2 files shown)


dlorenc (approved):

cloudbuild.yaml
```diff
@@ -60,3 +60,23 @@ steps:
...
+- name: gcr.io/cloud-builders/gcloud
+  entrypoint: sh
+  args:
+  - -c
+  - |
+    #!/bin/sh
+    set -o errexit
+    set -o xtrace
+
+    gcloud container clusters get-credentials provenance --zone=us-central1-c --project=distroless
```
mattmoor, line 74: Use `${PROJECT_ID}` here?

cloudbuild.yaml
```diff
@@ -60,3 +60,23 @@ steps:
...
+    
+    # Install tkn
+    curl -Lo tkn_0.20.0_Linux_arm64.tar.gz  https://github.com/tektoncd/cli/releases/download/v0.20.0/tkn_0.20.0_Linux_x86_64.tar.gz
+    tar -xvzf tkn_0.20.0_Linux_arm64.tar.gz -C /usr/local/bin/ tkn
```
mattmoor, line 78: Hoist a version variable for 0.20.0 to make bumping this easier?

provenance/provenance-task.yaml
```diff
@@ -0,0 +1,81 @@
...
+        value: insecure-registry.default.svc.cluster.local:5000
+      - name: COMMIT_SHA
+        value: $(params.CHAINS-GIT_COMMIT)
+      script: |
+        #!/bin/sh
+        set -ex
+
+        git clone $(params.CHAINS-GIT_URL)
+        cd distroless
+        git checkout $(params.CHAINS-GIT_COMMIT)
```
mattmoor, line 37: It might be good to make this another step?

priyawadhwa: agreed! i'm going to convert this to a pipeline in another PR, so i'll add it in there.

Issue #729, opened 2021-04-28
distroless/java missing libharfbuzz.so.0


dawwe:
Description of the issue: The docker images build ok with Jib, but I get an exception thrown when the container runs. The application is a spring boot 2.4.x app that runs ok in the IDE against java 11.

Started with the java:11 image a couple of days ago, older images works. 

Expected behavior: No exception

Steps to reproduce: Jib a spring boot 2.4.x app that uses apache poi and creates Excelfiles. 

Environment: Mac  / Java 11 / Maven / Spring boot

Log output: 

`Caused by: java.lang.UnsatisfiedLinkError: /usr/lib/jvm/java-11-openjdk-amd64/lib/libfontmanager.so: libharfbuzz.so.0: cannot open shared object file: No such file or directory
        at java.base/java.lang.ClassLoader$NativeLibrary.load0(Native Method)
        at java.base/java.lang.ClassLoader$NativeLibrary.load(ClassLoader.java:2442)
        at java.base/java.lang.ClassLoader$NativeLibrary.loadLibrary(ClassLoader.java:2498)
        at java.base/java.lang.ClassLoader.loadLibrary0(ClassLoader.java:2694)
        at java.base/java.lang.ClassLoader.loadLibrary(ClassLoader.java:2648)
        at java.base/java.lang.Runtime.loadLibrary0(Runtime.java:830)
        at java.base/java.lang.System.loadLibrary(System.java:1873)
        at java.desktop/sun.font.FontManagerNativeLibrary$1.run(FontManagerNativeLibrary.java:57)
        at java.base/java.security.AccessController.doPrivileged(Native Method)
        at java.desktop/sun.font.FontManagerNativeLibrary.<clinit>(FontManagerNativeLibrary.java:32)
        at java.desktop/sun.font.SunFontManager$1.run(SunFontManager.java:279)
        at java.base/java.security.AccessController.doPrivileged(Native Method)
        at java.desktop/sun.font.SunFontManager.<clinit>(SunFontManager.java:275)
        at java.base/java.lang.Class.forName0(Native Method)
        at java.base/java.lang.Class.forName(Class.java:398)
        at java.desktop/sun.font.FontManagerFactory$1.run(FontManagerFactory.java:82)
        at java.base/java.security.AccessController.doPrivileged(Native Method)
        at java.desktop/sun.font.FontManagerFactory.getInstance(FontManagerFactory.java:74)
[...]

PR #818, opened 2021-09-10
Add libharfbuzz0b to java image

peerapach:
I'm experiencing the issue #729 . 

This change will add the libharfbuzz0b and dependencies related to Debian 10-based Java image.

I've tested this locally and it seems to have fixed the issue.


peerapach:
@googlebot I signed it!

peerapach:
@googlebot I fixed it

PR #823, opened 2021-09-12
Refactor the cosign script.

mattmoor:
This refactors the cosign script for distroless, so that the arguments are passed through from `cloudbuild.yaml`, and shifts where `-key ...` is passed to be in `cloudbuild.yaml`.

This is preparation to start *also* signing the distroless images with a "keyless" flow, once https://github.com/sigstore/cosign/pull/652 lands and cosign 1.2 cuts.

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -57,6 +57,8 @@ steps:
   - PROJECT_ID=${PROJECT_ID}
   - COMMIT_SHA=${COMMIT_SHA}
   - REGISTRY=gcr.io
+  - KMS_VAL=gcpkms://projects/${PROJECT_ID}/locations/global/keyRings/cosign/cryptoKeys/cosign
   entrypoint: sh
   args:
-  - ./cloudbuild_cosign.sh 
+  - -c
+  - ./cloudbuild_cosign.sh -key $KMS_VAL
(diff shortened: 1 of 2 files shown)


mattmoor:
Basically what I'm thinking is we'll just add another block in `cloudbuild.yaml` with:

```yaml
- name: gcr.io/projectsigstore/cosign:v1.1.0@sha256:47bdf660ceadb5065aa05af62fd24b184db0fdcb34d5892247035dd0dd2d5c3f
  env:
  - PROJECT_ID=${PROJECT_ID}
  - COMMIT_SHA=${COMMIT_SHA}
  - REGISTRY=gcr.io
  - COSIGN_EXPERIMENTAL=true
  - GOOGLE_SERVICE_ACCOUNT_NAME=<THE SA WE WANT THINGS SIGNED WITH>
  entrypoint: sh
  args:
  - -c
  - ./cloudbuild_cosign.sh
```

mattmoor:
Note to self (or whoever does it!) we should make sure the service account we choose uses `${PROJECT_ID}` so that others can replicate builds with `gcloud builds submit`:

```yaml
- GOOGLE_SERVICE_ACCOUNT_NAME=<THE SA WE WANT THINGS SIGNED WITH>
```

dlorenc (approved):

PR #824, opened 2021-09-12
Bump cosign to the latest CI image.

dlorenc:
The bump to 1.1 accidentally picked up a change that removed a shell.
This image puts the shell back, fixing the signing here.

Signed-off-by: Dan Lorenc <dlorenc@google.com>


mattmoor (approved):
We should bump this again when 1.2 lands

PR #826, opened 2021-09-13
Fix user-defined substitutions and shell for cloudbuild_cosign script

priyawadhwa:
User defined env vars have to [start with an underscore](https://cloud.google.com/build/docs/configuring-builds/substitute-variable-values) and need to be specified on the trigger itself, so I updated the trigger with the value for `_KMS_VAL` and `_TKN_VERSION`

`cloudbuild_cosign.sh` also wasn't being run because the wrong shell was specified 

Builds have been broken since #823, hopefully this should fix it (cc @mattmoor )

### DIFF
--- a/cloudbuild_cosign.sh
+++ b/cloudbuild_cosign.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/busybox/sh
 
 set -o errexit
 set -o xtrace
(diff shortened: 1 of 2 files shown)


cloudbuild_cosign.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/busybox/sh
```
mattmoor, line 1: Will this work properly on GCB?

priyawadhwa: yup, tested on my on project

it fixes this error:

```
Step #0: sh: ./cloudbuild_cosign.sh: not found
```

PR #827, opened 2021-09-13
Add the keyless signing flow with GCB

mattmoor:
@dlorenc created this serviceaccount earlier in `distroless`, so this should work (famous last words, right?)


dlorenc (approved):
Builds are working again from the last break, so let's try again!

PR #829, opened 2021-09-13
Add pipeline for generating provenance

priyawadhwa:
Replaces the task that was generating provenance for a few images with a pipeline.

The pipeline runs 8 tasks in parallel at the end, which are each responsible for generating provenance for a subset of 30 images. we need to do this because only 30 fully qualified images can fit in the results section of a TaskRun.


provenance/distroless-provenance-task.yaml
```diff
@@ -0,0 +1,31 @@
...
+      - name: REGISTRY
+        value: insecure-registry.default.svc.cluster.local:5000
+      script: |
+        #!/bin/sh
+        set -ex
+
+        go install github.com/google/go-containerregistry/cmd/crane@latest
+
+        # install jq and create the images file
+        ./cloudbuild_jq.sh
```
dlorenc, line 28: Do we need this script still? Same for crane

priyawadhwa: ah yah we don't need crane anymore, removed it

the script basically parses through the manifest and stores all of the images built in a file called `images` which we still use

priyawadhwa:
tested on my own project, this should work

dlorenc (approved):

PR #831, opened 2021-09-14
Fix readers

loosebazooka:
Handle scanner errors and use bytes instead of relying on the built in readers on byte buffer


loosebazooka:
um hang on

dlorenc (approved):

PR #847, opened 2021-09-20
Bye bye python2.7

loosebazooka:
- https://www.python.org/doc/sunset-python-2/
- We need to regenerate ld.so.arm64.cache on the right hardware.
  This should just work anyway for py3 (it was before), but the
  tests wont be happy (they are ignored now also anyway).


loosebazooka:
This mostly moves some build rules over from python27 -> python3, but also revives the generator cause I didn't know what I was doing here: https://github.com/GoogleContainerTools/distroless/pull/777#discussion_r678642035

dlorenc (approved):

PR #879, opened 2021-10-25
A rough idea on how support will look for images.

loosebazooka:
Please comment on draft with reasoning on why support should be extended/retracted

### DIFF
--- a/SUPPORT_POLICY.md
+++ b/SUPPORT_POLICY.md
@@ -1,6 +1,6 @@
-Distroless is currently tracks debian on their [standard support timeline](https://wiki.debian.org/DebianReleases#Production_Releases).
+Distroless currently tracks Debian on their [standard support timeline](https://wiki.debian.org/DebianReleases#Production_Releases).
 
-The current estimation of EOL of images with the pattern 
+The current estimation of end of life for images with the pattern 
 
 `gcr.io/distroless/<image>-debian<version>:(latest|nonroot|debug|debug-nonroot)`
 
--- a/SUPPORT_POLICY.md
+++ b/SUPPORT_POLICY.md
@@ -7,9 +7,9 @@ The current estimation of EOL of images
 | static      | standard    | aug 2022      | debian 12 + 1yr |
 | base        | standard    | aug 2022      | debian 12 + 1yr |
 | cc          | standard    | aug 2022      | debian 12 + 1yr |
-| java        | standard    | aug 2022      | debian 12 + 1yr |
-| node        | standard    | aug 2022      | debian 12 + 1yr |
-| python      | standard    | aug 2022      | debian 12 + 1yr |
+| java        | standard    | jan 2022      | debian 12 + 3mo |
+| node        | standard    | jan 2022      | debian 12 + 3mo |
+| python      | standard    | jan 2022      | debian 12 + 3mo |
 
 
 Java
--- a/SUPPORT_POLICY.md
+++ b/SUPPORT_POLICY.md
@@ -1,6 +1,6 @@
 Distroless currently tracks Debian on their [standard support timeline](https://wiki.debian.org/DebianReleases#Production_Releases).
 
-The current estimation of end of life for images with the pattern 
+The current estimation of end of life for images with the pattern:
 
 `gcr.io/distroless/<image>-debian<version>:(latest|nonroot|debug|debug-nonroot)`
 
--- a/SUPPORT_POLICY.md
+++ b/SUPPORT_POLICY.md
@@ -2,14 +2,14 @@ Distroless is designed to follow debian on their [standard support timeline](htt
 
 The current estimation of EOL of images
 
-| Image       |  Support    | Debian 10 EOL | Debian 11 EOL   |
-| ----------- | ----------- | ------------- | --------------- |
-| static      | standard    | aug 2022      | debian 12 + 1yr |
-| base        | standard    | aug 2022      | debian 12 + 1yr |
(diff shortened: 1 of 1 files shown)


SUPPORT_POLICY.md
```diff
@@ -0,0 +1,25 @@
...
+
+| Image       |  Support    | Debian 10 EOL | Debian 11 EOL   |
+| ----------- | ----------- | ------------- | --------------- |
+| static      | standard    | aug 2022      | debian 12 + 1yr |
+| base        | standard    | aug 2022      | debian 12 + 1yr |
+| cc          | standard    | aug 2022      | debian 12 + 1yr |
+| java        | standard    | jan 2022      | debian 12 + 3mo |
+| node        | standard    | jan 2022      | debian 12 + 3mo |
+| python      | standard    | jan 2022      | debian 12 + 3mo |
+
```
loosebazooka, line 13: I think the language runtimes should be deprecated a bit quicker or the build will balloon like crazy.

SUPPORT_POLICY.md
```diff
@@ -0,0 +1,25 @@
...
+
+Java
+Java will only support current LTS version distributed by debian [see here](https://wiki.debian.org/Java).
```
timja, line 16: I don't think this works very well for Java (unless you mean all the LTS versions debian is shipping).

There's currently 3 actively supported LTS lines by the main vendors (8, 11, 17).

I assume Debian fixes the LTS version of default-jdk for the lifecycle of the OS version, given it's nearly 3 months since release and its pointing to JDK 11.

 Even so I'm not sure that debian is the best vendor for Java, Java 17 release was very slow, not sure how good they are for others. Would it not be better to get them from a vendor that specialises in Java? e.g. something like Eclipse Temurin or Zulu (which are two of the recommended ones in [actions/setup-java](https://github.com/actions/setup-java#basic). People generally don't want to touch Oracle because of the licensing stuff they've done and reputation.

(this may be off topic from the support policy, fine if so)

loosebazooka: I think `java-base` should provide users with the sufficient libraries for them to bring their own jre into a container image (either from a provider like temurin/zulu or using something like jlink to create a minimal jre for the application).

timja: yeah makes sense, upstreams are pushing `jlink` currently as you can massively decrease image size with it.

SUPPORT_POLICY.md
briandealwis: Why's that?  I think it's worth adding rationale in this doc.

Also worth adding that this affects the `:debian*` tags specifically: the `:latest` tags will just continue.

And what does "standard" entail?  Does it need to be mentioned here?

loosebazooka: Yeah I can see how "standard" is a bit confusing without context. I might just remove it. It was a reference to [debian's](https://wiki.debian.org/DebianReleases#Production_Releases) provided EOL date (not LTS or ELTS)

SUPPORT_POLICY.md
```diff
@@ -0,0 +1,27 @@
+Distroless is designed to follow debian on their [standard support timeline](https://wiki.debian.org/DebianReleases#Production_Releases).
```
briandealwis, line 1: Remove ` designed `? The use of Debian is an implementation detail.  The Debian release information is surfaced in the tags, but more to prevent breaches from Hyrum’s Law. Diatribes could switch to Nix (with a lot of implementation work) and most users wouldn’t know the difference.

yangsen01:
你好，邮件已收到。谢谢。

SUPPORT_POLICY.md
```diff
@@ -0,0 +1,27 @@
...
+
+The current estimation of EOL of images
```
briandealwis, line 3: Is it worth spelling out that this is for specific tags (or patterns of tags). This table is hard to make sense of if I just use `:latest`

loosebazooka: how's this look?

SUPPORT_POLICY.md
```diff
@@ -0,0 +1,29 @@
+Distroless is currently tracks debian on their [standard support timeline](https://wiki.debian.org/DebianReleases#Production_Releases).
```
timja, line 1:
```suggestion
Distroless currently tracks debian on their [standard support timeline](https://wiki.debian.org/DebianReleases#Production_Releases).
```

SUPPORT_POLICY.md
```diff
@@ -0,0 +1,29 @@
...
+
+The current estimation of EOL of images with the pattern
```
timja, line 3:
```suggestion
The current estimation of end of life for images with the pattern:
```

SUPPORT_POLICY.md
loosebazooka: good catch, I'm going to edit these in directly, sometimes the CLA bot gets confused with multiple authors

SUPPORT_POLICY.md
jonjohnsonjr: > Diatribes

Phenomenal typo.

loosebazooka: Also Nix while probably more convenient doesn't yet have full range of architecture support.

johngmyers: The Debian package metadata is exposed in the images and is consumed by some users for dependency/vulnerability scanning.

loosebazooka: right, this will also be published in an sbom of some sort -- in the future.

dekkagaijin (approved):

PR #894, opened 2021-11-07
update to use distroless static vs base, including qualified guidance

michael-basil:
This pull request suggests the following changes:
* Update to golang:latest 
* Addition to initialize the module, needed for compilation
* Disable of cgo as the example does not require it
* Default to Distroless static vs base
* Documentation guidance on when it is necessary to utilize larger distroless/base base image

### DIFF
--- a/examples/go/Dockerfile
+++ b/examples/go/Dockerfile
@@ -1,7 +1,7 @@
 FROM golang:latest as build-env
 
 WORKDIR /go/src/app
-COPY main.go .
+COPY *.go .
 
 RUN go mod init
 RUN go get -d -v ./...
--- a/examples/go/Dockerfile
+++ b/examples/go/Dockerfile
@@ -8,13 +8,8 @@ RUN go get -d -v ./...
 RUN go vet -v
 RUN go test -v
 
-# Details on when to utilize cgo: 
-#   https://pkg.go.dev/cmd/cgo  
 RUN CGO_ENABLED=0 go build -o /go/bin/app
 
-# Details on when to use Distroless Base vs Static:
-#   https://github.com/GoogleContainerTools/distroless/blob/main/base/README.md
-#FROM gcr.io/distroless/base
 FROM gcr.io/distroless/static
 
 COPY --from=build-env /go/bin/app /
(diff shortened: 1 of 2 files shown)


Shegox:
@googlebot I signed it!

examples/go/Dockerfile
```diff
@@ -1,12 +1,19 @@
...
 
 WORKDIR /go/src/app
-COPY . /go/src/app
+COPY main.go .
 
+RUN go mod init
 RUN go get -d -v ./...
 
-RUN go build -o /go/bin/app
+# Detaails on when to utilize cgo:
```
loosebazooka, line 9: detaails -> details

loosebazooka:
@jonjohnsonjr can you review this go stuff real quick?

examples/go/Dockerfile
michael-basil: corrected the typo

michael-basil:
bump - is there anything else i need to do?

examples/go/Dockerfile
```diff
@@ -1,12 +1,21 @@
...
+RUN go test -v
 
-RUN go build -o /go/bin/app
+# Details on when to utilize cgo: 
+#   https://pkg.go.dev/cmd/cgo  
+RUN CGO_ENABLED=0 go build -o /go/bin/app
+
+# Details on when to use Distroless Base vs Static:
+#   https://github.com/GoogleContainerTools/distroless/blob/main/base/README.md
+#FROM gcr.io/distroless/base
```
loosebazooka, line 17: If we're not going to use this from line, then probably just remove it.

michael-basil: removed ... updated PR

loosebazooka:
Sorry just trying to get someone who does a lot of `go` to review this, otherwise lgtm.

loosebazooka (approved):

examples/go/Dockerfile
```diff
@@ -1,12 +1,16 @@
-FROM golang:1.12 as build-env
+FROM golang:latest as build-env
```
loosebazooka, line 1: sorry I missed this, lets pin this to something: 1.17, it's the style we use in other examples and generally prevents accidental breakages.

michael-basil: There are pros and cons to both pinning the version versus defaulting to latest.

Since the other examples pin the version, the change back to a pinned version seems to make sense for coherence.

Updated.

dekkagaijin (approved):

PR #912, opened 2021-12-21
Update README.md - fix debug tags

ronensc:
Based on the following file, I believe that `debug` should be added as a prefix rather than as a suffix.
https://github.com/GoogleContainerTools/distroless/blob/b19f550dce74b65e09706e46fc82c0075434f563/BUILD#L29

### DIFF
--- a/README.md
+++ b/README.md
@@ -194,7 +194,7 @@ $ docker run --entrypoint=sh -ti my_debug_image
 /app # ls
 BUILD       Dockerfile  hello.py
 ```
-> Note: If the image you are using already has a tag, for example `gcr.io/distroless/java17-debian11:nonroot`, use the tag `<existing tag>-debug` instead, for example `gcr.io/distroless/java17-debian11:nonroot-debug`.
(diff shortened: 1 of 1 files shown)


loosebazooka:
right, good call. Would you be open to signing the CLA? I can just make a new PR with the same changes if you would prefer not to.

ronensc:
> right, good call. Would you be open to signing the CLA? I can just make a new PR with the same changes if you would prefer not to.

Sorry for my late response. I'm working on getting approval to sign the CLA.

ronensc:
@loosebazooka Please try to rerun the CLA check

loosebazooka:
it looks like the email address associated with the commit is not the one used to sign. Can you update it to the one you used to sign. I'm not sure there's a way to do this on github ui, but you can do it from the command line with:

```
git commit --amend --author="John Doe <john@doe.org>"

```

ronensc:
Updated the commit email

loosebazooka (approved):

PR #924, opened 2022-01-11
Bump LTS versions of node 12, 14 and 16

omBratteng:


loosebazooka:
if you would like to script this with something like a ./updateNodeDependencies.sh, I can integrate it into a cron job.

loosebazooka (approved):

omBratteng:
@loosebazooka LTS versions doesn't get updated that often, so I don't mind manually updating them. Just a few minutes of work

PR #925, opened 2022-01-12
use distro instead of distro_suffix

loosebazooka:
drop the suffix, it's cleaner


loosebazooka:
distro suffix was weirding me out, so removed suffix and cleaned up all references.

priyawadhwa (approved):
lgtm!

PR #929, opened 2022-01-14
Rework debian packagemanager

loosebazooka:
- Use a non-bazel tool to generate bazel config to each deb
  individually. It will also be responsible for managing updates
  
  this is just the go part of #926


debian_package_manager/README.md
```diff
@@ -0,0 +1,84 @@
+# debian-package-manager
+
+A debian package index parser with bazel config generation
+
+# Usage
+
+```sh
+# if new versions are available, update snapshots and bazel config
+$ go build -o update cmd/update
```
priyawadhwa, line 9: should we just use `make update` here and have the make command run both of these?

debian_package_manager/README.md
```diff
@@ -0,0 +1,84 @@
...
+$ ./update 
+
+# generate new config for current snapshot
+$ go build -o generate cmd/generate
```
priyawadhwa, line 13: similarly we could just have `make generate` here

debian_package_manager/internal/build/bazel_test.go
```diff
@@ -0,0 +1,117 @@
...
+
+import (
+	"bytes"
+	"testing"
+
+	"github.com/GoogleContainerTools/distroless/debian_package_manager/internal/deb"
+	"github.com/google/go-cmp/cmp"
+)
+
+func testdata() map[string]map[string]map[string]*deb.Package {
```
priyawadhwa, line 11: `map[string]map[string]map[string]*deb.Package` is wild lol

priyawadhwa (approved):

debian_package_manager/README.md
loosebazooka: oh yeah good call

debian_package_manager/internal/build/bazel_test.go
loosebazooka: This makes printing versions pretty easy (which may not be needed once we go full sbom? then we can just add a sorting function to deb.Package and we should get reproducible config generation).

Issue #928, opened 2022-01-14
Distroless TODOs


priyawadhwa:
- move go.mod and go.sum into provenance/
- store chains config in the repo
- install latest tekton/chains in the cloudbuild.yaml
-

PR #933, opened 2022-01-19
Clean up provenance generation

priyawadhwa:
fixes https://github.com/GoogleContainerTools/distroless/issues/928


loosebazooka (approved):

provenance/chains-config.yaml
```diff
@@ -0,0 +1,20 @@
+apiVersion: v1
```
loosebazooka, line 1: does the current config automatically pick this up or does it have to be "applied" some how on the cluster when updated?

priyawadhwa: the gcb step should automatically apply this everytime here:

https://github.com/GoogleContainerTools/distroless/blob/main/cloudbuild.yaml#L111

loosebazooka: sick

PR #943, opened 2022-01-27
Include libcrypt1 in debian11 java images (fix netty-tcnative)

### LINKED ISSUE #942: Debian11 Java images missing libcrypt.so.1 (libcrypt1)
Attempting to use netty-tcnative in a Java application running in a container created from an image that uses any of the following bases:
- `gcr.io/distroless/java:latest`
- `gcr.io/distroless/java11:latest`
- `gcr.io/distroless/java11-debian11:latest`
- `gcr.io/distroless/java17:latest`
- `gcr.io/distroless/java17-debian11:latest`

... results in the following error (rendering netty unable to use native OpenSSL):
```
java.lang.UnsatisfiedLinkError: /tmp/libnetty_tcnative_linux_x86_<snip>.so: libcrypt.so.1: cannot open shared object file: No such file or directory
```

Using the debian10 images works as expected, but for java17 there are no debian10 images published and therefore no work-around if a user needs java17 and needs to use a Java library that contains native code that links against `libcrypt.so.1`.

Debian 10 provides `libcrypt.so.1` with the `libc6` package (via symlink to `/lib/x86_64-linux-gnu/libcrypt-2.28.so`)

Debian 11 has improved this such that `libcrypt.so.1` is provided by a distinct `libcrypt1` package (via symlink to `/lib/x86_64-linux-gnu/libcrypt.so.1.1.0`)

Distroless Java images based on Debian 11 should include the `libcrypt1` package explicitly so that libraries like netty-tcnative will work.

chrisribble:
Fixes #942

### DIFF
--- a/java/BUILD
+++ b/java/BUILD
@@ -27,6 +27,7 @@ EXTRA_DEBS = {
     "debian10": [],
     "debian11": [
         "libbrotli1",
+        "libcrypt1",
     ],
 }
 


chrisribble:
@loosebazooka I see that you've been working on some changes to the way that packages are managed, so I'm not sure if this is how you want this fixed. I'm happy to adjust as needed.

I ran the bazel build, created both java11 and java17 images, created images from them, and exported to a tar with and without my change. Without the change, libcrypt.so.1 is not available; with the change, it exists in the correct location on disk.

Example output with this patch:
```
$ bazel build //...
$ bazel run java:java11_root_amd64_debian11
$ docker create --name bazel_java11 bazel/java:java11_root_amd64_debian11
$ docker export bazel_java11 > bazel_java11.tar
$ tar -tvf bazel_java11.tar | grep libcrypt.so
lrwxrwxrwx 0/0               0 1969-12-31 18:00 lib/x86_64-linux-gnu/libcrypt.so.1 -> libcrypt.so.1.1.0
-rw-r--r-- 0/0          202680 1969-12-31 18:00 lib/x86_64-linux-gnu/libcrypt.so.1.1.0
```

I then rebuilt my application's image using the `bazel/java:java11_root_amd64_debian11` image that I built with bazel. When I start the application, `netty-tcnative` behaves as expected and is able to load all necessary libraries.

priyawadhwa:
seems reasonable to me, but @loosebazooka is the java expert!

loosebazooka (approved):

PR #948, opened 2022-02-06
Add multi-arch to node version tags

### LINKED ISSUE #878: Tag version-specific nodejs ARM images
I'm not sure if this was intentionally done, but there's currently only a single `node:latest-arm64` image, with all other version-specific images being mapped to amd64 only (even though there is an arm64 configuration from what I understand).

Would it be possible to add arm64 tags for every version as well (assuming manifests are not supported to make both available under the same tag)?

Sebmaster:
This converts the :16 tags to be multi-arch based on separate -amd64 and -arm64 tags. This ensures consumers can actually pin to a supported version and take advantage of multi-arch.

Most tags are now constructed systematically, which ensures they can't be missed while manually constructing the image list. This includes :latest and :debug tags.

debian10 is not getting multi-arch images like before.

Fixes #878.

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -120,25 +120,40 @@ PYTHON3.update({
 })
 
 NODEJS = {
-    "{REGISTRY}/{PROJECT_ID}/nodejs:latest-amd64": "//nodejs:nodejs16_amd64_debian11",
-    "{REGISTRY}/{PROJECT_ID}/nodejs:latest-arm64": "//nodejs:nodejs16_arm64_debian11",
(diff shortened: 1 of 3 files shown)


loosebazooka:
Sorry, I deprecated a bunch of image/tags. This PR might need a rework.

Sebmaster:
@loosebazooka Should be fixed.

loosebazooka:
Can you also update the signer as well? https://github.com/GoogleContainerTools/distroless/blob/main/cloudbuild_cosign.sh#L36-L37

Sebmaster:
Added as well.

loosebazooka (approved):

loosebazooka:
thanks!

PR #949, opened 2022-02-08
Remove java, python, node -debian10 builds

loosebazooka:
Removes all python/java/nodejs debian10 images. They have reached EOL in distroless as defined in https://github.com/GoogleContainerTools/distroless/blob/main/SUPPORT_POLICY.md


base/distro.bzl
```diff
@@ -1 +1,5 @@
 DISTROS = ["debian10", "debian11"]
+
+# language images (java, node, python, etc) have shorter support windows
```
dekkagaijin, line 3: maybe "RUNTIME_DISTROS"? To be more descriptive as to what has a shorter support horizon

loosebazooka: yeah, unfortunately we still have CC support for a bit longer (I think kubernetes uses it). I'm okay renaming it, but I don't know if `RUNTIME_DISTROS` is any more descriptive? But I also don't have strong opinions on this. Kind of a problem of having all these combinations of config.

dekkagaijin: fair enough!

dekkagaijin (approved):

PR #969, opened 2022-03-03
Add schedule to dependabot

loosebazooka:


priyawadhwa (approved):

.github/dependabot.yml
```diff
@@ -1,6 +1,10 @@
 version: 2
 updates:
-- package-ecosystem: gomod
+- package-ecosystem: :gomod"
```
naveensrinivasan, line 3: `:gomod"` is invalid :)

naveensrinivasan (changes requested):

.github/dependabot.yml
loosebazooka: woops

naveensrinivasan:
Also, the setting has to be turned on the admin setting. I am sure you are already aware of it.

PR #983, opened 2022-03-21
POC Exclude `npm` and `corepack` from `node_modules`

omBratteng:


loosebazooka:
Be nice to verify in a test that they are not included. Not sure if you have a local bazel dev env to try it out.

omBratteng:
I don't have a local bazel dev env, sadly. Tried to set it up and build locally, but it just builds all images. If you just tell me what specific command I need to run to build node 16 debug, I can test locally

loosebazooka:
ofcourse, pretty solid that you got the build working though :D.

1. build just node images (`bazel build`)
```
$ bazel build //nodejs:nodejs:16_amd64_debian11
$ bazel build //nodejs:nodejs:16_debug_amd64_debian11
```

2. load them into docker (`bazel run`)
```
$ bazel run //nodejs:nodejs:16_debug_amd64_debian11
```

3. use busybox shell
```
$ docker run -it --rm --entrypoint /busybox/sh bazel/nodejs:nodejs16_debug_amd64_debian11
```

-----

for testing you probably want to write file existence tests?
See: https://github.com/GoogleContainerTools/distroless/blob/main/base/base.bzl#L129
and: https://github.com/GoogleContainerTools/distroless/blob/main/base/testdata/base.yaml

omBratteng:
Doesn't seem like adding them here https://github.com/omBratteng/distroless/blob/remove-npm/nodejs/BUILD.nodejs#L14-L15 worked.

Though, if I take the tar from `pkg_tar` and then use it in a `genrule`, where I extract it and package it again

loosebazooka:
ah sorry, I probably should have taken a better look. If you look at the intermediate build output in `/bazel-distroless/nodejs16_amd64/`, the files to exclude are actually (without the leading `nodejs`):
```
"lib/node_modules/npm/**",
"lib/node_modules/corepack/**",
```
It doesn't look like we need the genrule indirection. Lemme know if that works

omBratteng:
Seems like that works

omBratteng:
Tried to add a test, but it failed locally, even on main branch, so letting CI do it
I ran `bazel test --curses=no --test_output=errors //nodejs:nodejs16_debug_amd64_debian11_test`

loosebazooka:
you can run all the nodejs tests locally using
```
bazel test $(bazel query 'attr("tags", "amd64", "//nodejs/...")')
```

omBratteng:
Ah, thanks. Just left the office, so I’ll try it tomorrow. Meanwhile the CI does some testing

nodejs/BUILD
```diff
@@ -27,7 +27,10 @@ NODEJS_MAJOR_VERISONS = ("12", "14", "16")
 [
     container_test(
         name = "nodejs" + major_version + ("" if (not mode) else mode) + "_" + arch + "_" + distro + "_test",
-        configs = ["testdata/nodejs" + major_version + ".yaml"],
+        configs = [
+            "testdata/nodejs" + major_version + ".yaml",
+            "testdata/check_npm.yaml"
```
loosebazooka, line 32: add a `,` at the end to appease buildifier

omBratteng: Added trailing comma

omBratteng:
Will do some more testing tomorrow to see if removing `npm` doesn’t introduce unknown errors. But they shouldn’t, as it’s useless in distroless image. Only usable in debug image

loosebazooka:
Do you see many people using it in a debug image, that might be a more complicated build path, to re-include those dependencies in the debug image.

omBratteng:
I doubt they would be able to use it anyway, because the `npm` executable uses `#!/usr/bin/env node`, which doesn't do anything. I can't even start `npm`, so no need to worry about it.

omBratteng:
Seems like I can't run tests locally because of this issue https://github.com/golang/go/wiki/MacOS12BSDThreadRegisterIssue, so I'll let CI system do the testing. Pushed a test that should fail.

omBratteng:
Failed successfully.
```
=======================================
====== Test file: check_npm.yaml ======
=======================================
=== RUN: File Existence Test: npm
--- FAIL
duration: 0s
Error: Error examining file in container: Error retrieving file from container: API error ([404](https://github.com/GoogleContainerTools/distroless/runs/5640402044?check_suite_focus=true#step:5:404)): Could not find the file /nodejs/lib/node_modules/npm in container 711937909a89d91919b4d19cc20f2c1a53387cf3165[409](https://github.com/GoogleContainerTools/distroless/runs/5640402044?check_suite_focus=true#step:5:409)0994ea73774615f77a
=== RUN: File Existence Test: corepack
--- FAIL
duration: 0s
Error: Error examining file in container: Error retrieving file from container: API error (404): Could not find the file /nodejs/lib/node_modules/corepack in container bdbda8362173e4917f29cdad4a95797b9446698516c34b9b6b06cb1da36b726b
```

loosebazooka:
Thanks for doing this!

PR #995, opened 2022-03-24
Tekton timeout

loosebazooka:


provenance/build-distroless-task.yaml
```diff
@@ -51,6 +51,9 @@ spec:
         # Make sure python points to python3
         update-alternatives --install /usr/bin/python python /usr/bin/python3 0
 
+        # fetch all dependencies with retries and backoff
```
loosebazooka, line 54: also just adds some hardening to this

PR #1010, opened 2022-04-08
Support for latest version of PHP 8

piotrkardasz:
This PR brings the possibility to handle php-fpm `7.4`, `8.0`, `8.1` support from the Ondrej repository
- Add support for basic PHP extensions
- Support `root` and `non-root` users from base `//cc:cc`

Since `package_manager` is created the possibility exists to have a separate `http_file`'s set of archives for PHP so in this case, it would be easier to maintain multiple packages dependencies 🎉 

I also was taking a look at a discussion related to PR https://github.com/GoogleContainerTools/distroless/pull/563 and it would be nice to have ready to use PHP image 😊 

Waiting for feedback from maintainers 🤞


loosebazooka:
Sorry, the enthusiasm for distroless is appreciated, but supporting php is not in our plans.

PR #1018, opened 2022-04-19
Add node18 `18.0.0`

omBratteng:
Unsure wether to add it now, or when Active LTS starts in October.

### DIFF
--- a/node_archives.bzl
+++ b/node_archives.bzl
@@ -30,6 +30,15 @@ def repositories():
         urls = ["https://nodejs.org/dist/v16.14.2/node-v16.14.2-linux-x64.tar.gz"],
     )
 
+	http_archive(
+        name = "nodejs18_amd64",
+        build_file = "//nodejs:BUILD.nodejs",
+        sha256 = "6260d3526dff25d43451ea8e90e0174975b4cd067e8535dc1d85a6d6b29f3043",
+        strip_prefix = "node-v18.0.0-linux-x64/",
+        type = "tar.gz",
+        urls = ["https://nodejs.org/dist/v18.0.0/node-v18.0.0-linux-x64.tar.gz"],
+    )
+
     http_archive(
         name = "nodejs12_arm64",
         build_file = "//nodejs:BUILD.nodejs",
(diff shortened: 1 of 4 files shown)


loosebazooka:
I would wait for active LTS please.

omBratteng:
I just opened this now, as according to the [`SUPPORT_POLICY.md`](https://github.com/GoogleContainerTools/distroless/blob/main/SUPPORT_POLICY.md#node) it says even numbered releases that are current, active or LTS maintenance.
Which by my interpretation, v18 is now current release, then in October it enters Active LTS, and 2023-10-18 it enters Maintenance LTS.

I'm fine with either. But perhaps the [`SUPPORT_POLICY.md`](https://github.com/GoogleContainerTools/distroless/blob/main/SUPPORT_POLICY.md#node) could be changed to say only Active LTS or Maintenance LTS?

> Production applications should only use Active LTS or Maintenance LTS releases.
> _https://nodejs.org/en/about/releases/_

loosebazooka:
lol no you're right, I wrote that policy, I think I misunderstood what was going on here... whoops. Lets add it in.

omBratteng:
> Lets add it in.

node18 or the bit in the [`SUPPORT_POLICY.md`](https://github.com/GoogleContainerTools/distroless/blob/main/SUPPORT_POLICY.md#node)?

v18 was just released, so I haven't gotten to test it yet, so I'll do testing locally tomorrow, that uses `nodejs:18` image.

Also need to wait till [`node:18`](https://hub.docker.com/_/node) has been released too

loosebazooka:
Oh sorry, lets add the node18 image. It would be nice to standardize the naming like the other images though.

for example `node:16` -> `node16:latest`

omBratteng:
You mean renaming `gcr.io/distroless/nodejs:16` to `gcr.io/distroless/node:16`. And adding an alias `gcr.io/distroless/node16:latest`?
Could be done in another PR

loosebazooka:
yeah, makes sense I was thinking about it for java and brought it up here... no need to actually handle it here.

omBratteng:
Seems to work fine, and `node:18` images have finally been published, so this PR is ready for review

loosebazooka (approved):

PR #1025, opened 2022-04-27
Update cloudbuild.yaml

loosebazooka:
update bazel image reference


chanseokoh (approved):

omBratteng:
Interesting, perhaps the image was removed 🤔
Used the command reference in the previous PR. Or I did something wrong, not implausible.

omBratteng:
@loosebazooka document somewhere where you got these hashes from? So next time someone knows where to pull the correct from?

PR #1047, opened 2022-05-15
Add Java 18 to supported versions

tferi:


loosebazooka:
We only intend to support lts versions unfortunately. Your can add a Java 18 jre on top of have base though

PR #1061, opened 2022-06-08
Update example for Go 1.18 and switch to using `go mod download`

mjlshen:
I think `go mod download` is better suited for an example rather than `go get -d -v` as everyone should be using modules at this point.

Also consolidated the example and the one shown in the README as well as bumping to Go 1.18

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,17 +85,16 @@ Follow these steps to get started:
 
   ```dockerfile
   # Start by building the application.
-  FROM golang:1.17-bullseye as build
+  FROM golang:1.18 as build
 
   WORKDIR /go/src/app
-  ADD . /go/src/app
+  COPY . .
 
-  RUN go get -d -v ./...
-
-  RUN go build -o /go/bin/app
+  RUN go mod download
+  RUN CGO_ENABLED=0 go build -o /go/bin/app
 
   # Now copy it into our base image.
-  FROM gcr.io/distroless/base-debian11
+  FROM gcr.io/distroless/static-debian11
   COPY --from=build /go/bin/app /
   CMD ["/app"]
   ```
--- a/examples/go/Dockerfile
+++ b/examples/go/Dockerfile
@@ -1,16 +1,15 @@
-FROM golang:1.17 as build-env
+FROM golang:1.18 as build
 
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -85,17 +85,16 @@ Follow these steps to get started:
...
 
-  RUN go get -d -v ./...
-
-  RUN go build -o /go/bin/app
+  RUN go mod download
+  RUN CGO_ENABLED=0 go build -o /go/bin/app
 
   # Now copy it into our base image.
-  FROM gcr.io/distroless/base-debian11
+  FROM gcr.io/distroless/static
```
loosebazooka, line 97: I would prefer we tell users to pin to a debian version... `static-debian11`

examples/go/Dockerfile
```diff
@@ -1,16 +1,15 @@
...
 
-RUN go mod init
-RUN go get -d -v ./...
+RUN go mod download
 RUN go vet -v
 RUN go test -v
 
 RUN CGO_ENABLED=0 go build -o /go/bin/app
 
 FROM gcr.io/distroless/static
```
loosebazooka, line 12: I guess this should be updated too

loosebazooka:
@jonjohnsonjr @imjasonh can a go person check these over real quick?

README.md
mjlshen: That's a good idea - agreed!

imjasonh:
Are these examples executed during CI at all? It's a lot easier to tell if it works if something actually runs it, than by inspecting it with my pitiful human eyeballs.

loosebazooka:
yeah CI runs them

imjasonh:
I just copied the Dockerfile and ran it and it seems to work, so `/lgtm`, but it'd be nice to have these run in CI too if they're not. 👍

mjlshen:
@loosebazooka rebased, thanks for the reminder - I think it'll need an approval to run the new CI checks

mjlshen:
@loosebazooka just making sure there's nothing remaining to do on my end, thanks for guiding me through this so far

loosebazooka (approved):

Issue #1067, opened 2022-06-20
Enable Java Docker images for architecture s390x


haubenr:
**Describe the bug**
Currently the Java Docker images only support x86_64 and arm64 architectures. Support for architecture s390x should be added (those images are required by the Knative project).

**To Reproduce**
```
# on s390x Linux host
$ uname -p
s390x
$ docker pull gcr.io/distroless/java17-debian11:nonroot
nonroot: Pulling from distroless/java17-debian11
no matching manifest for linux/s390x in the manifest list entries
```

**Expected behavior**
The Java distroless Docker images can be pulled and actually run on a s390x Linux host.

**Console Output**
n/a

**Additional context**
The Knative component 'eventing-kafka-broker' uses the Java distroless Docker images for running various Kafka components in k8s. For extended platform support, the Java distroless images should be available for architecture s390x, too.

haubenr:
> Can you point to an issue in knative project for context?

https://github.com/knative-sandbox/eventing-kafka-broker/issues/2337

PR #1068, opened 2022-06-20
Added s390x support for Java Docker images

haubenr:
Signed-off-by: Dirk Haubenreisser <haubenr@de.ibm.com>

Fixed: #1067 

This PR contains the required changes to enable the Java Docker images for architecture s390x.


debian_packages.yaml
```diff
@@ -42,7 +42,7 @@
 
 # python and java only build on a subset of architectures
 - distros: ["debian11"]
-  archs: ["amd64","arm64"]
```
loosebazooka, line 45: I think I would just separate out the python and java deps here. It's pretty expensive to download some of these deps, and if python isn't building on s390x we shouldn't be fetching them (we prefetch all deps in CI)

haubenr: Sure, I can do that.

haubenr: Done with latest commit.

debian_packages.yaml
```diff
@@ -40,15 +40,38 @@
...
+
+# debian 11, limited architectures, python only
 - distros: ["debian11"]
   archs: ["amd64","arm64"]
   packages:
-  # python
     - "dash"
     - "libbz2-1.0"
     - "libcom-err2"
-    - "libcrypt1"  # TODO: glibc library for -lcrypt; maybe should be in cc?
```
loosebazooka, line 51: I think it's okay to leave this in the python one, duplicates are handled by the system.

loosebazooka (approved):
just a minor issue and then I can merge.

debian_packages.yaml
haubenr: Sure, I've re-added the libcrypt1 dependency to the list of Python dependencies in the latest commit.

loosebazooka (approved):

loosebazooka:
Thanks for contributing. This should trigger a build with the new s390x support

PR #1101, opened 2022-08-10
Ignore s390x and ppc64le security indexes debian10

### LINKED ISSUE #1100: s390x and ppc64le indexes missing from debian security
Update script is failing. Making those indexes optional results in some packages using older versions for those two architectures (which might be okay?).

see: https://security.debian.org/dists/buster/updates/main/

Currently waiting on a response from debian.

loosebazooka:
- They don't seem to be published anymore
- We will probably see some regressions in versions
  for those architectures unfortunately
- Better to not keep the rest of distroless hostage
  on this

fixes #1100

### DIFF
--- a/.github/workflows/ci-cache.yaml
+++ b/.github/workflows/ci-cache.yaml
@@ -1,11 +1,7 @@
 name: cache CI
 
 on:
-  workflow_dispatch: # Allow manual runs.
-  pull_request:
-    branches: [ 'main' ]
-  push:
-    branches: [ 'main' ]
+  workflow_dispatch: # Allow manual runs
 
 jobs:
   build-and-test:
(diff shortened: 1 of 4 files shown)


.github/workflows/ci-cache.yaml
```diff
@@ -1,11 +1,7 @@
 name: cache CI
 
 on:
-  workflow_dispatch: # Allow manual runs.
-  pull_request:
-    branches: [ 'main' ]
-  push:
-    branches: [ 'main' ]
```
chanseokoh, line 8: Are you going to restore these? Seems necessary. If so, another way is to ignore the build failure and force-submit this PR (after temporarily changing the repo setting, if necessary).

chanseokoh (approved):

.github/workflows/ci-cache.yaml
loosebazooka: the gcs cache is not in use. I was going to delete that code. But that gcs cache fails for the same reason that this update was previously failing.

loosebazooka: I will have an update to remove gcs cache completely.

PR #1119, opened 2022-09-06
introduce nossl base image variant

richard78917:
Some projects do not require libssl/openssl or they use
their own.
nossl image variant is used as an intermediate layer for the existing
base.


loosebazooka:
superseded by #1166

PR #1165, opened 2022-11-02
Bump nodejs14 to 14.21.0

omBratteng:


loosebazooka (approved):

loosebazooka:
@omBratteng I would like to add some automation here to at least detect updates (in case you go on vacation or something)

omBratteng:
That probably sounds like like a good idea. You want me to create some github workflow for it?
Thinking of maybe getting the tags from https://github.com/nodejs/node and check against a pattern and if they differ from what's set in node_archives.bzl file?

loosebazooka:
Yeah that seems like it should work. I was thinking we check https://nodejs.org/dist/ for the latest for each version... like https://nodejs.org/dist/latest-v14.x/.

If you've got time to do this, we always appreciate your contributions. If not, I think I can find someone on our side to take a stab at it.

mhio:
node distribute release json and tab files that make processing easier
```
curl -sf https://nodejs.org/download/release/index.json \
 | jq -r '.[] | select(.version|match("^v18")) | "\(.version) \(.date)"'
```
and https://nodejs.org/download/release/index.tab

omBratteng:
@loosebazooka I'm currently swamped with other stuff, so unsure when I'd be able to take a look at creating a workflow for it. So it'd probably be best if you find someone on your side to take a stab at it, or if someone else wants to contribute to it.

loosebazooka:
Sgtm

PR #1168, opened 2022-11-04
Fix Debian version in support policy table

jarda-wien:
I found the current version of the table a little non-informative.

Looking at the previous version, I got to understand how it is supposed to work.


loosebazooka:
It's actually correct as is, but confusing. It is supposed to say debian 12 release day + X time.

jarda-wien:
I've pushed a fix as described.

loosebazooka:
hrmm.. going to try to retrigger this.

Issue #1171, opened 2022-11-08
Fails due to lack of disk space for CI


takumakume:
Hi!

**Describe the bug**

CI fails due to the above error.

> Error: Error creating container: API error (500): failed to write resolv.conf hash file when setting up dns for sandbox 0492affe82b81d8782d48bfd6de19ebd12791cdf523d37feb64e6c2f06d48c18: write /var/lib/docker/containers/2642a4e3a5683f5a57496e0dde412fa886eb23e6b77b4a9b79f9c11ae267b375/resolv.conf.hash: no space left on device

Default remaining disk space for file system is about 30 GB.

```
/dev/root        84G   54G   30G  65% /
```

After `bazel build --curses=no //...` , there are only a few left.

```
/dev/root        84G   82G  1.7G  99% /
```

After that, `no space left on device` .

**To Reproduce**

refs: 
  - https://github.com/GoogleContainerTools/distroless/pull/1169#issuecomment-1305237911 ([CI Job](https://github.com/GoogleContainerTools/distroless/actions/runs/3406965104/jobs/5673491650))
  - https://github.com/GoogleContainerTools/distroless/pull/1170 ([CI Job](https://github.com/GoogleContainerTools/distroless/actions/runs/3414911940/jobs/5683462669))

**Additional context**

#### Proposal

Separate CI jobs to distribute disk usage.

Split by the following units:

```sh
$ bazel query 'attr("tags", "'amd64'", "//...")' | tr ':' '/' | awk -F '/' '{print $3}' | uniq
base
examples
experimental
java
nodejs
```

More specifically, an actions matrix is generated based on this list, and the jobs are divided.

Tested in my environment.

<img width="620" alt="image" src="https://user-images.githubusercontent.com/10905048/200492412-aded2fa7-09c2-4716-9a02-027dd7a906a0.png">

Is there a better way?

PR #1172, opened 2022-11-08
split CI job and resolve lack of disk space

takumakume:
Fix https://github.com/GoogleContainerTools/distroless/issues/1171


joycebrum (approved):
Thanks for the sugestion @takumakume, it was really a good approach to handle the problem.

PR #1187, opened 2022-11-26
Update examples.yaml

loosebazooka:
Fixes broken examples tests, ubuntu-latest tag appears to have been updated to 22.04 (even though docs currently say otherwise). Explicitly use 20.04 here which has a compatible libc mostly for the c examples which compile against it.


loosebazooka:
oh never mind I got this all wrong, it's examples being compiled from bazel. not docker.

PR #1203, opened 2022-12-22
Fix: Use `SOURCE_DATE_EPOCH` for the distroless images.

mattmoor:
:bug: This uses `SOURCE_DATE_EPOCH` so that instead of the Unix epoch, distroless images use the timestamp of the git commit from which they are built.

/kind bug

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -32,7 +32,7 @@ steps:
       export SOURCE_DATE_EPOCH=$(date --date="${_COMMIT_DATE}" +"%s")
     else
       # When a _COMMIT_DATE is not specified, set the SOURCE_DATE_EPOCH
-      # to zero to preserve the reprodusibility of the build.
+      # to zero to preserve the reproducibility of the build.
       export SOURCE_DATE_EPOCH=0
     fi
 
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -28,6 +28,14 @@ steps:
     set -o errexit
     set -o xtrace
 
+    if [ ! -z "${_COMMIT_DATE}" ] ; then
+      export SOURCE_DATE_EPOCH=$(date --date="${_COMMIT_DATE}" +"%s")
+    else
+      # When a _COMMIT_DATE is not specified, set the SOURCE_DATE_EPOCH
+      # to zero to preserve the reprodusibility of the build.
+      export SOURCE_DATE_EPOCH=0
+    fi
+
     # package manager needs wget
     apt-get install wget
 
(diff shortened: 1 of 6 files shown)


cloudbuild.yaml
```diff
@@ -28,6 +28,11 @@ steps:
     set -o errexit
     set -o xtrace
 
+    # Set this so that the build embeds the timestamp of
+    # the source commit it was built from, instead of the
+    # unix epoch.
+    export SOURCE_DATE_EPOCH=$(git log -1 --pretty=%ct)
```
loosebazooka, line 34: any ideas on handling unclean state?

loosebazooka: oh nevermind this is cloud build.

loosebazooka (dismissed):

cloudbuild.yaml
mattmoor: Well unfortunately this doesn't handle `gcloud builds submit` very well, so my ~local test failed to WAI.

Let me see if I can make that a bit more graceful at least :-/

mattmoor: Right now I see:
```
Step #0: + git log -1 --pretty=%ct
Step #0: fatal: not a git repository (or any parent up to mount point /)
Step #0: Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).
Step #0: + export SOURCE_DATE_EPOCH=
```

mattmoor: @loosebazooka talking to @imjasonh @wlynch about GCB's support it sounds like anything based on `.git` is going to be a bust.

Billy recommended `_COMMIT_DATE` from https://cloud.google.com/build/docs/configuring-builds/use-bash-and-bindings-in-substitutions#creating_substitutions_using_payload_bindings which will take some setting up on the GCB side (which obvs I cannot do anymore!), so lemme iterate on this a little bit and I'll circle back when this is RFAL.

mattmoor: Ok, so I finally managed to get things plumbed through: 
image

Where I pass `_COMMIT_DATE=2022-12-22T12:16:08-05:00` on the CLI.  I verified this with ~just a static build (commented out other targets and signing stuff), since this stuff takes forever to run, the full command I used was:
```
gcloud builds submit --project mattmoor-chainguard --substitutions _COMMIT_DATE=$(git log -1 --date=iso-strict --pretty=%cd)
```

mattmoor: Alright, I just did a fresh build (with the commit timestamp for this build) and got this: 
image

I'm doing a build now without `--stamp` and `_COMMIT_DATE` to see what we get 🤞

loosebazooka (commented):
Kk, I've updated the trigger and mostly lgtm.

The python build needs this update as well?
https://github.com/mattmoor/distroless/blob/set-source-date-epoch/experimental/python3/BUILD

loosebazooka:
@thesayyn no conflict with the new rules?

mattmoor:
@loosebazooka fixed the typo and added Python.

PR #1242, opened 2023-02-17
ci: introduce caching and update bazel

thesayyn:


.cloudbuild/release.sh
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env bash
+set -o errexit -o xtrace
+
+# setup remote cache
+curl -fsSL https://github.com/buchgr/bazel-remote/releases/download/v2.4.0/bazel-remote-2.4.0-linux-x86_64 -o bazel-remote
```
loosebazooka, line 5: can you ensure the hash here?
comparing sha256sum bazel-remote to some fixed value?

.cloudbuild/release.sh
```diff
@@ -0,0 +1,30 @@
...
+chmod +x bazel-remote
+mkdir .logs
+./bazel-remote --max_size 8 --dir ~/.cache/bazel-remote --experimental_remote_asset_api --grpc_address 0.0.0.0:4700 --gcs_proxy.bucket $REMOTE_CACHE_GCS --gcs_proxy.use_default_credentials > .logs/bazel-remote.log 2>&1 &
+
+# install bazel 6
+apt-get install bazel-6.0.0
+ln -sf /usr/bin/bazel-6.0.0 /usr/bin/bazel
+
+# setup docker-credential-gcr
+curl -fsSL "https://github.com/GoogleCloudPlatform/docker-credential-gcr/releases/download/v2.1.6/docker-credential-gcr_linux_amd64-2.1.6.tar.gz" | tar xz docker-credential-gcr
```
loosebazooka, line 15: I thought cloud build already had credentials configured with gcloud? This may not be necessary

.cloudbuild/release.sh
```diff
@@ -0,0 +1,30 @@
...
+chmod +x docker-credential-gcr && mv docker-credential-gcr /usr/bin/
+docker-credential-gcr configure-docker
+
+# setup remote caching and remote asset API.
+echo "common --remote_cache=grpc://0.0.0.0:4700" >> ~/.bazelrc
+echo "common --experimental_remote_downloader=grpc://0.0.0.0:4700" >> ~/.bazelrc
+echo "common --google_default_credentials" >> ~/.bazelrc
+echo "common --announce_rc" >> ~/.bazelrc
+
+for i in $(seq 5); do
```
loosebazooka, line 25: what does this do? does this prepopulate the cache?

private/stamp.bash
```diff
@@ -1,20 +1,15 @@
...
+#!/usr/bin/env bash
+set -o pipefail -o errexit -o nounset
 
 # Do not error if the user doesn't have gcloud installed.
 if [ -z "${PROJECT_ID:-}" ]; then
   if which gcloud > /dev/null; then
     PROJECT_ID="$(gcloud config get-value core/project)"
   else
-    PROJECT_ID="NO-PROJECT"
+    PROJECT_ID="no-project"
```
loosebazooka, line 9: why this change?

.cloudbuild/release.sh
thesayyn: i remember that this behavior provided via their official runner images. this might a left over from when i was trying multiple bazel images. I'll comment it out to see if it's necessary.

.cloudbuild/release.sh
thesayyn: exactly. this is in case there's a massive cache misses and bazel needs to download a bunch.

private/stamp.bash
thesayyn: this should error out as well since it's a generic name. while uppercase chars lead to cryptic messages in some registries, this will show meaningful error message like `no such project no-project`. I am fine reverting it though.

loosebazooka: kk lets leave it like this.

.cloudbuild/release.sh
thesayyn: done

PR #1255, opened 2023-03-08
fix: use cosign attest instead of attach

thesayyn:


private/oci/sign_and_push.bzl
```diff
@@ -8,9 +8,19 @@ tag="$(stamp "{TAG}")"
 
 [[ -n $EXPORT ]] && echo "$repository:$tag" >> $EXPORT
 
-"$(realpath {ATTACH_CMD})" --repository "$repository"
-"$(realpath {SIGN_CMD})" --repository "$repository" --key "$KEY" --attachment sbom
+# Push the image by its digest
+"$(realpath {PUSH_CMD})" --repository "$repository"
+
+# Attest the sbom
+"$(realpath {ATTEST_CMD})" --repository "$repository" --key "$KEY" --tlog-upload=false
```
loosebazooka, line 15: I think this is fine for now, but we probably want to upload everything to the tlog eventually.

loosebazooka (approved):

private/oci/sign_and_push.bzl
thesayyn: yes. I didn't like the idea that bazel hitting a server other than the registry. if/when goes down, our cd pipeline will do down with it.

loosebazooka: keyless has to hit the tlog anyway for it to be useful :shrug:

thesayyn: hmm, I didn't think about that. not sure what should be done here. add tlog back?

thesayyn: sounds good

PR #1256, opened 2023-03-10
refactor: use rules_oci instead of rules_docker

thesayyn:


.cloudbuild/cloudbuild.yaml
```diff
@@ -12,48 +12,11 @@ artifacts:
 steps:
   - name: gcr.io/cloud-builders/bazel@sha256:e82204cf0715edb80e4957fa55f944b9dd06b0b1c4eb4b0ca0022de7839726f6
     env:
-      - PROJECT_ID=${PROJECT_ID}
+      - PROJECT_ID=${PROJECT_ID}/testing
```
loosebazooka, line 15: lets remove `/testing` here

.cloudbuild/cloudbuild.yaml
```diff
@@ -12,48 +12,11 @@ artifacts:
 steps:
```
loosebazooka, line 12: if we can up the timeout at the top of this file to 2.5/3 hours for now till we remove some of the signing, I think that would be helpful to prevent the build from failing unnecessarily.

loosebazooka (commented):
minor cloud build config

.cloudbuild/cloudbuild.yaml
```diff
@@ -1,4 +1,4 @@
-timeout: 7200s  # 2 hours
+timeout: 10800s  # 2 hours
```
loosebazooka, line 1: comment -> 3 hours (confusing)

thesayyn: oh sorry. fixed now

loosebazooka (approved):

PR #1268, opened 2023-03-16
feat: support riscv64

ernado:


loosebazooka:
I probably need some strong justification for this. Adding arches to the build is quite expensive and makes it harder to maintain.

ernado:
Totally agree!

I'm currently working on riscv64 support for Kubernetes, which uses distroless (AFAIK) and also requires etcd image.

I've been asked to use distroless in https://github.com/etcd-io/etcd/pull/15490 while adding riscv64 support to etcd.

debian_packages.yaml
```diff
@@ -1,6 +1,6 @@
 # debian 11
 - distros: ["debian11"]
-  archs: ["amd64", "arm64", "arm", "s390x", "ppc64le"]
+  archs: ["amd64", "arm64", "arm", "s390x", "ppc64le", "riscv64"]
```
loosebazooka, line 3: riscv64 should be limited to only the static images and dependencies. This will reduce impact on the build.

There's no need to pull in all packages for riscv64

ernado: Thank you for your time!

How should I update this file?

Also I've got following error while trying `./updateWorkspaceSnapshots.sh`:
```
..URL: https://snapshot.debian.org/archive/debian/20230314T211539Z/dists/bullseye/main/binary-riscv64/Packages.xz, Status: 404 NOT FOUND
panic: failed (status: 404) to fetch remote file: "https://snapshot.debian.org/archive/debian/20230314T211539Z/dists/bullseye/main/binary-riscv64/Packages.xz"
```

So there is no snapshot for riscv64.
Is it even possible to proceed having this error?

loosebazooka: honestly I don't know where the riscv64 images even come from. I can't even find the non-snapshot repo (after taking a quick look). Best I could find is that the port is still "in progress" -- https://www.debian.org/ports/

ernado: Oh, looks like I've got it now, sorry, I've updated BUILD file.

loosebazooka: would probably need the team that owns that port to commit to some sort of snapshotting for this to even work.

PR #1279, opened 2023-03-21
Set Token Permissions to Github Workflows

joycebrum:
### Description

Configuring token permissions to increase distroless OpenSSF Scorecard score 😄 

### Changes

- Set token permission to all workflows
- Change update snapshots to use GITHUB_TOKEN instead of PAT since it is considered safer. That's because GITHUB_TOKEN has validity only during the workflow run and can not be exploited if the workflow has ended even if it (the token) got compromised. The PAT can be exploited anytime if compromised.

   I've got the GITHUB_TOKEN permission needed from the create pull request action documentation https://github.com/peter-evans/create-pull-request#workflow-permissions.

### Test runs

- Buildfier **success** https://github.com/joycebrum/distroless/actions/runs/4483015542
- CI **success** https://github.com/joycebrum/distroless/actions/runs/4482977438
- Examples **success** https://github.com/joycebrum/distroless/actions/runs/4483015527
- package manager CI - on pull request - **success** https://github.com/joycebrum/distroless/actions/runs/4483015534
- package manager CI - on push - **success** https://github.com/joycebrum/distroless/actions/runs/4482996957
- Pylint **success** https://github.com/joycebrum/distroless/actions/runs/4483015532
- update-snapshots **success** https://github.com/joycebrum/distroless/actions/runs/4483258410

### DIFF
--- a/.github/workflows/update-deb-package-snapshots.yml
+++ b/.github/workflows/update-deb-package-snapshots.yml
@@ -31,7 +31,7 @@ jobs:
     - name: Create Pull Request
       uses: peter-evans/create-pull-request@38e0b6e68b4c852a5500a94740f0e535e0d7ba54 # v4.2.4
       with:
-        token: ${{ secrets.GITHUB_TOKEN }}
+        token: ${{ secrets.ACTIONS_TOKEN }}
         commit-message: "Bumping packages to latest stable versions"
         author: "Distroless Bot <distroless-bot@google.com>"
         title: "Bumping packages to latest stable versions"


.github/workflows/update-deb-package-snapshots.yml
```diff
@@ -3,10 +3,15 @@ on:
...
     - cron:  '35 8 * * *'
+  
+permissions:
+  contents: read
+    
 jobs:
   build:
     runs-on: ubuntu-latest
     permissions:
+      contents: write
```
loosebazooka, line 14: I'm confused, is it overwriting what is set at line 8?

.github/workflows/update-deb-package-snapshots.yml
```diff
@@ -26,7 +31,7 @@ jobs:
     - name: Create Pull Request
       uses: peter-evans/create-pull-request@38e0b6e68b4c852a5500a94740f0e535e0d7ba54 # v4.2.4
       with:
-        token: ${{ secrets.ACTIONS_TOKEN }}
+        token: ${{ secrets.GITHUB_TOKEN }}
```
loosebazooka, line 34: ACTIONS_TOKEN is required here, this allows the generated PR to have actions triggered against it.

loosebazooka: https://docs.github.com/en/actions/using-workflows/triggering-a-workflow#triggering-a-workflow-from-a-workflow

although I think I should re-make ACTIONS_TOKEN to be repo scoped

.github/workflows/update-deb-package-snapshots.yml
joycebrum: Yes. The top level is set to `contents: read` which ensures that any new job eventually added to the workflow will have only permission to read the repo content.

The `contents: write` permission to the job is to allow the create_pull_request to run,

loosebazooka (approved):

PR #1288, opened 2023-04-01
Add precision-shell as a user of Distroless

groboclown:
Precision-Shell is a tool to provide 'just enough shell' to add a touch of functionality on top of a minimal docker image.  The tool uses Distroless in the [recipes](https://github.com/groboclown/precision-shell/tree/main/recipes) for using the tool.


loosebazooka:
Closing this for now. I think use of that section is basically deprecated. I don't think we'll be updating it.

Issue #1304, opened 2023-04-24
New OCI logic introduces unsafe dependency on debian packaging format


howardjohn:
**Describe the bug**
Since https://github.com/GoogleContainerTools/distroless/pull/1256, we assume the debian file has control.tar.xz. This is not always the case, for example: https://snapshot.debian.org/archive/debian/20230421T152012Z/pool/main/libn/libnfnetlink/libnfnetlink0_1.0.1-3+b1_armhf.deb has control.tar.gz

**To Reproduce**
This can be done by adding a dependency on a package that uses this format. None in core do, I assume, but we have a small fork with an extra package

```
debian_archive(
    name = "amd64_debian11_libnfnetlink0",
    package_name = "libnfnetlink0",
    sha256 = "6d650aae2fbfcb261557ed9032ba00c6d69ff30eded71050459af0fc0cf50926",
    urls = ["https://snapshot.debian.org/archive/debian/20230421T152012Z/pool/main/libn/libnfnetlink/libnfnetlink0_1.0.1-3+b1_amd64.deb"],
)
```

**Expected behavior**
Build works

**Console Output**
If applicable, add information from your container run

**Additional context**
I'll submit a fix

PR #1305, opened 2023-04-24
Fix debian compression assumptions

howardjohn:
Fixes https://github.com/GoogleContainerTools/distroless/issues/1304


loosebazooka:
thanks for this.
just kind of curious how you're using distroless right now?

howardjohn:
We have a forked image that just has glibc+iptables: https://github.com/istio/distroless/commit/iptables.

PR #1310, opened 2023-05-02
Updating busybox to 1.35.0

billdinger:
This is an update to busybox to [1.35.0](https://github.com/docker-library/official-images/blob/master/library/busybox). 

I was hoping to upgrade it to `1.36.0` or higher to get rid of this [CVE](https://nvd.nist.gov/vuln/detail/CVE-2022-28391) but unfortunately that is still marked as `unstable` in the docker-library tag list so seems "unwise".  `1.35.0` also matches alpine's currently used version

image

Busybox themselves [seems to have marked their last stable release as 1.34.1](https://busybox.net/news.html) in 2021 so it's possible I'm mistaken here and we want to for stability sake leave this as `1.34.1` and close this mr.

### DIFF
--- a/base/testdata/debug.yaml
+++ b/base/testdata/debug.yaml
@@ -7,4 +7,4 @@ fileExistenceTests:
 commandTests:
   - name: busybox
     command: ["/busybox/busybox"]
-    expectedOutput: ['BusyBox v1\.34\.1']
+    expectedOutput: ['BusyBox v1\.35\.0']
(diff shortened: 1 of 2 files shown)


loosebazooka:
The debug image isn't meant to be used in production? Is there a reason you need it?

billdinger:
Nah, not using it in prod only for debug workflows.

loosebazooka (approved):

PR #1315, opened 2023-05-09
Update sign_and_push.bzl

loosebazooka:
Use keyless for sbom attestations


loosebazooka:
hopefully the build doesn't take forever

PR #1325, opened 2023-06-01
Fix update when no snapshot exists for this month

hausi73:
I noticed that the update was not working on 1st of June because there is/was a "404 Not Found" error at least until 17:30 CET when requesting https://snapshot.debian.org/archive/debian/?year=2023&month=6.

### DIFF
--- a/debian_package_manager/internal/deb/snapshot.go
+++ b/debian_package_manager/internal/deb/snapshot.go
@@ -68,7 +68,7 @@ func latest(urltemplate string) (string, error) {
 }
 
 func latestFromUrl(snapshotURL string) (string, error) {
-	resp, err := http.Get(snapshotURL)
+	resp, err := rhttp.Get(snapshotURL)
 	if err != nil {
 		return "", err
 	}
--- a/debian_package_manager/internal/deb/snapshot.go
+++ b/debian_package_manager/internal/deb/snapshot.go
@@ -55,7 +55,20 @@ var (
 func latest(urltemplate string) (string, error) {
 	year, month, _ := time.Now().Date()
 	snapshotURL := fmt.Sprintf(urltemplate, year, month)
-	resp, err := rhttp.Get(snapshotURL)
+	resp, err := latestFromUrl(snapshotURL)
+	if err != nil {
+		snapshotURL1 := fmt.Sprintf(urltemplate, year, month - 1)
+		resp1, err1 := latestFromUrl(snapshotURL1)
+		if err1 != nil {
+			return "", err1
+		}
+		return resp1, err1
+	}
+	return resp, err
+}
+
+func latestFromUrl(snapshotURL string) (string, error) {
+	resp, err := http.Get(snapshotURL)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


debian_package_manager/internal/deb/snapshot.go
```diff
@@ -55,7 +55,20 @@ var (
 func latest(urltemplate string) (string, error) {
 	year, month, _ := time.Now().Date()
 	snapshotURL := fmt.Sprintf(urltemplate, year, month)
-	resp, err := rhttp.Get(snapshotURL)
+	resp, err := latestFromUrl(snapshotURL)
+	if err != nil {
```
loosebazooka, line 59: I think if we're going to do this. We should probably handle the new year case as well. If (month == 1) or whatever

debian_package_manager/internal/deb/snapshot.go
```diff
@@ -55,7 +55,20 @@ var (
...
+		if err1 != nil {
+			return "", err1
+		}
+		return resp1, err1
+	}
+	return resp, err
+}
+
+func latestFromUrl(snapshotURL string) (string, error) {
+	resp, err := http.Get(snapshotURL)
```
loosebazooka, line 71: We probably want to continue to use rhttp in some manner

loosebazooka:
@hausi73 thanks for this. Can you sign the cla so we may proceed?

hausi73:
The CLA requirement should now be fulfilled.
How can I retrigger the check?

bobcallaway:
> The CLA requirement should now be fulfilled. How can I retrigger the check?

looks like the CLA is all set, please rebase this and it should be good to go

loosebazooka:
I think rescanning instructions are in a failed scan: https://github.com/GoogleContainerTools/distroless/runs/13949870203

But a rebase is required here to pick up the linter updates

hausi73:
I did a rebase to the main branch and now the CLA is green.

loosebazooka:
Sorry I'm not really around good internet till Thursday. Will try to get it to then

debian_package_manager/internal/deb/snapshot.go
```diff
@@ -55,6 +55,25 @@ var (
 func latest(urltemplate string) (string, error) {
 	year, month, _ := time.Now().Date()
 	snapshotURL := fmt.Sprintf(urltemplate, year, month)
+	resp, err := latestFromUrl(snapshotURL)
+	if err != nil {
```
loosebazooka, line 59: I actually think we should only take this path if the currently ignored `day` from time.Now().Date() is "1"

loosebazooka: > And how should we react if the day is 15? 

So, we work under the assumption that snapshots.debian.org produces snapshot at least daily. We really should only be going to a previous snapshot if there are no snapshot available yet. 

Obtaining indexes could fail for any number of reasons as we generally experience some amount of random failures with this repository. We don't want to accidentally go back a month in that case.

PR #1351, opened 2023-06-30
fix tar on deb format file in locale

yangzuo0621:
Argument 1 is the deb file, it should translate into tar format, then apply the tar command to extract the expected files.
Otherwise, errors are like followings:

```
tar: This does not look like a tar archive
tar: Skipping to next header
tar: ./usr/lib/locale/C.UTF-8: Not found in archive
tar: ./usr/share/doc/libc-bin/copyright: Not found in archive
tar: Exiting with failure status due to previous errors
```

### DIFF
--- a/locale/locale.bzl
+++ b/locale/locale.bzl
@@ -16,8 +16,10 @@ def _impl(ctx):
         ],
         command = """
         set -o pipefail -o errexit -o nounset
+        ar -x "$1" data.tar.xz
         tmp=$(mktemp -d)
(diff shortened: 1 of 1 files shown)


locale/locale.bzl
```diff
@@ -16,8 +16,10 @@ def _impl(ctx):
         ],
         command = """
         set -o pipefail -o errexit -o nounset
+        ar -x "$1" data.tar.xz
```
loosebazooka, line 19: could this just happen after `tmp=$(mktemp -d)` and use the `--output` option in `ar`

yangzuo0621: Modify it according to the comments.

loosebazooka:
i believe this is handled now by the special debian package rules.

PR #1378, opened 2023-08-16
Use hashicorp retryable http

loosebazooka:
this appears to handle connection issues (premature connection closure from debian snapshots) much better and automatically compared to the custom impl we had.


bobcallaway (approved):
we've used this in sigstore and have been happy with it.

Issue #1382, opened 2023-08-22
CPU arch variant is not specified on all ARM Distroless images


lbcjbb:
## Describe the bug

The CPU arch variant should be specified in the manifest of all ARM images:

- `arm` => `v6` or `v7`
- `arm64` => `v8`

## To Reproduce

```shell
$ crane manifest gcr.io/distroless/static-debian11 | jq '.manifests[].platform'
{
  "os": "linux",
  "architecture": "amd64"
}
{
  "os": "linux",
  "architecture": "arm64"
}
{
  "os": "linux",
  "architecture": "arm"
}
{
  "os": "linux",
  "architecture": "s390x"
}
{
  "os": "linux",
  "architecture": "ppc64le"
}
```

## Expected behavior

```shell
$ crane manifest gcr.io/distroless/static-debian11 | jq '.manifests[].platform'
{
  "os": "linux",
  "architecture": "amd64"
}
{
  "os": "linux",
  "architecture": "arm64",
  "variant": "v8"
}
{
  "os": "linux",
  "architecture": "arm",
  "variant": "v7"
}
{
  "os": "linux",
  "architecture": "s390x"
}
{
  "os": "linux",
  "architecture": "ppc64le"
}
```

## Additional context

Note: [rules_oci](https://github.com/bazel-contrib/rules_oci) doesn't allow to create an image from these base images and to update this missing information, so we have to do without it.

PR #1383, opened 2023-08-22
Add missing CPU variant for all ARM images

lbcjbb:
Add missing CPU variant for all ARM images.

Closes: #1382

### DIFF
--- a/base/base.bzl
+++ b/base/base.bzl
@@ -1,6 +1,6 @@
 # defines a function to replicate the container images for different distributions
 load("//cacerts:cacerts.bzl", "cacerts")
-load("//:checksums.bzl", "ARCHITECTURES")
+load("//:checksums.bzl", "ARCHITECTURES", "VARIANTS")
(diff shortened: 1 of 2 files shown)


base/base.bzl
```diff
@@ -103,6 +103,7 @@ def distro_components(distro):
                 workdir = workdir,
                 os = "linux",
                 architecture = arch,
+                variant = VARIANTS.get(arch),
```
loosebazooka, line 106: presumably this just bubbles down to all images?

lbcjbb: Yes

loosebazooka:
Anything in particular you need this for, just out of curiousity? Was this blocking some process for you?

loosebazooka (approved):

lbcjbb:
No, it's not actually a barrier for me. This issue is trivial, but in the future with ARM64v9 (future AWS Graviton 4?) the CPU variant must be required.

PR #1384, opened 2023-08-23
Revert io_bazel_rules_go to v0.35.0

dahu33:
Fixes #1381

[rules_go](https://github.com/bazelbuild/rules_go) is broken for the ppc64le arch since version v0.36.0 which introduced [this change](https://github.com/bazelbuild/rules_go/pull/3336).

Until https://github.com/bazelbuild/rules_go/issues/3531 is resolved a good workaround is to revert to v0.35.0.

PR #1365, opened 2023-07-26
draft: feature: add bookworm 12

maxjuhlke:
This feature will try to enable distroless 12.

This tries to re-implement:

https://github.com/GoogleContainerTools/distroless/pull/1146

PR #1385, opened 2023-08-23
Add Debian 12 (bookworm) images

### LINKED ISSUE #1337: Debian 12 Support
Debian 12 "bookworm" has been released!

News article [here](https://www.debian.org/News/2023/20230610).

dahu33:
An attempt to fix #1337 and might supersede #1365

Requires #1384 to build on a fresh setup.

This builds `base`, `static`, `base-nossl` and `cc` images. Happy to work on the other targets after feedback from maintainers.


base/base.bzl
```diff
@@ -119,7 +119,7 @@ def distro_components(distro):
                 base = ":static_" + user + "_" + arch + "_" + distro,
                 tars = [
                     deb_pkg(arch, distro, "libc6"),
-                    deb_pkg(arch, distro, "libssl1.1"),
+                    deb_pkg(arch, distro, "libssl1.1" if distro == "debian11" else "libssl3"),
                     deb_pkg(arch, distro, "openssl"),
```
loosebazooka, line 123: I would prefer we remove openssl from debian12 images. It can be available in debug

cc/BUILD
```diff
@@ -10,6 +10,9 @@ DISTRO_DEBS = {
     "debian11": [
```
loosebazooka, line 10: if there are the same, just put them in the tars section. I feel like this might be remnants from debian10 days, and doesn't appear to be necessary anymore.

locale/locale.bzl
```diff
@@ -4,30 +4,32 @@ load("@rules_pkg//:providers.bzl", "PackageFilesInfo")
 load("@rules_pkg//:pkg.bzl", "pkg_tar")
 
 def _impl(ctx):
+    cutf8 = "C.UTF-8" if ctx.attr.distro == "debian11" else "C.utf8"
```
loosebazooka, line 7: any doc you can point to for this? for future reference?

README.md
```diff
@@ -98,10 +102,13 @@ Follow these steps to get started:
 
 * Pick the right base image for your application stack.
   We publish the following distroless base images on `gcr.io`:
+    * [gcr.io/distroless/static-debian12](base/README.md)
     * [gcr.io/distroless/static-debian11](base/README.md)
+    * [gcr.io/distroless/base-debian12](base/README.md)
     * [gcr.io/distroless/base-debian11](base/README.md)
```
loosebazooka, line 108: Sorry, this appears to be a previous oversight from the maintainers side, but since you're in here can we add in base-nossl here as well?

loosebazooka (commented):
Thanks this looks pretty good, just some minor things.

dahu33:
> Thanks this looks pretty good, just some minor things.

Thanks, addressed all your feedback, could you please have another look?

base/testdata/base.yaml
```diff
@@ -25,8 +25,11 @@ fileExistenceTests:
 - name: services
   path: '/etc/services'
   shouldExist: true
-- name: tzdata
-  path: '/usr/sbin/tzconfig'
+- name: tzdata_copyright
+  path: '/usr/share/doc/tzdata/copyright'
+  shouldExist: true
+- name: tzdata_zoneinfo
+  path: '/usr/share/zoneinfo'
```
dahu33, line 32: `/usr/sbin/tzconfig` has been removed in debian12. See: https://metadata.ftp-master.debian.org/changelogs//main/t/tzdata/tzdata_2023c-5_changelog
>   * Drop obsolete tzconfig command (use "dpkg-reconfigure tzdata" instead)

As a workaround, I've added the copyright file and the zoneinfo folder that exist in both debian11 and debian12.

README.md
```diff
@@ -30,6 +30,10 @@ The following images are currently published and updated by the distroless proje
 
 | Image                                 | Tags                                   | Architecture Suffixes             | 
 | ---                                   | ---                                    | ---                               |
+| gcr.io/distroless/static-debian12     | latest, nonroot, debug, debug-nonroot  | amd64, arm64, arm, s390x, ppc64le |
+| gcr.io/distroless/base-debian12       | latest, nonroot, debug, debug-nonroot  | amd64, arm64, arm, s390x, ppc64le |
+| gcr.io/distroless/base-nossl-debian12 | latest, nonroot, debug, debug-nonroot  | amd64, arm64, arm, s390x, ppc64le |
+| gcr.io/distroless/cc-debian12         | latest, nonroot, debug, debug-nonroot  | amd64, arm64, arm, s390x, ppc64le |
```
loosebazooka, line 36: These actually do not hold true with the current configuration of the BUILD. There are a few more modifications there needed to make a claim that we are building all the variants.

BUILD
```diff
@@ -54,6 +55,7 @@ BASE_VARIANTS = [
 BASE = {
     "{REGISTRY}/{PROJECT_ID}/base:{COMMIT_SHA}": "//base:base_root_amd64_debian11",
     "{REGISTRY}/{PROJECT_ID}/base-debian11:{COMMIT_SHA}": "//base:base_root_amd64_debian11",
+    "{REGISTRY}/{PROJECT_ID}/base-debian12:{COMMIT_SHA}": "//base:base_root_amd64_debian12",
```
loosebazooka, line 58: We need to modify the commands for each "static/base/base-nossl" to build the arch variants, the oci images indexes. 

I think it's actually fine to leave this as is for now, and deal with in a followup though.

loosebazooka (commented):
Looks like one test is failing, but there's also a minor documentation issue. otherwise lgtm.

README.md
dahu33: I'll try to find some time to address those today. Also, I've spent some time to try to fix the broken test but I couldn't figure why it fails. Any clue?

BUILD
dahu33: I'm not sure to follow here, isn't it already the case? Example here:
https://github.com/GoogleContainerTools/distroless/blob/43d0e68992a4a9b0b87aaf66c986088205979218/BUILD#L40-L45

loosebazooka: omg, yeah you're right, I'm a dumb. I misread the build file.

loosebazooka (commented):
Awesome, thanks for contributing!

loosebazooka (approved):

Sineaggi:
Would love to see the java targets added as well.

loosebazooka:
I think the language targets will be added slowly as we roll this out gradually.

dahu33:
@loosebazooka thanks for the quick review!

> Would love to see the java targets added as well.

@Sineaggi working on it, will see if I can get it out this week.

PR #1397, opened 2023-09-12
Add java 17 support for debian 12

Sineaggi:
Java 11 (jre/jdk) is not available under debian 12.

```
$ apt-get install openjdk-11-jdk-headless
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
Package openjdk-11-jdk-headless is not available, but is referred to by another package.
This may mean that the package is missing, has been obsoleted, or
is only available from another source
However the following packages replace it:
  openjdk-17-jre-headless

E: Package 'openjdk-11-jdk-headless' has no installation candidate
```


Sineaggi:
Referencing https://github.com/GoogleContainerTools/distroless/issues/1398

loosebazooka:
@Sineaggi ready for review?

Sineaggi:
> @Sineaggi ready for review?

I think so, sorry about the churn.

Sineaggi:
Few things to note. There's a bit of copy paste, some necessary and some potentially avoidable with a refactor.

There's a library difference in libpcre2 between debian 11 and 12.

I ran thru the list of packages that openjdk depends on and double-checked what was necessary and what wasn't. I'm wondering if we can move to nossl as the base for java images in the future? Seems like cups (and related dependencies) are the only things ssl would be needed for.

loosebazooka:
yeah we can probably remove libssl from the java image? Much easier to start without it and add it back in later if there's a legitimate reason

Sineaggi:
@loosebazooka would you prefer the nossl here, on in a separate pr?

loosebazooka:
here is good

Sineaggi:
> here is good

Sounds good, done.

loosebazooka (approved):

loosebazooka:
Thanks so much for doing this!

Issue #1398, opened 2023-09-12
Debian 12 language images


loosebazooka:
Finish up support for remain debian 12 images
- [ ] python3
- [ ] java17
- [ ] nodejs (18, 20)

PR #1403, opened 2023-09-16
feat: Add Python 3 support for Debian 12

siketyan:
This pull request adds Python (v3.11) language support for Debian 12 (bookworm).
References #1398.


loosebazooka:
Thanks for doing this, can you just add the test to mirror the test that exists for debian11?

siketyan:
@loosebazooka Added missing testdata for Debian 12. Could you please approval the workflow again?

loosebazooka (approved):

PR #1415, opened 2023-09-29
Rework python3 on debian12

loosebazooka:
- removes shell
- removes support for ctypes
- brings python3 out of experimental for debian12


loosebazooka:
@evanj @lathama would you have a minute to test these images before merging this pr?

add oci_tarball rule
```diff
--- a/python3/BUILD
+++ b/python3/BUILD
@@ -124,3 +124,9 @@ DISTRO_VERSION = {
     for arch in ARCHITECTURES
     for distro in DISTROS
 ]
+
+oci_tarball(                                                                                                         
+  name = "python3_tarball",                                                                                          
+  image = ":python3_root_amd64_debian12",                                                                            
+  repotags = ["python3_distroless:1"],                                                                               
+) 
```

build and load into docker daemon
```
$ bazel build //python3:python3_tarball
...
$ docker load < bazel-bin/python3/python3_tarball/tarball.tar
```

then run or use in a docker build (although I'm mostly interested if this runs your applications)
```
$ docker run python3_distroless:1 \-\-help
```

evanj:
I just tried it with a container I have that uses gunicorn and it appears to work for me. Thanks!

bobcallaway (approved):

waaffles:
Sorry to rehash an older PR that's already been merged but can we update the docs somewhere around how python3:debug doesn't actually include a shell? As of now, isn't the allure of debug the fact that you get a shell to be able to debug?

loosebazooka:
I believe you can execute commands with the shell, what are you trying to do exactly? The path to shell is slightly different. It's /busybox/sh ?

waaffles:
Ah, okay. So the debug option doesn't allow sh to be used within the dockerfile itself and instead sets the entrypoint to be `/busybox/sh`? I was also trying to run some commands from within the dockerifle as I needed to use `mkdir` and `ln`. Is it only `sh` that's offered or are other binaries in place as well? I wasn't sure where to check for the full list of which utilities we get with the `debug` option

PR #1417, opened 2023-10-04
Update libc6 to patch CVE-2023-4911

fellipecmfastlane:


agilgur5:
For posterity, this was replaced by #1418 then superseded by #1419

PR #1447, opened 2023-10-29
fail if subcommand fails in CI

loosebazooka:
assignment seems to surface the subcommand failure
also pre-fetch was failing for some other reasons, what a mess :shrug:


bobcallaway (approved):

loosebazooka:
CI is getting rate limited or something, or the debian-snapshots serving infra is getting hammered? unclear will try again later.

loosebazooka:
picked a different security-snapshot, maybe it's not borked like the other one?

loosebazooka:
holy crap, we did it!

PR #1448, opened 2023-10-29
resolves #1409 to note that Python on Debian 12 is out of experimental

### LINKED ISSUE #1409: Python3 in debian12 (regular and plus)
This is a tracking issue for python3 support for debian12. This is a breaking change for people simply doing a `python3-debian11` -> `python3-debian12` migration but I think we should change this early and now while we still can.

Python will move out of experimental and we will create two new images:

`python3`: python and deps
`python3-plus`: `python3` plus ctypes support and a shell

lathama:
resolves #1409 to note that Python on Debian 12 is out of experimental while it is still built for Debian 11 from experimental. Another PR would need to move all of Python builds out of experimental.

Update:
* README
* Support Policy
* Examples

I am focusing on the Debian 12 Python topic, the Debian 11 Python is still in the experimental builds.

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,7 +102,7 @@ ENTRYPOINT "myapp"
 ```
 
 For the same reasons, if the entrypoint is set to the empty vector, the CMD command should be specified in `vector` form (see examples below).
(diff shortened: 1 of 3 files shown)


lathama:
I will look at the failed example after lunch.

loosebazooka:
I'm fixing examples right now

loosebazooka (approved):

lathama:
Removed the psutil example as it should maybe be redone when Debian 11 is moved out of experimental

loosebazooka (approved):

PR #1459, opened 2023-11-14
java21 improvements

loosebazooka:
- put temurin ca certs in debian common area and link temurin install ca certs to that, this provides some continuity if people want to install their own roots
- create intermediate base that is shared by java-base and java21, this allows us to remove libharfbuzz and deb-cacerts from the java21 install

### DIFF
--- a/java/BUILD
+++ b/java/BUILD
@@ -109,18 +109,12 @@ JAVA_VERSIONS_PER_DISTRO = [
     for distro in DISTROS
 ]
 
-DISTRO_SPECIFIC_LIBRARIES = {
-    "debian11": [
-        "libpcre3",
-    ],
-    "debian12": [
-        "libpcre2-8-0",
-    ],
-}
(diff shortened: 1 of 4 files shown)


thesayyn (approved):
LGTM

bobcallaway (approved):

Sineaggi:
If the plan is to slowly migrate to temurin based builds for all supported java versions, wouldn't it make sense to also publish the new build base? Or replace the existing build base with the new build base?

loosebazooka:
Yeah maybe, I'll still kinda of working this out.

Sineaggi:
@loosebazooka I checked out corretto, azul and upstream openjdk images, and they also bundle harfbuzz. Perhaps for the sake of devs wanting to create their own base images, I think we should consider replacing the existing build base with one that doesn't contain harfbuzz.

Issue #1393, opened 2023-09-07
support merged-usr filesystem layout


mythi:
**Describe the bug**
To follow the merged-usr approach, I'm installing a statically linked toybox (with sh) to `/install_root/usr/bin` and then

```
...
FROM gcr.io/distroless/static
COPY --from=builder /install_root /
```

Distroless `/bin` is a directory so I loose the typical shell shebangs (e.g., `/bin/sh`). I've tried to workaround it by creating the `/bin` symlink in `/install_root` but Buildkit is not happy about it since it cannot deal with `/bin/ already being a directory:

```
ERROR: failed to solve: cannot replace to directory /srv/docker/overlay2/nkktg2540y5ygxuk7fmauz7ec/merged/bin with file
```

(FWIW, it works with `DOCKER_BUILDKIT=0` but this is deprecated in favor of Buildkit).

**To Reproduce**
see the description.

**Expected behavior**
"merged `/usr` directories scheme, i.e. the `/{bin,sbin,lib}/` directories becoming symbolic links to `/usr/{bin,sbin,lib}/`. "

**Console Output**
N/A

**Additional context**
"In February 2021, the Technical Committee has resolved that Debian 'bookworm' should support only the merged-usr root filesystem layout, dropping support for the non-merged-usr layout."

There seems to be strong connection between Debian 12 and the latest distroless so I though this could apply.

loosebazooka:
Hrmm yeah I think we can adjust for this by creating those symlinks. Not sure what's creating that bin directory though

mythi:
> Distroless `/bin` is a directory so I loose the typical shell shebangs (e.g., `/bin/sh`). I've tried to workaround it by creating the

offtopic: I noticed that the `:debug` images fail on this. What triggered this issue originally was that I'm installing toybox to be able to run simple init scripts using initcontainers. Looks like the debug images (with busybox) cannot be used for this purpose either. (fixing `:debug` shell shebangs is separate from fixing this also).

mythi:
> Hrmm yeah I think we can adjust for this by creating those symlinks. Not sure what's creating that bin directory though

I looked into this today and my current thinking is that this is not going to be fully possible even with `debian12`. 

I built two new layers:
```
diff --git a/base/base.bzl b/base/base.bzl
index b4b602e..8682a3c 100644
--- a/base/base.bzl
+++ b/base/base.bzl
@@ -96,6 +96,8 @@ def distro_components(distro):
[...]

PR #1477, opened 2023-12-07
support usrmerged /bin and /sbin

mythi:
Distros are adopting the merged /usr directories scheme where the /{bin,sbin,lib}/ directories are made symbolic links to /usr/{bin,sbin,lib}/.

There are use cases where this could be justified for Distroless as well. For example, a statically built Busybox/Toybox w/ built-in shell can be installed to /usr/bin and still have the typical shellscript shebangs (/bin/sh or /bin/bash) working.

However, Debian 11/12 base packages (base-files, libc6 etc) do not support the merged /usr so it's not very straightfoward to enable for Distroless.

One approach to support this for /bin/ and /sbin is to create a layer with whiteouts and symlinks. Tests are also added to ensure any Distroless image flavor does not overwrite the symlinks created in that lower layer.


mythi:
> But we'd start to get into very strange territory.

Yup, agreed. I'm currently testing one idea on my side to workaround the limitation so maybe I'm able to sort this out without any Distroless changes needed after all. Just wanted to share this draft since I had it ready.

mythi:
For me, `#!/usr/bin/env bash` worked so I think I'll just close this one. Perhaps it's fine to keep #1393 open for tracking purposes since I think the ask makes sense.

loosebazooka:
Yeah unfortunately it's a rather annoying problem given our build

PR #1497, opened 2024-01-16
Update Jetty to latest 9.* version.

tim-fdc:


loosebazooka:
it seems fine, but I don't know if we will keep building the jetty image after debian11

tim-fdc:
> it seems fine, but I don't know if we will keep building the jetty image after debian11

Thanks for the info and it is not unexpected. @loosebazooka I needed to update the `sha256`

loosebazooka (approved):

PR #1511, opened 2024-02-02
Remove action for creating pr and use gh cli instead

joycebrum:
Related to #1506 discussion

I've used sigstore similar PR as reference https://github.com/sigstore/root-signing/pull/1097

It is correctly creating the PR https://github.com/joycebrum/distroless/pull/2 (I've forced a change just for it to have something to commit), although it created with my user because I've provided the AUTH_KEY.


joycebrum:
I noticed it will now fail if there is nothing to commit: https://github.com/joycebrum/distroless/actions/runs/7756341825/job/21153447586#step:6:14 , do we want it to fail if there is nothing to update or should I handle this error to be a success instead (with a warning or log explaining what happened perhaps)?

bobcallaway:
> I noticed it will now fail if there is nothing to commit: https://github.com/joycebrum/distroless/actions/runs/7756341825/job/21153447586#step:6:14 , do we want it to fail if there is nothing to update or should I handle this error to be a success instead (with a warning or log explaining what happened perhaps)?

`git commit` has an option `--allow-empty` but I'm not sure that's the right thing to do here - if there's nothing to commit, should we really just gracefully exit the workflow?

loosebazooka:
I think we can check the diff and if nothing has changed, end the job successfully there?

joycebrum:
> I think we can check the diff and if nothing has changed, end the job successfully there?

Agreed, no need to get failures when simply there is nothing to be done. I'll do that.

joycebrum:
After struggling a little with it, I was able to ensure some of the features the actions provided:

1. It won't break if there is nothing to commit 
https://github.com/joycebrum/distroless/actions/runs/7917065811
image

2. It can successfully create the pull request if there are changes and no PR exists https://github.com/joycebrum/distroless/actions/runs/7917017677

3. It can successfully update an existing PR 
https://github.com/joycebrum/distroless/actions/runs/7917040800
image

PR created and updated https://github.com/joycebrum/distroless/pull/7

.github/workflows/update-deb-package-snapshots.yml
```diff
@@ -29,13 +29,43 @@ jobs:
...
         done
         bazel run @//private/pkg/test/oci_image:test_sboms
-    
+
+    - name: Create commits
+      run: |
+        # Check work tree and end successfully if empty
+        if [ -z "$(git status --porcelain)" ]; then
+          echo "Nothing to commit";
+          exit 0;
```
loosebazooka, line 38: feel like this should set an output, and the step `Create Pull Request` should run conditionally on that using `if`, that would negate the need to check lines 57-60

joycebrum: yeah it makes sense. Will do that!

joycebrum: Done. I've also tested if the if clause is working properly with https://github.com/joycebrum/distroless/actions/runs/8237496450/job/22526374534 and https://github.com/joycebrum/distroless/actions/runs/8237449333

joycebrum: @loosebazooka PTL

PR #1524, opened 2024-02-21
Update rust example

jtagcat:
Current rust builds want newer glibc only present in debian12, else CMD errors in runtime.

distroless/cc: README: Any other tags are considered deprecated and are no longer updated → use distroless/cc-debian12

build-env: lessen cargo pin to only major version. There is no practical need for it to be pinned to patch.

### DIFF
--- a/examples/rust/BUILD
+++ b/examples/rust/BUILD
@@ -12,7 +12,7 @@ package(default_visibility = ["//visibility:public"])
 rust_image(
     name = "rust_example",
     srcs = ["src/main.rs"],
-    base = "//cc:cc_root_amd64_debian11",
+    base = "//cc:cc_root_amd64_debian12",
     tags = [
         "amd64",
         "manual",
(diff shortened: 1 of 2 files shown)


jtagcat:
Do note I have no clue about Bazel, feel free to exclude the BUILD file change.

jtagcat:
Example of glibc errors:
```
./remindee-bot: /lib/x86_64-linux-gnu/libm.so.6: version `GLIBC_2.35' not found (required by ./remindee-bot)
./remindee-bot: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.32' not found (required by ./remindee-bot)
./remindee-bot: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.34' not found (required by ./remindee-bot)
./remindee-bot: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.33' not found (required by ./remindee-bot)
```

loosebazooka (approved):

PR #1525, opened 2024-02-22
feat: add support for arm architecture in python3 builds

zenire:
I would like to have ARMv7 support since not all devices are ready to support arm64. Therefore, I've added the architecture in the build process, the same way JAVA_ARCHITECTURES is used.


loosebazooka:
you need to do more than this to get the arm images. You have to update the debian package consumer -- debian_packages.yaml

loosebazooka:
Also, ensure that we are not building for old debian11 images, only debian12

PR #1531, opened 2024-02-26
feat: use cloudflare snapshot as the primary url

thesayyn:
Upon my experiments with this, after the first download from the cloudflare snapshot, the subsequent downloads were 30x faster.  According to the headers sent by Cloudflare proxy, the TTL for cache is `31536000` which is basically forever.

/ Risks
There are no risks to doing this as the fallback to origin server is still there, in case cloudflare mirror goes down (not likely unless they shut it down.). Also the deb packages are checked against the checksum so there's no risk of tampering.

This is going to take CI time down significantly.


thesayyn:
The difference in speed can be seen here; 

Origin server first: https://github.com/GoogleContainerTools/distroless/actions/runs/8023783624/job/21920961029#step:5:1
Cloudflare first; https://github.com/GoogleContainerTools/distroless/actions/runs/8056212440/job/22004830829?pr=1531#step:5:1

PR #1465, opened 2023-11-17
refactor: start using rules_distroless

thesayyn:
@loosebazooka please take your time reviewing and testing this as rules from rules_distroless might not be identical to their counterparts here. I'll push new commits instead of rebasing to make it easier for you to review them.

Changes 

- replace `locale` 
- replace `cacerts`
- replace `java_cacerts`
- remove rules_docker
- replace group passed and home
- create `common` directory to move locale & cacerts & group & passwd into single BUILD file for reuse.

PR #1533, opened 2024-02-28
Revert "refactor: start using rules_distroless"

loosebazooka:
Reverts GoogleContainerTools/distroless#1465


loosebazooka:
I think we unintentionally broke some stuff, reverting for now. We'll have to inspect permissions some other way (in home?)

clrxbl:
Would like to mention that this also broke the Java images completely:

```
{"msg":"exec container process /usr/bin/java: Permission denied","level":"error","time":"2024-02-28T09:50:56.021343Z"}
```

loosebazooka:
looks like our examples might need local builds to catch these failures

PR #1538, opened 2024-03-05
ci: implement image diff pipeline

thesayyn:
This implements an image diffing pipeline where images built with changes and checked against the images built at HEAD.

### DIFF
--- a/.github/workflows/buildifier.yaml
+++ b/.github/workflows/buildifier.yaml
@@ -2,31 +2,28 @@ name: Buildifier
 
 on:
   pull_request:
-    branches: [ 'main' ]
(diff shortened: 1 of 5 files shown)


loosebazooka:
I'm not fully sure how to process this error log, the report seems to be empty too? Is this just the local registry failure. Is it possible to differentiate between a failure and a diff?

thesayyn:
Oh no, this is not ready yet. As there's no diff, the report will be empty. error log is just there, maybe should rename it to verbose.log

thesayyn:
Busybox archives are the cause of these diffs.

loosebazooka (approved):

PR #1539, opened 2024-03-05
fix: make busybox reproducible

thesayyn:
With https://github.com/GoogleContainerTools/distroless/pull/1538 i have discovered that busybox archives are non-reproducible.  

Its uid/gid mtime and permissions are directly affected by Bazel and the Host. 

Screenshot 2024-03-05 at 3 54 35 PM

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -2,6 +2,22 @@ workspace(name = "distroless")
 
 load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
 
+# rules_distroless setup
+http_archive(
+    name = "rules_distroless",
+    sha256 = "4b6d6a4bd03431f4f680ff5f6feea0b8ccf52c0296a12818d2c9595392e45543",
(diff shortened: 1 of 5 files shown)


WORKSPACE
```diff
@@ -2,6 +2,22 @@ workspace(name = "distroless")
 
 load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
 
+# rules_distroless setup
```
loosebazooka, line 5: should this workspace file be in this change?

loosebazooka (commented):
lgtm except this WORKSPACE change?

WORKSPACE
thesayyn: Yea as it being Bazel lib which has the tar rule. I can bring in Bazel lib directly but I will change it once we start using rules distroless. Do you prefer I bring in Bazel lib directly?

loosebazooka: Oh I see, yeah this is fine.

loosebazooka (approved):

PR #1542, opened 2024-03-09
feat: replace debian_package_manager with apt

thesayyn:
This the effort for replacing debian package manager with https://github.com/GoogleContainerTools/rules_distroless/pull/16

And a follow up to #1541


.github/workflows/update-deb-package-snapshots.yml
```diff
@@ -15,27 +15,27 @@ jobs:
...
+      - name: Update snapshots and lock using "knife"
+        run: ./knife update-snapshots && ./knife lock
 
-    - name: Run update script
-      run: ./updateWorkspaceSnapshots.sh
+      - name: Run update sboms script
+        run: |
+          for i in $(seq 5); do
+            bazel --bazelrc=$GITHUB_WORKSPACE/.github/workflows/ci.bazelrc fetch //private/pkg/test/oci_image:test_sboms && break || sleep 20;
+          done
```
thesayyn, line 30: Auto format :(

README.md
```diff
@@ -217,16 +223,15 @@ Distroless tracks the upstream Debian releases, using [Github actions to automat
 
 ### Debug Images
 
-Distroless images are minimal and lack shell access.  The ```:debug``` image set for each language provides a busybox shell to enter.
+Distroless images are minimal and lack shell access. The `:debug` image set for each language provides a busybox shell to enter.
```
thesayyn, line 226: @loosebazooka we should probably run pre-commit on the whole repo and rebase here. Can you do that?

knife
```diff
@@ -0,0 +1,166 @@
+#!/usr/bin/env bash
```
loosebazooka, line 1: I think I would prefer these be separate scripts that had more descriptive names rather than knife?

loosebazooka (approved):
LGTM, we can make that knife change in a followup?

knife
thesayyn: naming is tbd, i can split them as well.

thesayyn:
> LGTM, we can make that knife change in a followup?

Yes, i'll remove it from here.

thesayyn:
Oh, actually we can't as there's no replacement for what debian_package_manger did to upgrade snapshots etc. now those logic lives in knife.

PR #1574, opened 2024-04-22
Make debian12 the default image

loosebazooka:
This should make any image without an explicit debian version now just point to debian12 (instead of debian11) images.

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -4,6 +4,8 @@ load("//base:distro.bzl", "DISTROS", "LANGUAGE_DISTROS")
 
 package(default_visibility = ["//visibility:public"])
 
+DEFAULT_DISTRO = "debian12"
+
 ## STATIC
 STATIC_VARIANTS = [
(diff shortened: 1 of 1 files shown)


BUILD
```diff
@@ -4,6 +4,8 @@ load("//base:distro.bzl", "DISTROS", "LANGUAGE_DISTROS")
 
 package(default_visibility = ["//visibility:public"])
 
+default_distro = "debian12"
```
loosebazooka, line 7: @thesayyn is this bazel normal convention? for variable naming?

loosebazooka:
oh these are just debian11 -> debian12 diffs, should be good to go.

BUILD
thesayyn: constants are usually uppercase.

loosebazooka: hah yeah thanks, I eventually found some docs.

bobcallaway (approved):

PR #1578, opened 2024-04-24
Add nodejs 22

omBratteng:
Adds nodejs 20, which is the new LTS version

### DIFF
--- a/.github/workflows/image-check.yaml
+++ b/.github/workflows/image-check.yaml
@@ -12,8 +12,6 @@ concurrency:
 jobs:
   diff:
     runs-on: distroless-ci-large-ubuntu-20.04 # custom runner most compatible with debian 11
-    permissions:
-      pull-requests: write
     steps:
       - uses: actions/checkout@v4
       - name: Set up Go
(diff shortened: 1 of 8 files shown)


.github/workflows/image-check.yaml
```diff
@@ -12,6 +12,8 @@ concurrency:
 jobs:
   diff:
     runs-on: distroless-ci-large-ubuntu-20.04 # custom runner most compatible with debian 11
+    permissions:
+      pull-requests: write
```
omBratteng, line 16: I added this as the workflow only gets `read` permissions if the author of a PR is outside the organisation. But the `peter-evans/create-or-update-comment` action needs `write` access to be able to update the PR.

loosebazooka: I don't think we should expand permissions. Don't worry about the diff image check.

omBratteng: reverted

smorimoto:
Can we move forward with this PR?

loosebazooka:
Oh yeah I'll look when I'm back at work tomorrow

smorimoto:
Thanks a lot!

loosebazooka (approved):

loosebazooka:
Thanks again @omBratteng for your commitment to the distroless node images!

PR #1582, opened 2024-05-01
fix: remove pkg_tar workaround from temurin archive

thesayyn:
Removes the workaround for executable files in temurin_archive and replaces it with `tar` from bazel-lib.  This PR introduces a new system dependency `awk` for the build. @loosebazooka is awk installed on cloudbuild runners?

Some diff is expected, because `tar` rule preserves both numeric and non-numeric owner for files and defaults to file mode on disk.

Also https://github.com/aspect-build/bazel-lib/pull/829 landed, i will remove genrule once there's a new release of bazel-lib.


loosebazooka:
@thesayyn 
seeing directories with odd permissions (from the diff report)
```
File     usr/lib/jvm/temurin21_jre_arm64/conf/security/policy/                                                Mode 0x1ed                                                    Mode 0x0
```
Current 0755
New 0000

which looks incorrect to me. Is that just what's provided by the temurin archive?

loosebazooka:
found some other issues here. Duplicate entries of the same file (ex:`/usr`) wont load up in docker.

thesayyn:
@loosebazooka okay i fixed the permission issues after a lengthy debug session; 

rest of the diff is harmless; 

```

🚧 Diffing gcr.io/distroless/java21-debian12:latest-amd64 against localhost:4564/stage/java21-debian12:latest-amd64

localhost:4564/stage/java21-debian12@sha256:9eb1a61b2196de0aa2a8984d986a4995d8b8c86306fe1a860d88ee72817e6694
TYPE     NAME                    INPUT-0                                                       INPUT-1
Layer    ctx:/layers-27/layer    length mismatch (328 vs 329)                                  
Layer    ctx:/layers-27/layer    name "usr/" appears 1 times in input 0, 2 times in input 1    
Layer    ctx:/layers-27/layer    name "usr/" appears 1 times in input 0, 2 times in input 1  
```

private/remote/temurin_archive.bzl
```diff
@@ -1,59 +1,64 @@
...
+# a symlink created in the original location for completeness.
+usr/lib/jvm/%s/lib/security/cacerts nlink=0 time=946684800.0 mode=777 gid=0 uid=0 type=link link=/etc/ssl/certs/java/cacerts
+"""
+
+AWK = """\
+{
+    sub("^"  "output/lib/security/cacerts", "./etc/ssl/certs/java/cacerts")
+    sub("^"  "output", "./usr/lib/jvm/%s")
+    sub(/time=[0-9\\.]+/, "time=946684800.0");
+    if ($1 ~ ".*legal/.*" || $1 ~ ".*conf/.*") {
```
thesayyn, line 17: This is all to match the previous permissions set by pkg_tar workaround, ideally we'd want to remove these.

loosebazooka:
I'll check the awk dependency.

loosebazooka:
kk awk appears to be on the cloud-build bazel image. So we should be good there.

private/remote/temurin_archive.bzl
loosebazooka: I don't really see where the executable bits are set? Are they all passed through?

thesayyn: yes, we preserve what is in the original executable, with some exceptions (to have no diff)

thesayyn:
this should be ready to land then?

private/remote/temurin_archive.bzl
loosebazooka: Sorry I actually meant to preserve all executables in this change. Are we intentionally removing executable bits to get no diff?

loosebazooka (approved):

private/remote/temurin_archive.bzl
thesayyn: The original archive already sets file permissions for executables correctly, so we just need to preserve those. 

However, for entries in `legal` and `conf`, we set some specific permissions so that they match the previous permissions set by pkg_tar.

thesayyn: Ideally at some point we should remove those logic from this file and have a diff, but i'd rather have that diff after we confirm this PR doesn't create problems for us.

PR #1597, opened 2024-05-16
Exclude node headers

omBratteng:


loosebazooka (approved):
So these are not necessary at runtime?

omBratteng:
> So these are not necessary at runtime?

I will do some testing, but I am pretty sure these are not needed. It is just `.h` header files. They could probably be necessary when you install packages that needs compilation, but without npm (or shell), it's not possible

omBratteng:
@loosebazooka I've changed the glob in pkg_tar for node, instead of excluding all the files, I am only including the LICENSE file and node binary.

Seeing as @pkwarren just pull the binary from the image in this PR https://github.com/bufbuild/plugins/pull/1118, I think it should work fine just include the binary and license.

loosebazooka:
@omBratteng I'm going to do a little research before I merge this. I'm afraid of breaking all the node users.

omBratteng:
That's fine, and makes sense

ianlewis:
> @omBratteng I'm going to do a little research before I merge this. I'm afraid of breaking all the node users. But it looks fine otherwise. I will most like merge this tomorrow.

I think this should be fine but agree that it should be tested. I'm not sure how distroless expects folks to include projects with dependencies that were compiled but I assume it would use the [JavaScript build rules](https://bazel.build/docs/bazel-and-javascript) which wouldn't use the runtime that's included in the resulting image.

omBratteng:
> > @omBratteng I'm going to do a little research before I merge this. I'm afraid of breaking all the node users. But it looks fine otherwise. I will most like merge this tomorrow.
> 
> 
> 
> I think this should be fine but agree that it should be tested. I'm not sure how distroless expects folks to include projects with dependencies that were compiled but I assume it would use the [JavaScript build rules](https://bazel.build/docs/bazel-and-javascript) which wouldn't use the runtime that's included in the resulting image.

Usually the way I've done it, is not with bazel. So I just use official node images with debian to install dependencies and build the code, then copy the node_modules directory and any other built files.
In 99% of cases, that does include everything you might need.

loosebazooka:
yeah we're not expecting anyone to use distroless images as build time images, they should be using fully featured build containers -- distroless is exclusively for runtime. An alternative is the debug images keeping those tools and stuff. I'll go ahead and merge this.

PR #1601, opened 2024-05-25
Add workspace references to deb repositories rule

implr:
This allows pulling those dependencies into other workspaces importing this repository, eg. for creating derived images with Bazel.

Tested on amd64 and aarch64.


loosebazooka:
I don't think we want people to depend directly on the distroless build. If you want to extend distroless, try using rules_distroless.

PR #1615, opened 2024-07-01
chore: update snapshot location

thesayyn:
Updates snapshot urls to be `https://snapshot-mlm-01.debian.org` instead of `snapshot-cloudflare.debian.org` as the latter does not receive new snapshots anymore.


loosebazooka:
I think we need to add back in the failure resiliency for snapshots repos

thesayyn:
> I think we need to add back in the failure resiliency for snapshots repos

That's too bad. It will lead to longer CI times.  I'll that back. 

We should probably ask debian snapshot team if they are going to introduce cloudflare frontend to the new service.

loosebazooka (approved):

PR #1619, opened 2024-07-04
Add retries to locking and update snapshots

loosebazooka:
add some more resilience when using flaky snapshot server


thesayyn (approved):

loosebazooka:
Ugh forgot to regen the sboms. Will redo after the beach ⛱️

Issue #1620, opened 2024-07-04
Proposal to Add Apache Tomcat as a New Option for Distroless Container


ptrecenti:
Dear Maintainers,

I am writing to propose the addition of Apache Tomcat as a new option for the distroless container images provided by GoogleContainerTools. Apache Tomcat is a widely used and robust Java servlet container that many developers rely on for their web application deployments. Integrating Tomcat into the distroless images would offer several advantages, including enhanced security, reduced image size, and alignment with best practices for containerized applications.

**Rationale**:

**Security**: Distroless images are designed to contain only the necessary dependencies, reducing the attack surface by excluding unnecessary binaries and package managers. Incorporating Tomcat into this model would significantly enhance the security posture of applications that rely on Tomcat.

**Efficiency**: The lean nature of distroless images results in smaller image sizes and faster startup times. By offering a Tomcat variant, users can benefit from these efficiencies without compromising on the features and stability provided by Tomcat.

**Best Practices**: Adopting distroless images with Tomcat aligns with modern DevOps and cloud-native best practices, encouraging the use of minimal base images for production deployments.

**Implementation Details:**

**Base Image:** The proposed image would be built upon the existing Java distroless base image, leveraging the established and optimized environment.
**Tomcat Version:** To start, the latest stable version of Apache Tomcat (e.g., Tomcat 10) would be included. Future updates would follow the standard release cycle.
**Configuration:** The image should allow for easy configuration through environment variables and support mounting custom configuration files.
**Compatibility:** The image should be compatible with common deployment platforms like Kubernetes and support integration with CI/CD pipelines.

**Additional Information:**

I have initiated work on this proposal in my forked repository, which can be found at [ptrecenti/distroless](https://github.com/ptrecenti/distroless). This fork includes preliminary work on integrating Tomcat with the distroless images and can serve as a starting point for further development and collaboration.

**Benefits to the Community:**

**Wider Adoption:** Providing a distroless Tomcat image would attract users who are currently deploying traditional Tomcat images, driving wider adoption of distroless images.
**Community Trust:** As Tomcat is a trusted and established server, its inclusion would add to the credibility and utility of the distroless project.
**Ease of Use:** Users can benefit from the reduced complexity and enhanced security without needing to build custom images, streamlining their workflows.

[...]

PR #1621, opened 2024-07-04
Proposal to Add Apache Tomcat as a New Option for Distroless Container

ptrecenti:


loosebazooka:
see: https://github.com/GoogleContainerTools/distroless/issues/1620#issuecomment-2209566985

PR #1638, opened 2024-08-08
Stop failing image-diff on PRs from forks

loosebazooka:


loosebazooka:
diff is from a timing issue with a build going on right now. we're fine

omBratteng (approved):

thesayyn:
Hmm, i am confused a little. Does the change makes Diff pipeline to work on forks only?

loosebazooka:
It only reports in a comment on non-forks

thesayyn (approved):

PR #1649, opened 2024-08-27
Add temurin update script

loosebazooka:
This needs to be integrated as an action, but should run daily so we can pick up the latest temurin builds for java21 at the moment and other java builds in the future

It was an interesting use of a day.


loosebazooka:
@thesayyn perhaps there is a more bazel-y way to do this? But this covers the process.

update_java_archives.sh
```diff
@@ -0,0 +1,95 @@
+#!/usr/bin/env bash
+set -o pipefail -o errexit -o nounset
+
+# Copyright 2024 Google Inc. All rights reserved.
+
+# Licensed under the Apache License, Version 2.0 (the "License");
```
thesayyn, line 6: Can this go into the knife script?

thesayyn:
Also this could definitely be a bazel script instead...

thesayyn (approved):
Sweet

PR #1656, opened 2024-09-03
remove debian11 builds and references

loosebazooka:
This is planned deprecation of debian11: https://github.com/GoogleContainerTools/distroless/blob/main/SUPPORT_POLICY.md

We're a few months late on this but our updates have also been failing because debian11 snapshots are missing updates for architectures (s390x at least for now)


loosebazooka:
java diffs are because layers are swapped at ./java/BUILD:154 (https://github.com/GoogleContainerTools/distroless/pull/1656/files#diff-737cc99926a55f4457d01dc2f990efc8c88eb814d7a7a9576726f3c2749ed77aR154)

bobcallaway (approved):

PR #1668, opened 2024-09-13
fix build

loosebazooka:
cloud build being wonky with an old version of bazel


loosebazooka:
Ignore the diff, it's cause the last cloud build failed

bobcallaway (approved):

PR #1670, opened 2024-09-13
fix: register a default toolchain with root ignored

thesayyn:
This ought to fix the issue


MODULE.bazel
```diff
@@ -12,6 +12,13 @@ bazel_dep(name = "rules_rust", version = "0.49.1")
 bazel_dep(name = "container_structure_test", version = "1.16.0")
 bazel_dep(name = "rules_oci", version = "1.7.5")
 bazel_dep(name = "rules_distroless", version = "0.3.7")
+bazel_dep(name = "rules_python", version = "0.35.0")
+
+### PYTHON ###
```
loosebazooka, line 17: I think this needs a (or something)
```
python = use_extension("@rules_python//python/extensions:python.bzl", "python")
```

thesayyn: Right, done

loosebazooka (approved):

Issue #1661, opened 2024-09-05
`/home` incorrectly has ownership changed to `nonroot:nonroot` (`65532:65532`)


polarathene:
**Describe the bug**

Images (at least Debian 12 variants) seem to add `/home/nonroot` layer that modifies the ownership of `/home`:

> ![image](https://github.com/user-attachments/assets/8de621fa-980e-412d-8342-72da350360c2)

Presumably that was unintentional and only meant to be applied to `/home/nonroot`?

**To Reproduce**

You can inspect the responsible layer where this change has been highlighted above via the `dive` tool:

```bash
docker run --rm -it \
  -v /var/run/docker.sock:/var/run/docker.sock \
  wagoodman/dive:latest \
  gcr.io/distroless/static-debian12:latest
```

While I've not done a diff of the file content, it does look rather similar to `gcr.io/distroless/static-debian12:nonroot`. That might explain why the `/home/nonroot` user is created?

`docker inspect gcr.io/distroless/static-debian12:nonroot` does show metadata differs, such as the default working directory and user to run as.

**Expected behavior**

`/home` should keep it's `0:0` ownership?

loosebazooka:
interesting, lemme go see

loosebazooka:
@thesayyn could this be happening in rules_distroless?

thesayyn:
This is probably due to rules_distroless home rule. It creates the home directory just in case which is not needed in this case

polarathene:
@thesayyn I don't mind the `/home/nonroot` directory being created, but `/home` itself should not have the UID/GID for `nonroot`? So perhaps the rule should keep `/home` as `0:0`?

thesayyn:
According to https://github.com/GoogleContainerTools/rules_distroless/blob/35a7d5a37b34e68f1d58d7e452147afe941f3e5a/distroless/private/home.bzl#L18-L26

home rule only creates subdirectories under /home with desired permissions, so i don't know where the problem is.

polarathene:
> so i don't know where the problem is.

I'm not familiar with the build system itself, but I do know if using `COPY --link` in a `Dockerfile` this has the effect of treating the source as `FROM scratch` stage of it's own implicitly that is then a `COPY` to the stage invoking `--link`. The `FROM scratch` has that source path copied to it at the target path.

[...]

PR #1704, opened 2024-10-25
fix: the ownership of /home should be 0:0

hunshcn:
fix https://github.com/GoogleContainerTools/distroless/issues/1661


loosebazooka (approved):

loosebazooka:
Oh nice. I guess you found it. Thanks for this.

PR #1738, opened 2025-01-22
refactor: deduplicate images before pushing and signing

thesayyn:
Following changes has been made to speed up pushing and signing of images. 

- Push and sign only the first occurrence of an image,
- Use `crane copy` to duplicate images at remote registry for the same images but different tags,

As a result;

- Drops the number of pushes & sign invocation 856 to 236 
- And introduces a 600 new really efficient copy at the remote tag operation. 

This should make distroless distribution pipeline significantly faster and more efficient.


loosebazooka:
shoot @thesayyn can you rebase this, I'll merge it

loosebazooka (approved):

PR #1769, opened 2025-03-17
Fix dedup logic to handle different repos from the same image

loosebazooka:
gcr.io/distroless/(cc|cc-debian12):<tag>

since bazel doesn't allow dicts of dicts of lists in attrs this looks a bit crazy


loosebazooka:
These diffs appear to be from the images we were failing to tag

bobcallaway (approved):

Issue #816, opened 2021-09-06
Add mime.types to static image


Xjs:
Could the file `/etc/mime.types` be added to the static image? This had been requested in #256 but was apparently overlooked. The file `/etc/mime.types` is provided by the Debian package `media-types`, cf. https://packages.debian.org/sid/media-types/filelist

Thanks :)

loosebazooka:
Is this only required for go? As described in https://github.com/GoogleContainerTools/distroless/issues/256#issuecomment-451569173

Xjs:
My personal use-case is Go, if that's what you're asking. Generally, I don't know which applications consume this file, but I believe it is assumed present in many Linux distributions.

loosebazooka:
So the point of this file is to point mime-types to applications (is it?)? In the case of distroless does it have any use? The only application on the container is yours?

Does golang fail critically if this file is missing?

Xjs:
The point of the file is to associate file extensions and MIME types. Go uses the file in package mime, in my example I was using https://pkg.go.dev/mime#TypeByExtension to determine the MIME type to declare in an e-mail header for an attachment. 

Go applications that use package mime are reduced to the built-in table (currently https://cs.opensource.google/go/go/+/refs/tags/go1.17.1:src/mime/type.go;l=60) for this kind of lookups. This is no hard failure. Applications can register specific associations themselves as well (this is what I'm currently doing as a workaround). 

The only application on the container is mine, yes. It's a statically linked Go binary. I'm using the image through https://github.com/bazelbuild/rules_docker#go_image, which makes it comparatively hard to include the file myself (it would require me to maintain a version of the file, or pin a specific e. g. Debian package source in my WORKSPACE as well. Since distroless already maintains Debian package source versions it would be simpler overall to include the file here.

FTR: I can live with a rejection of this issue, I just thought it had been overlooked in #256 and wanted to reopen the discussion.

0x91:
hello, don't want to necro a very old issue but is there any appetite to reopen this discussion?

this is still a problem for Go applications and i think one could make the argument that having mime.types inside the image offers similar utility to things like tzdata

vincentbernat:
No need to explicitly call the mime package to run into this issue. Just serving files with the builtin HTTP server returns incorrect MIME types.

loosebazooka:
Can someone find the deb with these files in it?

loosebazooka:
this? https://packages.debian.org/bookworm/media-types

[...]

PR #1832, opened 2025-08-04
Add /etc/mime.types to static

loosebazooka:
This should apply the file to all images

fixes: #816


loosebazooka:
just updated tests

Issue #1269, opened 2023-03-17
feat: support riscv64


ernado:
# riscv64 architecture support for static images

Add support for riscv64 architecture, so riscv64 support can be added for dependant projects that are building on distroless.

> riscv64 should be limited to only the static images and dependencies. This will reduce impact on the build.

The static images support would be enough.

### Current issues
```
..URL: https://snapshot.debian.org/archive/debian/20230314T211539Z/dists/bullseye/main/binary-riscv64/Packages.xz, Status: 404 NOT FOUND
panic: failed (status: 404) to fetch remote file: "https://snapshot.debian.org/archive/debian/20230314T211539Z/dists/bullseye/main/binary-riscv64/Packages.xz"
```

There are no packages snapshot for riscv64, so `updateWorkspaceSnapshots.sh` will fail.

### Reference

- https://wiki.debian.org/RISC-V (in-progress debian riscv64 port)
- Debian-riscv [mailing list message](https://lists.debian.org/debian-riscv/2023/03/msg00001.html) from me
- https://github.com/carlosedp/riscv-bringup/ (attempt to bring riscv64 support to container infra from 2019 with lots of additional info)
- https://github.com/etcd-io/etcd/pull/15490 (etcd PR where distroless image is required)
- https://github.com/GoogleContainerTools/distroless/pull/1268 (my failed attempt to add riscv64 to distroless)

### Upd

I've made a fork with CI/CD for building [distroless riscv64 images](https://github.com/go-riscv/distroless/pkgs/container/distroless%2Fstatic-unstable) on Debian Sid (unstable):

https://github.com/go-riscv/distroless

Looks like I will stick to it until Debian Trixie is released, where riscv64 will be (hopefully) stable, I think it will be in 2025.

ernado:
I've made another attempt to support riscv64, but it requires medium size non-trivial refactoring: https://github.com/GoogleContainerTools/distroless/pull/1270

mengzhuo:
ping
Debian 13 has released
https://github.com/GoogleContainerTools/distroless/issues/1826

Mcfloy:
Unfortunately linked to #1871 and GitHub will think this will resolve it but this won't. As supporting a new architecture will have a bigger impact than just adding it just for Java, another PR will be created after aiming to support: `static`, `base`, `base-nossl`, `cc` and `java25`.

loosebazooka:
[...]

PR #1850, opened 2025-08-22
Add builds for debian13 static, base, base-nossl

loosebazooka:
Also mention they are in preview in README

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -1,5 +1,5 @@
 load("//:checksums.bzl", "ARCHITECTURES", "BASE_ARCHITECTURES")
-load("//base:distro.bzl", "DISTROS")
+load("//base:distro.bzl", "DISTROS", "PREVIEW_DISTROS")
 load("//private/oci:defs.bzl", "sign_and_push_all")
 load("//nodejs:node_arch.bzl", "node_arch")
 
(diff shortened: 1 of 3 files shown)


gfrankliu:
Shouldn't we upgrade rules_distroless and enable mergedusr for trixie as mentioned https://github.com/GoogleContainerTools/distroless/blob/main/private/repos/deb/deb.MODULE.bazel#L21

mengzhuo:
@loosebazooka Would you mind adds riscv64 architecture into 13? Thanks.
Ref https://github.com/GoogleContainerTools/distroless/issues/1269

thesayyn (approved):

thesayyn:
There is https://registry.bazel.build/modules/rules_distroless/0.5.3 which can be bumped here.

loosebazooka:
@thesayyn I can't seem to bump rules_oci without some modification about where crane is being pulled from

loosebazooka:
> There is https://registry.bazel.build/modules/rules_distroless/0.5.3 which can be bumped here.

this was bumped in #1853

loosebazooka:
for mergedusrfs, I think we need to
- [ ] add base symlinks
- [ ] check that no layer overwrites these symlinks

armingerten:
Any updates on this? :)

floriankoch:
@loosebazooka can i help with somethig?

mprimeaux:
@bobcallaway When will this PR be merged? Happy to contribute where needed.

loosebazooka:
Updated this to include merged usrfs, I think we can do this build and lets users try it out. I checked manually that the build doesn't overwrite the symlinks anywhere. This may be different with the language images, but we can start from here.

@bobcallaway @thesayyn , this is ready for review now?

bobcallaway (approved):

c-zuniga-hiwin:
Thanks for adding Debian 13 builds for static/base! Do you have an ETA for cc and the language images (Python, Java, Node)?

loosebazooka:
no ETA, but hopefully soon

PR #1855, opened 2025-08-27
Move java21 over so we can add java17 for debian13

loosebazooka:


MODULE.bazel
```diff
@@ -36,25 +36,13 @@ go_deps.module(
 )
 use_repo(go_deps, "com_github_spdx_tools_golang")
 
-### JETTY ###
```
loosebazooka, line 39: we don't do jetty builds anymore

knife
```diff
@@ -136,11 +136,14 @@ function cmd_lint () {
...
-    old_version=$(get_java_version)
-    generate_java_archives > private/repos/java.MODULE.bazel
-    new_version=$(get_java_version)
+    source knife.d/update_java_archives.sh
+    old_version=$(get_java_version 21)
+    generate_java_archives 21 > private/repos/java_temurin/java_21.MODULE.bazel
+    new_version=$(get_java_version 21)
     update_test_versions_java21 $old_version $new_version
+
+    # java 17 set up
```
loosebazooka, line 145: we don't actually have java17 builds yet for debian13, and we need to update this when we do.

knife
```diff
@@ -177,6 +180,9 @@ function cmd_test () {
     echo ""
 
     local arch=$(uname -m)
+    if [ ${arch} == "x86_64" ]; then
```
loosebazooka, line 183: uname works funny, so convert to debian arch names

private/repos/java_temurin/java.MODULE.bazel
```diff
@@ -0,0 +1,22 @@
+"repositories for java"
```
loosebazooka, line 1: @thesayyn can't tell if this is bazel idiomatic

thesayyn: yes this is encouraged to keep main module.bazel lean.

thesayyn (approved):

PR #1890, opened 2025-10-13
add compression to cc

### LINKED ISSUE #1887: compression lib for cc image debian13
> @loosebazooka - Thanks for knocking this out! I gave `gcr.io/distroless/cc-debian13` a shot. When executing the resulting image, I ran into errors like this: `libzstd.so.1: cannot open shared object file: No such file or directory`. I was able to `apt install` both`zlib1g` and `libzstd1`,  copy both libz.so.1 and libzstd.so.1 into the final image with cc-debian13 as the base and make things work. With `gcr.io/distroless/cc-debian12`, the image functioned without separately copying over those shared libraries.
> 
> I have no idea whether this is an expected change or not. But I thought I'd report it in case it was unintended! Thanks again
> 
> 
> ```
> FROM rust:1.90.0AS build
> RUN apt-get update && apt-get install -y --no-install-recommends zlib1g libzstd1
> 
> # Other stuff...
> 
> FROM gcr.io/distroless/cc-debian13
> COPY --from=build /lib/*/libz.so.1 /lib/
> COPY --from=build /lib/*/libzstd.so.1 /lib/
> 
> # Other stuff...
> ```
>  

 _Originally posted by @kyle-leonhard in [#1851](https://github.com/GoogleContainerTools/distroless/issues/1851#issuecomment-3383501287)_

loosebazooka:
fixes #1887


loosebazooka:
Oh right I changed the order of some things

PR #1896, opened 2025-10-22
Rework build into reusable pieces

loosebazooka:
- only static, base and cc
- standardize their debug image naming (`//image:image_debug` instead of `//image:debug`)

## Summary
First step at organizing the build into reusable parts. Still need to rework java, node, python.
It might even make sense to pull out the deb list from each and push that through the defs in the BUILD, but this should be a decent start.


loosebazooka:
~~hrmm... we shouldn't have image changes~~

oh it's the order of busybox because we build base_debug off static_debug now (so we only see diffs in debug images)

PR #1911, opened 2025-10-29
rework nodejs build, add deb13

loosebazooka:
same as #1902 #1896  for nodejs

### DIFF
--- a/nodejs/BUILD
+++ b/nodejs/BUILD
@@ -1,107 +1,33 @@
-load("@container_structure_test//:defs.bzl", "container_structure_test")
-load("@rules_oci//oci:defs.bzl", "oci_image", "oci_image_index")
-load("@rules_pkg//:pkg.bzl", "pkg_tar")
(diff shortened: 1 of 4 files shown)


nodejs/BUILD
```diff
@@ -1,107 +1,33 @@
...
-    "_debug",
-]
+NODEJS_DISTROS = DISTROS + PREVIEW_DISTROS
 
-USER = [
-    "root",
-    "nonroot",
-]
+# NODEJS_ARCHITECTURES = ARCHITECTURES
+# using node_arch for now, but we can probably get rid of that
```
omBratteng, line 12: Once we drop support for 20 and 22, we can drop node_arch in favor of `BASE_ARCHITECTURES + ["s390x", "ppc64le"]` (similar to ARCHITECTURES, just without `arm`)

loosebazooka: I think maybe we don't need a centralized ARCH var that every different package munges with. I'll try to clean that part up too in the future.

PR #1920, opened 2025-11-04
Move configuration around

loosebazooka:
- more convenient changes
- root BUILD is tied to individual image config
- add action to detect changes to root BUILD config


loosebazooka:
alright these all look like ordering changes, which is fine.

thesayyn (approved):

PR #1929, opened 2025-11-17
PoC for command injection in buildifier workflow

Dragon-Room:


loosebazooka:
This seems like spam, if you have an actual something you'd like to show, please use responsible security disclosure practices on github.

Issue #1851, opened 2025-08-22
debian 13 to do


loosebazooka:
- [ ] merged usr fs
   - only for debian13 (not 12)
   - ensure no overwrites, all images should have a test that checks for overwrites
   - update any paths
- [ ] mtime -> 0 (we seem to use a variety of numbers all over the place to minimize diffs)

huakaibird:
when 13 will be released?

loosebazooka:
I don't have a timeline, hopefully in the next few weeks.

IanDavey:
Hello,

What is the timeline for release here?  My employer is waiting on a production deployment.

Thanks.

maxime-perocheau-hs:
@loosebazooka is there anything left outstanding for this? Happy to help if I can, would love to see Debian 13 images released!

jayvdb:
Appears to be happening at https://github.com/GoogleContainerTools/distroless/pull/1850

harryzcy:
> Appears to be happening at [#1850](https://github.com/GoogleContainerTools/distroless/pull/1850)

No, it only adds Debian 13 to root-level bazel build. It does not address any of the TODOs listed in this issue.

nappa85:
Sorry to bother, but any news on this? It's becoming annoying since most of build systems have already increased glibc version and builds won't work with debian 12

loosebazooka:
I believe @thesayyn is working on it this week

ndthayer:
@thesayyn , @loosebazooka : Is there an ETA you can share on the debian13 distroless base image being GA?

loosebazooka:
okay we're buliding `static`, `base`, `base-nossl` debian13 images now -- disclaimer here: https://github.com/GoogleContainerTools/distroless/?tab=readme-ov-file#debian-13-preview

kyle-leonhard:
@loosebazooka - Thanks for knocking this out! I gave `gcr.io/distroless/cc-debian13` a shot. When executing the resulting image, I ran into errors like this: `libzstd.so.1: cannot open shared object file: No such file or directory`. I was able to `apt install` both`zlib1g` and `libzstd1`,  copy both libz.so.1 and libzstd.so.1 into the final image with cc-debian13 as the base and make things work. With `gcr.io/distroless/cc-debian12`, the image functioned without separately copying over those shared libraries.

I have no idea whether this is an expected change or not. But I thought I'd report it in case it was unintended! Thanks again


```
[...]

Issue #1826, opened 2025-07-23
Support for Debian13


huww98:
Debian 13 is [just around the corner](https://lists.debian.org/debian-devel-announce/2025/07/msg00003.html). Any plans to support it? Do you welcome/need community contribution on this?

loosebazooka:
Hello, we already have someone provisioned for this, but would always like extra hands to review/test as we figure out the new deb13 builds.

gfrankliu:
Hopefully merged-usr is finally supported as discussed in https://github.com/GoogleContainerTools/distroless/issues/1393

loosebazooka:
> Hopefully merged-usr is finally supported as discussed in #1393 

Yeah that's the goal

Ardakilic:
Hello,

Debian 13 is out, so wanted to ping here just in case:

https://9to5linux.com/debian-13-trixie-is-now-available-for-download-heres-whats-new

davinkevin:
If required, we can run our product CI on top of those images and report issue if we find some. 

We mainly use `java21`, `base-nossl` and `cc`.

harryzcy:
This start to become a blocking issue as many language images are using Debian 13, which use glibc 2.41. But Debian 12 use glibc 2.36.

harryzcy:
> > Hopefully merged-usr is finally supported as discussed in [#1393](https://github.com/GoogleContainerTools/distroless/issues/1393)
> 
> Yeah that's the goal

The functionality is added to rules_distroless [already](https://github.com/GoogleContainerTools/rules_distroless/commit/cd8abc5d1c25d4cb87e3f48390514a946aa715eb), but not released yet. Can any maintainer cut a new release for it?

gfrankliu:
> > > Hopefully merged-usr is finally supported as discussed in [#1393](https://github.com/GoogleContainerTools/distroless/issues/1393)
> > 
> > 
> > Yeah that's the goal
> 
> The functionality is added to rules_distroless [already](https://github.com/GoogleContainerTools/rules_distroless/commit/cd8abc5d1c25d4cb87e3f48390514a946aa715eb), but not released yet. Can any maintainer cut a new release for it?

There is [a ticket](https://github.com/GoogleContainerTools/rules_distroless/issues/174) open but that project seems slow nowadays.

reneleonhardt:
[...]

PR #1933, opened 2025-11-24
fix: Add zlib/libzstd to base-debian13 image

rinsuki:
because OpenSSL 3.2+ uses those libraries to support TLS Certificate Compression, OpenSSL now requires zlib and libzstd (see https://packages.debian.org/trixie/libssl3t64 ).

without this change, user need to add zlib/libzstd manually in their Dockerfile for use OpenSSL, otherwise they would get errors like `libzstd.so.1: cannot open shared object file: No such file or directory`.

see:
* https://salsa.debian.org/debian/openssl/-/commit/ca7e7b08746d408da112fe63d5a3bf941c345ec0

(note: Debian disables brotil compression so we don't need brotil libraries at this time, see https://salsa.debian.org/debian/openssl/-/commit/26109eddd699cce5c3aecb42ec1c5178cbde7d84 )

This should fix https://github.com/GoogleContainerTools/distroless/issues/1887 and other comments, e.g.

* https://github.com/GoogleContainerTools/distroless/issues/1826#issuecomment-3506839428
* https://github.com/GoogleContainerTools/distroless/issues/1851#issuecomment-3383501287

p.s. I couldn't run the full test suite since I dont have a enough free space to run that.

### DIFF
--- a/base/README.md
+++ b/base/README.md
@@ -21,6 +21,11 @@ of the packages in `gcr.io/distroless/static`, and
 
 * glibc
 * libssl
+  * for Debian 13 or later, libssl dependencies (zlib, libzstd) are also included.
+
+Note: Debian 13 base image does not include [OpenSSL legacy algorithms](https://docs.openssl.org/3.5/man7/OSSL_PROVIDER-legacy/#operations-and-algorithms).
+
+If you want to use them, you will need to add `openssl-legacy-provider` yourself.
 
 ## Usage
 


rinsuki:
What I Tested:

Previously (`gcr.io/distroless/cc-debian13:nonroot@sha256:2e14174aa2ffeb547d1beaf8a6b72ba2721965be790c0f5541d699980dc64fa8`), my private app (which is built with Rust and uses OpenSSL) doesn't work (error: `error while loading shared libraries: libzstd.so.1: cannot open shared object file: No such file or directory`)

After change base image to this patch (`//cc:cc_nonroot_arm64_debian13`), my Rust app works fine without shared libraries error.

base/config.bzl
```diff
@@ -12,6 +12,9 @@ BASE_PACKAGES = {
     "debian13": [
         "libc6",
         "libssl3t64",
+        "libzstd1",
+        "zlib1g",
+        "openssl-provider-legacy",
```
loosebazooka: I'd prefer not to include legacy algorithms

rinsuki: done @ https://github.com/GoogleContainerTools/distroless/pull/1933/commits/b3bb52170d73f40be0a0b0c369257cadd03d92e7

loosebazooka:
Thanks! I'll merge this when tests pass

PR #1936, opened 2025-11-25
Add bun support and update nodejs

iamleniac:
- Added support for bun images
- Updated nodejs versions
- linter has changed some import ordering


loosebazooka:
I think this is unlikely to happen unfortunately. But there's nothing preventing you from creating and releasing these distroless images on your own. In the future to prevent doing a whole PR, please file an issue so we can discuss it there before putting in the work.

PR #1961, opened 2026-01-04
docs(python3): clarify Python version compatibility and unsafe site-p…

### LINKED ISSUE #1960: Python distroless: ABI mismatch errors when copying site-packages from builder (request clearer docs & versioned tags)
### Description

When using distroless Python images in a multi-stage Docker build, copying Python site-packages from a builder image frequently results in runtime failures such as:

SyntaxError: Non-UTF-8 code starting with '\x80' in file /usr/bin/python3

or:

No module named <dependency>

This typically occurs when:
- Dependencies are installed using Python X.Y (e.g., 3.12) in a builder image
- The distroless runtime image uses a different Python minor version (e.g., Debian 12 Python 3.11)
- `/usr/local/lib/pythonX.Y` or site-packages are copied into the distroless image

The failure mode is non-obvious and appears as Python corruption, but is actually an ABI/version mismatch.

---

### Root Cause

Distroless Python images ship a Python interpreter provided by the Debian base distribution.
They do not support arbitrary Python minor versions or mixing site-packages built against a different interpreter.

This limitation is not clearly documented and can easily surprise users building modern Python applications.

---

### Proposal

1. Improve documentation to clearly state:
   - Python minor versions must match exactly between builder and distroless runtime
   - Copying `/usr/local/lib/pythonX.Y` across versions is unsafe
   - Distroless Python is not recommended for ML or native-extension-heavy workloads

2. Consider publishing versioned Python tags (e.g. python3.11-debian12, python3.12-debian12)

3. Provide an official example showing safe usage patterns and common pitfalls

---

### Why this matters

Many modern Python workloads (FastAPI, ML inference, embeddings, pandas, torch, chromadb) rely on native extensions.
Without clear guidance, users often encounter runtime crashes that are difficult to debug.

Better documentation would significantly improve developer experience.

---

### Environment

- distroless image: python3-debian12
- builder image: python:3.12-slim
- workload: FastAPI + uvicorn + ML dependencies

kaushalacts:
### Summary

This PR improves documentation for `gcr.io/distroless/python3` by clarifying
Python version compatibility requirements and common pitfalls when using
multi-stage builds.

Several users (myself included) encounter confusing runtime failures when
copying Python site-packages built with a different Python minor version into
a distroless runtime image. These failures can appear as Python corruption
but are actually caused by ABI mismatches.

This change adds clear guidance to help users avoid this class of issues.

---

### What this PR does

- Documents that the Python minor version in the builder and runtime **must match**
- Warns against copying `/usr/local/lib/pythonX.Y` across mismatched versions
- Explains why ABI-related runtime errors can occur
- Clarifies the execution model (no shell, absolute paths required)
- Keeps scope limited to documentation only (no behavior changes)

---

### What this PR does NOT do

- Does not change image contents or build logic
- Does not propose runtime behavior changes
- Does not introduce new tags or versions

---

### Why this is useful

Modern Python workloads (FastAPI, data tooling, ML inference, etc.) often rely on
native extensions and newer Python versions. Without explicit documentation,
users can spend significant time debugging failures that are working as designed.

Clear documentation here should reduce support burden and improve user experience.

---

Fixes #1960


python3/README.md
```diff
@@ -6,11 +6,58 @@ This image contains a minimal Linux, Python-based runtime.
...
+
+For applications with native extensions (e.g. numpy, pandas, torch, ML workloads),
+consider using `python:<version>-slim` images instead.
+
+---
+
+### Execution Model
+
+Distroless images do not include a shell or PATH resolution. Commands must use
+absolute paths when invoking the Python interpreter:
```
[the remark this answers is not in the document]
kaushalacts: Thank you for the detailed review and the kind feedback — much appreciated.

Good catch on the wording in the *Execution Model* section. You’re absolutely right
that the python3 image’s default ENTRYPOINT allows providing a script via CMD,
as shown in the examples/python3 directory.

I agree that the current phrasing is too strong and could be confusing.
I’ll update the section to:
- Acknowledge the default ENTRYPOINT behavior
- Explain that absolute interpreter paths are a recommended, more explicit pattern,
  rather than a strict requirement

I’ll push a small revision shortly to reflect this. Thanks again for the guidance.

loosebazooka:
So a bit has changed since this PR was created. But there might be some relevant information still in here. If you want to revisit this PR, please do. Otherwise I'll close and apply the necessary changes to the README.

kaushalacts:
Hi @loosebazooka ,Could you please merge it now? I have reviewed and resolved the conflicts.

python3/README.md
```diff
@@ -17,6 +17,62 @@ On debian13+ we include a pre-generated `ld.so.cache` to support user/framework
 
 ## Usage
 
-The entrypoint of this image is set to "python", so this image expects users to supply a path to a .py file in the CMD.
+This image is intended to run Python applications directly.
```
loosebazooka, line 20: I think we should still call out what the entrypoint is.

python3/README.md
```diff
@@ -17,6 +17,62 @@ On debian13+ we include a pre-generated `ld.so.cache` to support user/framework
...
 
 See the Python [Hello World](../examples/python3/) directory for an example.
+
+---
+
+## ⚠️ Important Usage Notes
+
+### Python Version Compatibility
+
+The Python version included in distroless images is provided by the underlying
```
loosebazooka, line 30: I think this section can be compressed a little, but also add in an actual recommendation:

Use matching Debian 13 build image (e.g., python:3.13-slim-trixie) to ensure library compatibility.

python3/README.md
```diff
@@ -17,6 +17,62 @@ On debian13+ we include a pre-generated `ld.so.cache` to support user/framework
...
+
+### Recommended Usage
+
+Distroless Python images are best suited for:
+
+- Pure-Python applications
+- Minimal dependencies
+- Environments where Python version alignment is guaranteed
+
+For applications with native extensions (e.g. numpy, pandas, torch, ML workloads),
```
loosebazooka, line 58: I think the ctypes support for the debian13 images is in much better shape. I'm not a big python user but I think numpy would work. This section seems less relevant now.

python3/README.md
```diff
@@ -17,6 +17,62 @@ On debian13+ we include a pre-generated `ld.so.cache` to support user/framework
...
+The `python3` image provides a default ENTRYPOINT that invokes the Python
+interpreter, which allows users to supply a script directly via `CMD`
+(as demonstrated in the `examples/python3` directory).
+
+However, for maximum clarity and to avoid ambiguity—especially in more complex
+container configurations—it is often preferable to explicitly invoke the
+interpreter using an absolute path:
+
+```dockerfile
+CMD ["/usr/bin/python3", "-m", "your_module"]
```
loosebazooka, line 78: I'm not convinced this is the right usage of CMD. It feels like it would still use the ENTRYPOINT defined in the image spec.

python3/README.md
```diff
@@ -17,6 +17,62 @@ On debian13+ we include a pre-generated `ld.so.cache` to support user/framework
...
+
+### Execution Model
+
+Distroless images do not include a shell.
+
+The `python3` image provides a default ENTRYPOINT that invokes the Python
+interpreter, which allows users to supply a script directly via `CMD`
+(as demonstrated in the `examples/python3` directory).
+
+However, for maximum clarity and to avoid ambiguity—especially in more complex
```
loosebazooka, line 73: I would refrain from making these recommendations. Distroless is it what it is, and users can decide on how they want to use it, but we certainly don't make this recommendation.

lathama:
@kaushalacts is this still of interest to you or does this need to be closed.

PR #1970, opened 2026-01-14
First python3-debian13 images for testing

loosebazooka:
- Remove distutils dep and add venv. I'm not sure this is sufficient for all python users, but hopefully it will be.
- Not too many changes otherwise
- Not using [pbs](https://github.com/astral-sh/python-build-standalone) this round, will investigate for debian14


loosebazooka:
> This pull request has modified the root BUILD

hrmm modified incorrectly it looks like

PR #1974, opened 2026-01-19
refactor: support latest rules_oci

thesayyn:
Upgrade `rules_distroless` to 0.6.1 and rules_oci to 2.2.7 and standardize compression tooling with pigz.

Differences from production

All differences are cosmetic metadata or compression artifacts. The container filesystem is byte-for-byte identical (diff_ids match).

<table>
<tr><th>Gzip field</th><th>Production (Go <code>compress/gzip</code>)</th><th>Local (<code>zstd --format=gzip</code>)</th></tr>
<tr><td>Flags (byte 3)</td><td><code>0x00</code> (none)</td><td><code>0x08</code> (FNAME set)</td></tr>
<tr><td>Mtime (bytes 4-7)</td><td><code>0x00000000</code></td><td>File timestamp</td></tr>
<tr><td>XFL (byte 8)</td><td><code>0x04</code> (fastest)</td><td><code>0x00</code> (default)</td></tr>
<tr><td>Deflate stream</td><td>Go's encoding</td><td>zstd's encoding (~15% smaller)</td></tr>
</table>

These differences occur because:
- FNAME flag: zstd embeds the original filename; Go omits it
- Mtime: Go defaults to zero; zstd uses input file's mtime
- XFL: Go marks compression level; zstd doesn't set this hint
- Deflate encoding: Different LZ77 match-finding and Huffman tree heuristics produce
different (but equally valid) compressed output

Proof of equivalence

Layer 1:  Production sha256:fab8c4b3...  Local sha256:526604...  → diff_id: a33ba213... ✓
Layer 2:  Production sha256:bfb59b82...  Local sha256:990a9c...  → diff_id: 8fa10c01... ✓
Layer 3:  Production sha256:017886f7...  Local sha256:ef49c2...  → diff_id: 4840c7c5... ✓
Layer 4:  Production sha256:62de241d...  Local sha256:bf7a41...  → diff_id: 114dde0f... ✓
Layer 10: Production sha256:4aa0ea14...  Local sha256:ce4534...  → diff_id: 1a73b54f... ✓

### DIFF
--- a/common/BUILD.bazel
+++ b/common/BUILD.bazel
@@ -18,7 +18,7 @@ tar(
         "gnutar",
     ],
     compress = "gzip",
-    mtree = ["./ type=dir uid=0 gid=0 time=0.0"],
+    mtree = ["./ type=dir uid=0 gid=0 mode=0755 time=0.0"],
 )
 
 tar(


static/static.bzl
```diff
@@ -3,31 +3,46 @@
...
+def _nsswitch():
+    """Create compressed nsswitch tarball matching crane's compression.
+
+    Uses a genrule to directly compress the original nsswitch.tar with pigz,
+    preserving the exact tar structure without adding directory entries.
+    """
+    if native.existing_rule("nsswitch"):
+        return
+
+    native.genrule(
```
loosebazooka: Can we change this to using the contents of nsswitch.tar directly? I don't like having to depend on pigz like this:
- extract `/etc/nsswitch.conf` from `nsswitch.tar`
- remove `nsswitch.tar`
- create a pkg_tar rule to generate nsswtich.tar?

private/util/tar.bzl
```diff
@@ -0,0 +1,29 @@
+"tar macro with compression matching crane/go-containerregistry"
+
+load("@rules_pkg//:pkg.bzl", _pkg_tar = "pkg_tar")
+
+def pkg_tar(name, extension = "tar.gz", **kwargs):
```
loosebazooka: I think I would prefer we didn't overload this name `pkg_tar` to prevent any confusion.

pkg_tar should still be available where it is needed, but maybe this should be called pkg_tar_gz or something that is the macro here?

loosebazooka (commented):
Just a clarificaiton on this:
Mtime (bytes 4-7)	
0x00000000 vs File timestamp

I imagine we prefer the 0 over the file timestamp?

static/static.bzl
thesayyn: that's reasonable.

private/util/tar.bzl
thesayyn: probably yes, do you still want to keep rules_pkg here? tar.bzl is faster at same job.

loosebazooka: I don't see any need to keep it around if it's not required

static/static.bzl
loosebazooka, line 52: for some reason `rootfs` is being created with mode 750, I think we need to update the rule in common to actually explicitly set the mode to 0755

thesayyn: pushed a commit that should do it.

loosebazooka (approved):

PR #1999, opened 2026-02-12
update rules_distroless to 0.6.2

mmoll:


loosebazooka:
dunno what rules_distroless is doing here, but none of these tests are gonna pass because `gawk-5.3.2.tar.xz` is not obtainable form the mirror right now

thesayyn:
This is done already with: https://github.com/GoogleContainerTools/distroless/pull/1974

mmoll:
Will defer to #1974 then and leave this only open for reference until the other one gets in.

PR #2002, opened 2026-02-15
docs: clarify openssl-legacy-provider is a package

paastha1:
Minor docs clarification - openssl-legacy-provider is a package name, not just a concept.


paastha1:
Closing - will resubmit with a cleaner change.

PR #2004, opened 2026-02-15
docs: update ldd link from HTTP to HTTPS

paastha1:
Update the ldd man page link in README.md from `http://` to `https://`.

The man7.org site supports HTTPS, so this ensures users aren't redirected through an insecure connection.


paastha1:
Closing - no longer needed

Issue #1996, opened 2026-02-05
tzdata-legacy should be included in distroless/static image


jegorbunov:
**Describe the bug**  
My Go application parses timezones provided by external sources.  
It uses `gcr.io/distroless/static:nonroot` as the base image.

After updating to newer image versions (e.g. digest `f9f84bd96843`), the application started failing to resolve some timezones that previously worked. Investigation suggests this is related to the Debian 13 change that splits timezone data into `tzdata` and `tzdata-legacy`:  
https://www.debian.org/releases/trixie/release-notes/issues.en.html#timezones-split-off-into-tzdata-legacy-package

It appears that legacy zones are no longer present in the image filesystem.

Is this expected behavior?  
I believe this is not desirable for compatibility, and that `tzdata-legacy` should be included in new `static` images.

---

**To Reproduce** 

1. Build/run a Go application that loads timezones via `time.LoadLocation()`  
2. Use `gcr.io/distroless/static:nonroot` as the runtime base image  
3. Attempt to load a legacy timezone (example: "Europe/Kiev")  
4. Observe failure resolving the timezone in newer image digests (e.g. `f9f84bd96843`)

---

**Expected behavior**

Previously valid timezone identifiers should continue to resolve successfully.  
The container image should include all timezone definitions necessary for backward compatibility, including those now provided by `tzdata-legacy`.

---

**Console Output**  
(not applicable)

loosebazooka:
Yeah that's not unreasonable. Feel free to add a pr or I'll get it to today/tomorrow.

loosebazooka:
Also, we generally don't recommend using the non-debian version suffixed image references: `gcr.io/distroless/static:nonroot` for this exact reason. I would instead, for some stability, use `gcr.io/distroless/static-debian12:nonroot` or `gcr.io/distroless/static-debian13:nonroot` and update in a more controlled manner.

PR #2007, opened 2026-02-15
docs: update README to reflect Debian 13 as new default

paastha1:
## Summary

PR #1987 changed `DEFAULT_DISTRO` to `debian13`, but the README was not updated. The "Base Operating System" section still says:

> Distroless images are based on Debian 12 (bookworm)... will currently select `-debian12` images

This is causing user confusion (see #1996) since the untagged images now resolve to Debian 13.

This PR updates the README to reflect the current default.

## Changes

- Updated Debian version from 12 (bookworm) to 13 (trixie)
- Updated example suffix from `-debian12` to `-debian13`
- Removed "but that will change in the future" since the change already happened

### DIFF
--- a/README.md
+++ b/README.md
@@ -184,7 +184,7 @@ See here for more information on how these images are [built and released](RELEA
 
 ### Base Operating System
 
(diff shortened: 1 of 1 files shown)


paastha1:
Closing - no longer needed

loosebazooka:
Please avoid using this repository as a test for your experiments.

PR #2022, opened 2026-02-27
Add JAVA support for RISC-V

mmoll:


mmoll:
here are the build failures again 😅

loosebazooka:
oh this just wont work with the current code setup, the debs contain zst archives for riscv64. I don't think rules_distroless supports this or we need to muck around with the config

loosebazooka:
oh wait no never mind, it's in our custom spdx generator. zst is supported by rules_distroless

loosebazooka:
try adding the extensions `.tar.zst` to [/private/pkg/debian_spdx.bzl](https://github.com/GoogleContainerTools/distroless/blob/main/private/pkg/debian_spdx.bzl#L49-L50)

mmoll:
woohoo 💚

PR #2055, opened 2026-04-07
feat(python3): add riscv64 support for debian13

brianredbeard:
This change enables riscv64 architecture support for the Python 3 distroless images based on Debian 13 (Trixie).

Updates include:
- Adding riscv64 to the trixie_python debian repository configuration.
- Updating Bazel lockfiles and repository lock files with riscv64 package metadata.
- Adding the required ld.so.cache for riscv64.
- Updating the README to reflect supported architectures.


python3/config.bzl
```diff
@@ -1,7 +1,7 @@
 PYTHON_DISTROS = ["debian12", "debian13"]
 PYTHON_ARCHITECTURES = {
     "debian12": ["amd64", "arm64"],
-    "debian13": ["amd64", "arm64"],
+    "debian13": ["amd64", "arm64", "riscv64"],
```
loosebazooka, line 4: I think this needs to become a map. I don't think we should make new releases for Debian 12

brianredbeard: Thanks for the review! I'm a bit confused by the `map()` suggestion - could you elaborate on what you have in mind? The current structure is a dict keyed by distro and we are only mutating "debian13" (not "debian12"), which is the same pattern used in `base/config.bzl`, `cc/config.bzl`, `static/config.bzl`, and `java/config.bzl`. I want to make sure I'm not missing something - are you thinking of a different approach for how architectures should be defined here?

loosebazooka: I'm dumb, I was reading this on mobile and must've misread

loosebazooka:
Can I get some additional info about the usecase here? Also is the current state of the `ld.so.cache` sufficient for what you need to do?

loosebazooka:
Actually looks like you need to run the rule to generate the `ld.so.cache` for riscv:
You need [docker with multiplatform support](https://docs.docker.com/build/building/multi-platform/) to do this
```
bazel run //python3:update_ldconfig
```

brianredbeard:
Thanks for the reminder about the ldcache generation.  I hacked past it because of a transient issue and forgot to implement that step. 

As to the use case, in the [distro-integration](https://wiki.riseproject.dev/display/HOME/Distro+Integration+WG)* working group meetings of @riseproject-dev, we have been identifying the need for RISC-V contributions across the the open-source ecosystem to ensure feature parity with other architectures.

We have a special focus on Kubernetes (we're currently using Kubernetes to run the [recently announced](https://riseproject.dev/2026/03/24/announcing-the-rise-risc-v-runners-free-native-risc-v-ci-on-github/) native RISC-V GitHub actions runners), but also generally on cloud-native infrastructure.  Put simply, if one wants to have RISC-V support in the cloud, all of the RISC-V undercloud components need to exist.

I would have also submitted PRs for NodeJS, but I noticed that the distroless policy seems to be to load the binary releases directly from nodejs.org.  Unfortunately, RISC-V is only "experimental" status and so does not publish official RISC-V binaries (though the various distros do publish their own binary builds, limited to their individual packaging policies).

*: The irony of me opening a pull-request in `distroless` is not lost on me. 😂

loosebazooka (approved):
Thanks! I vaguely recall someone asking for this for their AI workloads. Curious to see how that plays out.

PR #2008, opened 2026-02-15
docs: fix incorrect debian version in CONTRIBUTING.md

paastha1:
## Summary

`CONTRIBUTING.md` references `debian17` in build examples (lines 16 and 27), which doesn't exist. The current default distro is `debian13` (since PR #1987). 

This fixes both occurrences to use `debian13` so new contributors can follow the examples correctly.

## Changes

- `bazel build //base:static_root_amd64_debian17` → `debian13`
- `oci_load` example image reference updated to `debian13`

PR #2086, opened 2026-05-13
docs: fix contributing build examples

immanuwell:
Tiny docs cleanup. The CONTRIBUTING examples pointed at a target that does not exist, so copy/paste goes boom.

Repro:
```sh
bazel query //base:static_root_amd64_debian17
# ERROR: no such target
```

This switches the examples to `//base:base_root_amd64_debian13`, which exists and builds. Also pins the buildifier install command to `3.2.0`, same as CI. no surprises there.

Checked:
```sh
bazel query //base:base_root_amd64_debian13
bazel build //base:base_root_amd64_debian13
./knife lint --check
git diff --check
```

Related: #2008

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -13,7 +13,7 @@ Follow either of the two links above to access the appropriate CLA and instructi
 
 ## How to Build and Test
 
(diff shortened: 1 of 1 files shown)


loosebazooka:
oh weird, how did debian17 happen?

PR #2101, opened 2026-06-02
Harden Node archive Starlark generation

yoshihiro999:
This hardens the Node archive updater by validating release metadata before generation and consistently quoting generated Starlark strings.

Changes:

- Validate Node.js versions from the release index with a conservative `x.y.z` allowlist.
- Validate archive suffixes, distroless architecture names, and SHA256 values before writing Starlark.
- Quote generated Starlark string values instead of interpolating raw metadata into string literals.
- Add a small regression test for invalid metadata values and Starlark string escaping.

Local checks run:

```text
node knife.d/update_node_archives_test.js
node --check knife.d/update_node_archives.js
node --check knife.d/update_node_archives_test.js
bash -n knife
git diff --check
```

### DIFF
--- a/knife
+++ b/knife
@@ -176,6 +176,10 @@ function cmd_update_node_archives () {
             | sort_by(.date) | reverse | .[0].version
         ')
         latest_nov=$(echo "$latest_version" | sed 's/^v//')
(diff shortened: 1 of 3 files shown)


yoshihiro999:
Superseded by #2102 after the CLA check on this PR did not update cleanly.

PR #2110, opened 2026-06-19
Fix crane version so symlink oci images don't fail

loosebazooka:
see: https://github.com/google/go-containerregistry/pull/2306


bobcallaway (approved):
If this is going to the right long term approach I’d switch this to an explicit go.mod with dependabot  support + go tool install

PR #2111, opened 2026-06-23
Add OCI source annotation and custom version annotation to images

### LINKED ISSUE #686: Version pinning support?
I am a bit confused about the version tagging scheme for distroless images. My current understanding is that the tags (e.g. base-debian10) are updated each time a new version is published. This seems to indicate that version pinning is not possible. Are there other tags (e.g. git commit-id) that can be used to pin? If not, is the idea that new versions of distroless tags are automatically rolled out to your images the next time they are built (this seems dangerous!)?

Maybe I've missed something, but it would be very helpful to be able to specify which version of a distroless tag (e.g. base-debian10-20210220 instead of base-debian10) is used to build.

joebowbeer:
Adds OCI image annotations to all distroless image types, addressing the version pinning request in #686.

Annotations added:

* `org.opencontainers.image.source` - set to the repository URL (from OS_RELEASE["HOME_URL"] in variables.bzl) on all image types: base, cc, static, python3, nodejs, and java
* `com.google.distroless.nodejs.version` - the specific Node.js archive version (e.g. 24.18.0), per architecture
* `com.google.distroless.java.version` - the specific Temurin JRE/JDK version installed

The `org.opencontainers.image.source` annotation is the most essential annotation to add, per

https://snyk.io/blog/how-and-when-to-use-docker-labels-oci-container-annotations/

The Java and NodeJS images are the only images where it makes sense to add a separate version annotation, and for these I think a custom namespace (`com.google.distroless`) is needed.

More annotations can be added later as needed.

Implementation notes for NodeJS versioning:

Rather than adding a static NODEJS_VERSIONS dict to config.bzl, a new node_versions_repo repository rule is generated by the existing node module extension in node.bzl. This creates an @node_versions//:versions.bzl file that nodejs.bzl loads at analysis time - the same pattern used for Debian package versions. The update_node_archives.js script is updated to regenerate the node_versions_repo() call alongside the existing node_archive() calls, so there is no version data to keep in sync manually.

Closes #686

### DIFF
--- a/nodejs/nodejs.bzl
+++ b/nodejs/nodejs.bzl
@@ -1,8 +1,9 @@
 "nodejs image definitions"
 
 load("@container_structure_test//:defs.bzl", "container_structure_test")
+load("@node_versions//:versions.bzl", "NODEJS_VERSIONS")
 load("@rules_oci//oci:defs.bzl", "oci_image", "oci_image_index")
-load("//common:variables.bzl", "DEBUG_MODE", "USERS")
+load("//common:variables.bzl", "DEBUG_MODE", "OS_RELEASE", "USERS")
 load("//private/util:deb.bzl", "deb")
 load("//private/util:tar.bzl", "tar")
 
@@ -45,6 +46,14 @@ def nodejs_image(distro, major_version, arch, packages):
         packages: any deb packages to add to the image
     """
 
+    _version_key = major_version + "_" + arch
+    if _version_key not in NODEJS_VERSIONS:
+        fail("No version found for Node.js major version/arch: " + _version_key)
+    _annotations = {
+        "org.opencontainers.image.source": OS_RELEASE["HOME_URL"],
+        "com.google.distroless.nodejs.version": NODEJS_VERSIONS[_version_key],
+    }
+
     for mode in DEBUG_MODE:
         for user in USERS:
             oci_image(
@@ -57,6 +66,7 @@ def nodejs_image(distro, major_version, arch, packages):
                 ] + [
                     "@nodejs" + major_version + "_" + arch,
                 ],
+                annotations = _annotations,
             )
 
     _check_certificates_tar()
--- a/private/extensions/node.bzl
+++ b/private/extensions/node.bzl
@@ -88,6 +88,24 @@ node_archive = repository_rule(
     },
 )
 
+_NODE_VERSIONS_TMPL = """\
+"node versions"
+
+# AUTO GENERATED. DO NOT EDIT.
+NODEJS_VERSIONS = {versions}
+"""
+
+def _node_versions_repo_impl(rctx):
+    rctx.file("versions.bzl", _NODE_VERSIONS_TMPL.format(versions = str(rctx.attr.versions)))
+    rctx.file("BUILD.bazel", 'exports_files(["versions.bzl"])')
+
+node_versions_repo = repository_rule(
+    implementation = _node_versions_repo_impl,
+    attrs = {
+        "versions": attr.string_dict(),
+    },
+)
+
 def _node_impl(module_ctx):
     mod = module_ctx.modules[0]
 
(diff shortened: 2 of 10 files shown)


nodejs/nodejs.bzl
```diff
@@ -45,6 +48,14 @@ def nodejs_image(distro, major_version, arch, packages):
         packages: any deb packages to add to the image
     """
 
+    version = NODEJS_VERSIONS.get(major_version, "")
+    _annotations = {
+        "org.opencontainers.image.source": _DISTROLESS_SOURCE_URL,
+        "org.opencontainers.image.url": _DISTROLESS_SOURCE_URL,
+    }
+    if version:
+        _annotations["org.opencontainers.image.version"] = version
```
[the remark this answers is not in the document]
joebowbeer: @loosebazooka applied

private/extensions/node.bzl
```diff
@@ -229,8 +247,18 @@ def _node_impl(module_ctx):
         control = "//nodejs:control",
     )
 
+    node_versions_repo(
```
loosebazooka, line 250: It's not unusual for different node versions/architectures to be updated separately, look-ups should allow finding a version per architecture.

nodejs/nodejs.bzl
```diff
@@ -1,11 +1,14 @@
 "nodejs image definitions"
 
 load("@container_structure_test//:defs.bzl", "container_structure_test")
+load("@node_versions//:versions.bzl", "NODEJS_VERSIONS")
 load("@rules_oci//oci:defs.bzl", "oci_image", "oci_image_index")
 load("//common:variables.bzl", "DEBUG_MODE", "USERS")
 load("//private/util:deb.bzl", "deb")
 load("//private/util:tar.bzl", "tar")
 
+_DISTROLESS_SOURCE_URL = "https://github.com/GoogleContainerTools/distroless"
```
loosebazooka: this can come from `common/variables.bzl`

nodejs/nodejs.bzl
```diff
@@ -45,6 +48,14 @@ def nodejs_image(distro, major_version, arch, packages):
         packages: any deb packages to add to the image
     """
 
+    if major_version not in NODEJS_VERSIONS:
+        fail("No version found for Node.js major version: " + major_version)
+    _annotations = {
+        "org.opencontainers.image.source": _DISTROLESS_SOURCE_URL,
+        "org.opencontainers.image.url": _DISTROLESS_SOURCE_URL,
+        "org.opencontainers.image.version": NODEJS_VERSIONS[major_version],
```
loosebazooka: I think as per @imjasonh's [comment](https://github.com/GoogleContainerTools/distroless/issues/686#issuecomment-4329500152), this maybe could be `com.google.distroless.nodejs.version` or something.

Though I'm not exactly sure what the best option is here.

loosebazooka (commented):
sorry was ooo last week.

joebowbeer:
> @loosebazooka Should I extend this PR to annotate other kinds of images?
> 
> Additional annotations:
> 
> All images: add source + url annotations to base.bzl, cc.bzl, static.bzl, python.bzl, java.bzl Java: add version using jre_ver(deb.version(...) already computed for the JAVA_VERSION env var Python: add version using deb.version(arch, distro, "python3", ...) available via the same deb.version() mechanism
> 
> * base, cc, static don't have an application-level version to advertise, but org.opencontainers.image.source and org.opencontainers.image.url would still be useful for all of them. Renovate uses source to link PRs back to the repo regardless of versioning.
> * Java in particular would be high value since the original issue mentions it specifically and the version string is already sitting in the JAVA_VERSION env var computation.

Done: Added source annotation to all images.

joebowbeer:
@loosebazooka PTAL

loosebazooka:
Thanks, this looks pretty good to me, I will run the diffs and see what changed in the outputs to make sure

loosebazooka:
in the meantime I think you need to run `bazel mod deps --lockfile_mode=update` to update the bazel lockfile

joebowbeer:
> in the meantime I think you need to run `bazel mod deps --lockfile_mode=update` to update the bazel lockfile

Lockfile updated

loosebazooka (approved):

PR #2147, opened 2026-08-04
fix(knife): GPG-verify Node.js SHASUMS256.txt.asc before committing checksums

rishuranjanofficial:
## Problem

`knife.d/update_node_archives.js` computes SHA-256 checksums by downloading
each Node.js tarball directly and hashing it locally. The resulting hashes are
committed to `private/extensions/node.bzl` and later used by Bazel to verify
downloads.

If the nodejs.org CDN, a reverse proxy, or DNS resolution is compromised at the
time the nightly `update-node-archives` workflow runs, an attacker can serve a
malicious tarball. The script will compute and commit the correct SHA-256 of the
*malicious* file — Bazel will then accept it on every subsequent build, silently
backdooring all `gcr.io/distroless/nodejs*` images.

## Solution

Node.js publishes a GPG-signed `SHASUMS256.txt.asc` alongside every release,
signed by the Node.js release team's keys (documented at
https://github.com/nodejs/node#release-keys). Verifying this signature before
trusting any checksum establishes a chain of custody back to the Node.js release
team and eliminates the attack vector above.

This PR:
1. Imports the 10 current Node.js release GPG keys into an isolated temporary
   keyring at startup (with fallback to a secondary keyserver).
2. Downloads `SHASUMS256.txt` and `SHASUMS256.txt.asc` for each Node.js version.
3. Runs `gpg --verify` — the script **hard-fails** if the signature does not
   validate, preventing any bad hashes from being committed.
4. Parses the verified `SHASUMS256.txt` to extract per-architecture SHA-256
   values, replacing the previous approach of hashing downloaded tarballs.
5. Cleans up all temporary files in a `finally` block.

The nightly workflow already runs on `ubuntu-latest`, which has `gpg`
pre-installed. No runner or workflow changes are required.

## Security impact

| Scenario | Before this fix | After this fix |
|---|---|---|
| CDN/BGP/DNS MITM during nightly cron | Malicious SHA-256 committed | GPG verification fails; workflow aborts |
| nodejs.org serving wrong tarball | Bad hash committed | GPG verification fails; workflow aborts |
| SHASUMS256.txt tampered without valid key | Not checked | GPG verification fails; workflow aborts |

## Testing

Verified locally that:
- GPG signature validation passes for Node.js 22 current release SHASUMS.
- The SHA-256 values parsed from the verified SHASUMS file match those produced
  by the previous tarball-hash approach.
- A tampered SHASUMS file causes the script to exit non-zero, preventing PR
  creation.


rishuranjanofficial:
@loosebazooka 

I've submitted this pull request and would appreciate a review when you have some bandwidth. Please let me know if any changes or additional context are needed.

Thank you for your time and consideration.

rishuranjanofficial:
Hi @loosebazooka @nlopezgi @bobcallaway 

Gentle ping on #2147. The PR adds GPG verification of SHASUMS256.txt.asc before committing Node.js checksums, closing a supply chain attack vector in the nightly update-node-archives workflow. Happy to address any feedback. Thanks!

lathama:
I am not a node person but did a quick check. I feel the `execSync` to connect to `keyserver.ubuntu.com` will fail on some build environments that limit outbound network connections. I think you are trying to fix a real issue but I doubt the shell execution is the best option.

rishuranjanofficial:
Hi @lathama 

Thanks for the review!

You're right - fetching from a keyserver at runtime breaks in restricted network environments, and using execSync for shell commands is fragile.

Better approach: bundle the Node.js release team's public keys directly in the repo as a committed file (sourced from https://github.com/nodejs/node#release-keys) and replace the gpg shell call with the openpgp npm package for pure Node.js verification - no system dependencies, no network calls at build time.

These are public keys, not secrets - they're already published openly by the Node.js team and are safe to commit. Think of them like a list of trusted signatures you keep on file to check documents against.

The only maintenance consideration is that Node.js occasionally adds new release signers. When that happens the keys file needs a one-line update via PR, which is better than fetching keys silently at runtime - changes go through review and are tracked in git history.

Does this direction sound good to you, or do you have a different approach in mind?

lathama:
@rishuranjanofficial I don't have any better ideas, just calling out what I saw. Not sure if there is any solution to this at this time.

loosebazooka:
Please don't ping random users like you've done [here](https://github.com/GoogleContainerTools/distroless/pull/2147#issuecomment-5237315173). The team is aware of this PR and have added it to our backlog. There is no timeline when someone will be available to review this.

rishuranjanofficial:
Apologies - I didn't mean to bother anyone, and I'll stop the pings. Thanks for the update; glad it's on the backlog. Happy to address any feedback whenever the team has bandwidth, no rush on my end.

loosebazooka:
alright thanks for your patience with this:
1. Lets fetch and commit the nodejs release keys. Only keep active keys (maybe https://github.com/nodejs/release-keys/tree/main/gpg-only-active-keys/pubring.kbx converted to a single nodejs_keys.asc using gpg tooling). Only checking in that ascii formatted key list file.
2. add a script to keep the keys in sync, update-node-keys or something that runs on a daily cron.

rishuranjanofficial:
Done - implemented as you outlined:

Committed the active Node.js release keys as a single knife.d/nodejs_keys.asc (exported from nodejs/release-keys/gpg-only-active-keys via gpg).
Verification now imports that in-repo file offline (no keyserver), replacing the gpg --recv-keys calls.
Added knife.d/update_node_keys.sh plus a daily update-node-keys cron that opens a PR when the keys change (mirrors the existing update-node-archives bot-PR pattern).

One expected, by-design behavior I'll call out so it's not a surprise (not a concern): verification is fail-closed against the committed keys. If Node.js introduces a brand-new release signer, a release signed only by that new key won't verify until the daily cron refreshes nodejs_keys.asc (or the script is run manually) - a self-healing ~1-day window. It only touches the internal update-node-archives maintenance job, not contributors, PRs, or builds, and it's strictly more secure than the previous static key list. PTAL.

loosebazooka:
> One expected, by-design behavior I'll call out so it's not a surprise (not a concern): verification is fail-closed against the committed keys. If Node.js introduces a brand-new release signer, a release signed only by that new key won't verify until the daily cron refreshes nodejs_keys.asc (or the script is run manually) - a self-healing ~1-day window. It only touches the internal update-node-archives maintenance job, not contributors, PRs, or builds, and it's strictly more secure than the previous static key list. PTAL. 

I think this is fine and should allow for manual key update by maintainers. I expect this occur somewhat infrequently.

rishuranjanofficial:
Thanks! That's everything implemented on my end:

active Node.js release keys committed as knife.d/nodejs_keys.asc
offline verification against that file (no keyserver)
update_node_keys.sh + daily update-node-keys cron, with on-demand manual refresh (run the script or trigger the workflow)

I believe this is ready for review - happy to make any further changes if needed.

knife.d/update_node_archives.js
```diff
@@ -13,51 +16,118 @@ const architectures = ["amd64", "arm64", "arm", "ppc64le", "s390x"];
...
+ * version, verify the signature against the imported release keys, and return
+ * a map of { filename → sha256 }.
+ *
+ * Throws if GPG verification fails — no checksums are returned in that case.
+ */
+const fetchVerifiedShasums = async (nodeVersion, gpgHome) => {
+  const base = `https://nodejs.org/dist/v${nodeVersion}`;
+  const [shasumsText, shasumsAsc] = await Promise.all([
+    fetchText(`${base}/SHASUMS256.txt`),
+    fetchText(`${base}/SHASUMS256.txt.asc`),
```
loosebazooka: looks like the .txt.asc file is inline signature, so either we use 
- `gpg --verify SHASUM256.txt.sig SHASUM256.txt` to validate the `.txt` using the `.txt.sig` file
or
- `gpg --decrypt SHASUM256.txt.asc` to unwrap the `.txt.asc` and use it

.github/workflows/update-node-keys.yml
```diff
@@ -0,0 +1,53 @@
...
+    permissions:
+      contents: write
+      pull-requests: write
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
+        with:
+          persist-credentials: false
+
+      - name: Update node release keys
+        run: ./knife.d/update_node_keys.sh
```
loosebazooka: I know our knife script structure is a bit silly. But update_node_keys.sh should be exposed directly in the knife script to be called like `./knife update-node-keys`

loosebazooka (commented):
a few minor things, but this seems almost done

.github/workflows/update-node-keys.yml
rishuranjanofficial: Nice catch - switched to gpg --output - --decrypt SHASUMS256.txt.asc, which verifies the inline signature and returns the authenticated checksums, and dropped the separate plaintext SHASUMS256.txt fetch so nothing unsigned is ever parsed. Validated against a real release (v22.11.0): good signature from a Node.js release signer + correct checksums, and confirmed it fails closed (non-zero exit) when the signer isn't in the committed keys. Pushed.

loosebazooka:
looks like you got the review notes mixed up, there's one remaining item and I Think we can merge

.github/workflows/update-node-keys.yml
rishuranjanofficial: added an update-node-keys subcommand to knife (cmd_update_node_keys) and switched the workflow to run ./knife update-node-keys, matching the update-node-archives pattern. Pushed.

knife.d/update_node_archives.js
```diff
@@ -13,51 +16,122 @@ const architectures = ["amd64", "arm64", "arm", "ppc64le", "s390x"];
...
+/**
+ * Import the committed Node.js release public keys into an isolated temporary
+ * keyring. No network access - keys come from the in-repo file.
+ * Caller is responsible for deleting the returned GNUPGHOME.
+ */
+const importReleaseKeys = () => {
+  const tmpHome = fs.mkdtempSync(path.join(os.tmpdir(), "gpg-distroless-"));
+  fs.chmodSync(tmpHome, 0o700);
+  const env = { ...process.env, GNUPGHOME: tmpHome };
+  execSync(`gpg --batch --import ${NODE_KEYS_FILE}`, { stdio: "pipe", env });
```
loosebazooka: AI recommends execFileSync with arg array to prevent potential shell expansion issues.

rishuranjanofficial: Sure, switched both gpg calls (the key import and the SHASUMS --decrypt) from execSync to execFileSync with argument arrays, so gpg is invoked directly with no shell and no expansion of the file paths. Pushed. Thanks!

loosebazooka (approved):

loosebazooka:
sorry for the slow turn around, I'll merge this once the last check finishes. Thanks!

PR #2154, opened 2026-08-10
feat: Python standalone images from python-build-standalone

reneleonhardt:
## Summary

New `python/` package (mirrors `nodejs/`) building distroless Python images from [python-build-standalone](https://github.com/astral-sh/python-build-standalone) (PBS) prebuilt tarballs instead of Debian-packaged python. The image base stays Debian (`//cc` provides glibc); only the python distribution method changes.

Matrix: 3.13.15 / 3.14.7 x debian13 x amd64 / arm64 / s390x / riscv64 x root/nonroot x normal/debug.

Tags: `python3.13-debian13`, `python3.14-debian13` (+ per-arch and default-distro variants, nodejs publish pattern).

## Why PBS

- One tarball serves all archs; runtime deps are glibc only (full stdlib import walk passes with zero deb packages) — PBS statically links zlib/bz2/sqlite3/OpenSSL/ffi/lzma/ncurses into libpython, so there are no shared objects to ship and no PYTHON_PACKAGES list to maintain (only libc-bin for ldconfig).
- Decouples Python versions from the Debian release (3.13 and 3.14 on debian13; 3.15 added by the updater when it reaches a stable release).
- Pinned to PBS release 20260807; shas from the release SHA256SUMS, updated via `knife update-python-archives`.

## Platform exclusions (verified, not oversights)

- **armv7**: PBS publishes soft-float gnueabi builds (interpreter `/lib/ld-linux.so.3`); the distroless armhf base ships `/usr/lib/ld-linux-armhf.so.3` - the binary cannot exec (confirmed in Docker). Removed.
- **ppc64le**: not published by PBS.
- s390x / riscv64 loaders verified against the cc bases (`/usr/lib/ld64.so.1`, `/usr/lib/ld-linux-riscv64-lp64d.so.1`).

## Notes

- debian12 intentionally absent: new debian12 builds are being deprecated upstream.
- `find_library()` works for system libs (c/ssl/z verified): image ships `/sbin/ldconfig` (libc-bin) + a per-arch `ld.so.cache` (python3/ parity, maintainer request); regenerated via `bazel run //python:update_ldconfig`, checked by `//python:check_ldconfig_tests` (4/4) + a CST test. Statically-linked PBS libs have nothing to find; `python -m venv` works (ensurepip wheels bundled); `crypt` moot (removed in 3.13).
- testdata schema 2.0.0 (nodejs parity); `python3/` (deb) untouched.

## Verification

- `bazelisk build //python/...` green (debian13, 3.13+3.14, 4 archs).
- 8/8 structure tests via colima docker (arm64, debian13, 3.13+3.14, root+nonroot).
- amd64 image executed in Docker: `3.14.7 x86_64`, ssl certs OK.
- s390x/riscv64 built; OCI config (arch/os/annotations) verified; runtime in CI.
- All required cc base permutations verified (arch configs correct).
- armv7/ppc64le exclusions verified (loader/ABI evidence, see Platform exclusions).
- find_library('c'/'ssl'/'z') resolves in the image (ldconfig + ld.so.cache); `//python:check_ldconfig_tests` 4/4 (amd64/arm64/s390x/riscv64); CST includes a find_library test; CDLL(soname) works (libc/libssl/libgcc/libpython).
- Updater (`knife update-python-archives`): no-op idempotent; `--verbose`/`--dry-run`; fills new stable minors, drops EOL tail, sorts+dedupes the matrix; hermetic smoke test `//python:update_python_archives_test`; pre-flight validation (RED mutates nothing; recovery via git, lockfile reset documented).
- buildifier check + lint clean.
- CI wiring: images build via `kind(oci_image, deps(:sign_and_push))`; CST tagged `manual`+arch run via `attr(tags, "amd64", ...)` — same rails as python3/; `check-ldconfig.yaml` extended to cover `//python:check_ldconfig_tests`.

## Checklist

- [x] Google CLA signed (per CONTRIBUTING.md)
- [x] Signed-off-by on all commits (repo convention)
- [ ] CI green on fork: `bazelisk build //python/...`, structure tests


loosebazooka:
At a highlevel:
1. don't add any new debian12 builds (deprecating those in a month)
2. there's something in here I don't understand (but maybe it just means we don't need any new debs here)
    > runtime deps are glibc only (full stdlib import walk passes with zero deb packages) 
3. images should just be 3.13 and 3.14, the code should be general purpose enough to add 3.15 when it is released
4. an updater script should be included to find and update new releases of python

reneleonhardt:
> don't add any new debian12 builds (deprecating those in a month)

Removed. `python/` is debian13-only (config.bzl, README, publish map).

> "runtime deps are glibc only (full stdlib import walk passes with zero deb packages)" — unclear

Explained above. PBS linux builds statically bundle zlib/bz2/sqlite3/OpenSSL/libffi/lzma/ncurses; the tarball binaries reference only glibc core libs, so the image adds zero python-specific deb packages. The 502-module stdlib import walk (ssl, sqlite3, zlib, ctypes, https) passes in-image with none; deliberate skips (GUI, `_pyrepl._minimal_curses`, `dbm.gnu`, windows-only, `test`) are documented so "zero debs" is exact, and any future missing lib fails the walk.

> images should just be 3.13 and 3.14; code general enough to add 3.15 when it is released

Done. Matrix is 3.13 + 3.14; `_bin_version` + the matrix-driven config make 3.15 a one-entry addition when it reaches stable (no rc in the matrix anymore).

> an updater script should be included to find and update new releases of python

Included: `knife update-python-archives` (knife.d/update_python_archives.sh). Reads the matrix from `python/config.bzl`, queries the latest PBS release, replaces patch versions + SHAs in place, fills new stable minors (3.15 once 3.15.0 ships), honors EOL tail drops, keeps testdata + MODULE.bazel `use_repo` in sync. `--verbose`/`--dry-run` flags, hermetic smoke test (`//python:update_python_archives_test`), pre-flight validation (a failed run mutates nothing).

loosebazooka:
I'm sorry I will get to this soon, lots of other things on my plate.

reneleonhardt:
No stress, my awesome contribution everyone's been waiting for isn't running away 😄

loosebazooka:
I'm no python expert, but the ldconfig seemed somewhat important to people. is CDLL sufficient in this case? are people not going to use "find_library()" ?

loosebazooka:
I had a teammate recommend we keep the ldconfig binary and continue to generate the ldconfig.so.cache to allow find_library to continue to work

reneleonhardt:
Good catch! Now python3/ parity.

- Image ships `/sbin/ldconfig` (libc-bin) + a per-arch `ld.so.cache` (amd64/arm64/s390x/riscv64), generated in-image via `bazel run //python:update_ldconfig`.
- `//python:check_ldconfig_tests` 4/4 (also wired into `check-ldconfig.yaml`); CST `find_library` test added.

Verified in the image: `find_library('c') -> libc.so.6`, `find_library('ssl') -> libssl.so.3`, `find_library('z') -> libz.so.1`.

Scope note (deep-audited against the python3 image): PBS statically links zlib/bz2/sqlite3/OpenSSL/ffi/lzma/ncurses INTO libpython, so those modules work with no shared object to find - `find_library()` covers system libs (c/ssl/z), which is what the cache lists. `python -m venv` works (ensurepip wheels are bundled); `crypt` is moot (module removed in 3.13, PEP 594); dbm.gnu/tkinter are absent in both images (parity).

python/README.md
```diff
@@ -0,0 +1,68 @@
...
+and verified by `bazel test //python:check_ldconfig_tests` (also in CI).
+
+Libraries that python-build-standalone links statically into the interpreter
+(sqlite3, ncurses, bz2, lzma, expat, zlib, ...) have no shared object to find —
+the corresponding stdlib modules work out of the box. `python -m venv` works and
+bootstraps pip from the bundled `ensurepip` wheels.
+
+## Provenance
+
+Python comes from python-build-standalone release `20260807`
```
loosebazooka: I think this part is a bit odd, and will have to be updated on every update. I think just linking to astral's pbs should be good. But not a specific version.

knife.d/update_python_archives.sh
```diff
@@ -0,0 +1,335 @@
+set -o pipefail -o errexit -o nounset
```
loosebazooka, line 1: gemini seems to think there's an issue with this script where a release tag change will go unnoticed 20260814 -> 20260815 which may be due to underlying library native library changes while the python version remains unchanged.

loosebazooka (commented):
I have some minor concerns as to how scanners will pick up the PBS embedded native libraries. If one were to scan the python 3.14 image, what happens when the embedded zlib has a CVE? Is its version discoverable?

loosebazooka:
> I have some minor concerns as to how scanners will pick up the PBS embedded native libraries. If one were to scan the python 3.14 image, what happens when the embedded zlib has a CVE? Is its version discoverable?

I didn't mentioned this eariler, but these images would be considered new and will be categorized "experimental" for now. We don't have to solve the sbom/scanner problem now. I think we can try to get pbs themselves to produce and sbom and augment ours with theirs -- but none of this should block this PR or doing builds of these images.

reneleonhardt:
> I didn't mentioned this eariler, but these images would be considered new and will be categorized "experimental" for now. We don't have to solve the sbom/scanner problem now. I think we can try to get pbs themselves to produce and sbom and augment ours with theirs -- but none of this should block this PR or doing builds of these images.

Good... because trivy couldn't find a way to dissect pbs, embedded native libraries keep being embedded 😄
https://github.com/astral-sh/python-build-standalone/releases/tag/20260814
No SBOM in the release directly.

python/README.md
reneleonhardt: Done — Provenance links to astral's PBS, no version pin. Exact pin lives in //private/extensions:python.bzl, managed by the updater.

knife.d/update_python_archives.sh
reneleonhardt: Fixed. Was comparing patch-only; now compares the full patch+release pin, so a tag-only bump triggers an update. It was live: branch sat at 20260807 while 20260814 (expat 2.8.2 -> 2.8.3, same 3.13.15/3.14.7) was current — old logic said NO_CHANGE. Re-pinned to 20260814. Hermetic contract test covers tag-only bump, patch bump, new-minor fill, convergence.

python/README.md
```diff
@@ -0,0 +1,90 @@
...
+- Python 3.14 (`python3.14-debian13`)
+
+Each image contains everything in the [base image](../base/README.md),
+plus a standalone CPython install under `/python`:
+
+- `/python/bin/python3.X` (with `/python/bin/python` and `/python/bin/python3` symlinks)
+- `/python/lib/python3.X/` (standard library)
+- `/python/lib/libpython3.X.so.1.0`
+
+No shell, no include files. A bundled pip is on PATH; for reproducible
```
loosebazooka: it looks like we exclude pip above? from site packages, but I don't know what that means, can one still use the pip command? The distroless images typically do not include tools like pip.

python/pbs_cve_check.py
```diff
@@ -0,0 +1,129 @@
...
+# SBOM component name -> NVD CPE vendor/product (all verified extractable from
+# the binary by python/pbs_embedded_versions.py).
+CPES = {
+    "openssl-3.5": ("openssl", "openssl"),
+    "sqlite": ("sqlite", "sqlite"),
+    "zlib": ("zlib", "zlib"),
+    "expat": ("libexpat", "expat"),
+    "bzip2": ("bzip2", "bzip2"),
+    "ncurses": ("gnu", "ncurses"),
+    "xz": ("tukaani", "xz"),
```
loosebazooka: gemini says: missing zstd and mpdecimal? for 3.14?

python/README.md
```diff
@@ -0,0 +1,90 @@
...
+the corresponding stdlib modules work out of the box. `python -m venv` works and
+bootstraps pip from the bundled `ensurepip` wheels.
+
+## Provenance
+
+Python comes from a [python-build-standalone](https://github.com/astral-sh/python-build-standalone)
+prebuilt tarball (`cpython-3.X.Y+<release>-<triple>-unknown-linux-gnu-install_only.tar.gz`).
+The exact release, versions and shas are pinned in `//private/extensions:python.bzl` and
+updated via `knife update-python-archives` (see `knife.d/update_python_archives.sh`).
+A generated SPDX SBOM for the pinned release is committed at `python/pbs-sbom.spdx.json`
```
loosebazooka, line 72: we need to clarify that this is the original sbom and not the trimmed sbom.

reneleonhardt: As there is no original SBOM, this is our own from the script I created.
You wanted a SBOM to identify CVEs embedded into pbs, so I had to build the closest one I could without an official one.

loosebazooka: ah okay. I think I misinterpreted this. But there is an sbom checked into the repo.

loosebazooka:
Sorry this is probably going to require a light rebase and some minor changes:
1. rebase on main
2. update `split("~")` to `split("+")` to match that bazel change when we update in https://github.com/GoogleContainerTools/distroless/pull/2172
3. use `package_spdx` instead of `debian_spdx` to contribute spdx info from these packages

reneleonhardt:
Since the last review:

- rebased onto main;
- switched to package_spdx and updated package parsing to split("+");
- removed the Debian control-file path;
- made the updater generate the PBS SPDX SBOM and verify every selected matrix archive, including no-op runs;
- added zstd/mpdecimal extraction checks and embedded/fixed-version PBS CVE output;
- aligned Python image and symlink checks;
- updated fixtures, contract tests, and README.

loosebazooka:
Sorry this is going to take a little bit as I inspect it all. I'm concerned we're a little bit over engineered on the CVE/sbom stuff. And I don't want to have all this code lying around to maintain.

The one thing I would note for now is that we probably shouldn't be loading in arbitrary remote python (downloads.py). It looks like this can be avoided using an python ast to parse the data directly from the file instead.

reneleonhardt:
No stress, it's a big feature contribution.

## PBS update

- Updated Python 3.13 and 3.14 archives to PBS `20260901` across all matrix architectures.
- Regenerated the lockfile and PBS SPDX SBOM.
- Hardened archive extraction, cleanup, and executable-mode preservation.
- Switched manifest parsing to AST literals; downloaded `downloads.py` is never executed.

## Scope

- Removed the bespoke NVD CVE checker and fixtures.
- Kept the SBOM as PBS release provenance, not a complete image SBOM.
- Updated contract coverage for versioned `libpython` files and embedded-library verification.

## Documentation

- Proofread Python image documentation and related comments and tests.

reneleonhardt:
### Improvements

- Added opt-in PBS beta/RC selection for matrix minors; stable discovery remains stable-only.
- Added clear failures for missing opted-in architectures or assets.
- Preserved manual opt-out without automatic re-addition.
- Kept generated archives, versions, metadata, and `MODULE.bazel` visibility synchronized.
- Added numeric minor ordering and prerelease binary-name handling.
- Hardened PBS `libpython` extraction against tar path traversal.
- Expanded offline contract coverage for prereleases, opt-out, failures, ordering, and unsafe archives.

### Prerelease opt-in

- Matrix contains `3.15`:
  - PBS `3.15.0b4` is selected.
  - Later `3.15.0rc1` replaces it.
  - The newest beta/RC is selected by version sorting.

- Maintainer removes `3.15`:
  - The updater removes its generated archives and repository visibility.
  - Prereleases are no longer auto-added.
  - Stable `3.13`/`3.14` updates still proceed normally.

- When a stable `3.15.x` asset appears, the updater detects it and automatically adds `3.15` back with the stable archive.

The contract test covers beta → RC, opt-out cleanup, and stable re-addition.

If you don't want to opt-in safely to beta/RC, I can remove that support.

private/pkg/test/oci_image/BUILD.bazel
```diff
@@ -7,6 +7,7 @@ oci_image(
     name = "image_arm64",
     architecture = "arm64",
     os = "linux",
+    tags = ["debian12"],
```
loosebazooka: I'm not sure why this is happening, but this shouldn't be necessary

private/util/validate_usr_symlinks.bzl
```diff
@@ -11,6 +11,8 @@ load("@aspect_bazel_lib//lib:tar.bzl", "tar_lib")
 def _validate_usr_symlink_impl(target, ctx):
     if target.label.name.find("debian12") != -1:
         return []
+    if hasattr(ctx.rule.attr, "tags") and "debian12" in ctx.rule.attr.tags:
```
loosebazooka: this code seems a little out of place.

loosebazooka:
I'm sorry, I got a little distracted, I'll try to get back to this soon

PR #2166, opened 2026-08-21
Use bazel bsdtar instead of system tools

loosebazooka:
Turns out our cloud build doesn't like the compression that the adoptium control files were using. So use the toolchain for tar instead of the system tools.

### DIFF
--- a/private/pkg/debian_spdx.bzl
+++ b/private/pkg/debian_spdx.bzl
@@ -1,28 +1,35 @@
+load("@aspect_bazel_lib//lib:tar.bzl", "tar_lib")
+
 SPDX_CMD = """\
+readonly bsdtar="$1"
+readonly control="$2"
+readonly data="$3"
+readonly package_name="$4"
+readonly generator="$5"
+shift 5
+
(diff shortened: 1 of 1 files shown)


private/pkg/debian_spdx.bzl
```diff
@@ -1,28 +1,35 @@
...
+readonly control="$2"
+readonly data="$3"
+readonly package_name="$4"
+readonly generator="$5"
+shift 5
+
 tmp="$(mktemp -d)"
 
-tar -xf "$1" -C "$tmp" "./control" || tar -xf "$1" -C "$tmp" "control"
+"$bsdtar" -xf "$control" -C "$tmp" "./control" 2>/dev/null || "$bsdtar" -xf "$control" -C "$tmp" "control" 2>/dev/null || true
```
bobcallaway: you want the `|| true` here?

bobcallaway (approved):

private/pkg/debian_spdx.bzl
loosebazooka: definitely not.

